PR #356, opened 2025-11-04
Enable no local timesteping

willem4:
# What was done 

- minimum transport time step `dtmin_transp` set now computed in setdtorg()
- if transportlocaltimestep=0 then reduce the current flow time step `dts` by the `dtmin_transp` if smaller than the current flow time step.
 
# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Test in issue is correct now

# Tests 
- [x] Tests updated \
e02_f022_c107
- [ ] Not applicable 

# Documentation  
- [x]	Documentation updated \
Created issue: UNST-9393
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/get_jaupdatehorflux.f90
```diff
@@ -60,59 +59,55 @@ subroutine get_jaupdatehorflux(nsubsteps, limtyp, jaupdate, jaupdatehorflux)
 
       if (timon) call timstrt("get_jaupdatehorflux", ithndl)
 
-      if (nsubsteps == 1) then
```
FlorisBuwaldaDeltares: it is not clear to me why this check (and the corresponding input argument) could be removed. It is also not clear to me why they were replaced by a check on limtyp.

FlorisBuwaldaDeltares (commented):
I don't understand the whole local timestepping business but apart from 1 suggestion and 1 question the code looks good. If @V-Chavarrias agrees with the implementation then go ahead and merge

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/get_ndeltasteps.f90
```diff
@@ -86,10 +86,10 @@ subroutine get_ndeltasteps()
             ndeltasteps(ln(1, LL)) = ndeltasteps(ln(2, LL))
          end do
 
-!      if ( nsubsteps.gt.1 ) then
-!         write(6,*) dtmin
-!      end if
-
+         if (ja_transport_local_time_step==2) then
```
V-Chavarrias: Could this be moved right after `nsubsteps` is computed (new line 71) and prevent the loop on `ndxi`? E.g.:

```
      else
         logtwo = log(2.0_dp)
         nsubsteps = max(1, 2**int(log(dts / dtmin) / logtwo + 0.9999_dp))
         if (ja_transport_local_time_step==2) then   
             ndeltasteps = nsubsteps
             numnonglobal = ndxi
         else

         dtmin = dts / nsubsteps

!     get number of substeps
         do kk = 1, Ndxi
            dt = dtmax(kk)
            if (dt < dts) then
               ndeltasteps(kk) = min(2**int(log(dt / dtmin) / logtwo), nsubsteps)
               numnonglobal = numnonglobal + 1
            else
               ndeltasteps(kk) = nsubsteps
            end if
         end do
!     fictitious boundary cells
         do LL = Lnxi + 1, Lnx
            ndeltasteps(ln(1, LL)) = ndeltasteps(ln(2, LL))
         end do
endif
         
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/ini_transport.f90
```diff
@@ -255,10 +256,10 @@ subroutine ini_transport()
...
 !  local timestepping
       time_dtmax = -1.0_dp ! cfl-numbers not evaluated
-      nsubsteps = 1
-      ndeltasteps = 1
-      jaupdatehorflux = 1
-      numnonglobal = 0
+      nsubsteps = imiss
+      ndeltasteps = imiss
+      jaupdatehorflux = imiss
+      numnonglobal = imiss
```
V-Chavarrias: For me to remember; nothing to be done with this comment. 

`imiss` is -999. This is filled in `alloc_transport`:

```
      call realloc(ndeltasteps, Ndx, keepExisting=.false., fill=1)
      call realloc(jaupdate, Ndx, keepExisting=.false., fill=1)
      call realloc(jaupdatehorflux, Lnx, keepExisting=.false., fill=1)
```
 and `numnonglobal` in `get_ndeltasteps`, and it is always set.

V-Chavarrias (commented):
Thanks!

PR #376, opened 2025-11-12
Don't read 'value' for initialvertical-quantities

JulienGroenenboom:
# What was done 
- Don't read 'value' for initialvertical-quantities
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [x]	Documentation updated: revision 80670
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
@@ -655,7 +655,7 @@ contains
          end if
 
          ! read value
-         if ((filetype == inside_polygon) .and. .not. strcmpi(quantity, 'initialvertical', 15)) then
+         if (filetype == inside_polygon .and. method == METHOD_CONSTANT) then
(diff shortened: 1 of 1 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
```diff
@@ -655,7 +655,7 @@ subroutine readIniFieldProvider(inifilename, node_ptr, groupname, quantity, file
          end if
 
          ! read value
-         if (filetype == inside_polygon) then
+         if ((filetype == inside_polygon) .and. .not. strcmpi(quantity, 'initialvertical', 15)) then
```
arthurvd: Let's try and not introduce hardcoded checks on quantity names anymore, unless absolutely unavoidable.
In this case, there's a more elegant alternative:
```suggestion
         if (filetype == inside_polygon .and. method == METHOD_CONSTANT) then
```

PR #412, opened 2025-11-28
Fm/feature/unst 9255 2d 2d contacts

FlorisBuwaldaDeltares:
# What was done 

- second set of contacts called contacts_2d2d
- count 2d2d contacts outside of 1D network writing in unc_write_net_ugrid_2
- simply duplicate the ugrid_write-mesh contacts with the appropriate arguments for the 2D-2D contacts
- new function ionc/ug_get_contacts_topology_dimensions that retreives the "contacts" attribute and parses the meshes the contacts lie between and queries their topology
- Dynamically connect contacts to 1D node or create new 2D node in unc_read_net_ugrid
- add 2D2D contacts to "open spots" in the link administration in preparecells()
 

# Evidence of the work done 

Load and save file "contacts_net.nc" in the attached JIRA issue

- [ ]	Video/figures \
- [ x]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
@@ -15695,7 +15695,7 @@ contains
       use m_partitioninfo
       use m_alloc
       use dfm_error
-      use m_save_ugrid_state !stores the contactname_1D2D and other saved ugrid names
+      use m_save_ugrid_state ! stores the mesh contacts, and other network topologies
       use m_CrossSections
       use m_flowtimes, only: handle_extra
       use Timers
@@ -16776,7 +16776,7 @@ contains
             ierr = ionc_get_contacts_topology_dimensions(ioncid, im, mesh1_topo_dim, mesh2_topo_dim) ! networkIndex not used here
             ierr = ionc_get_contacts_count_ugrid(ioncid, im, numl1d2d_read)
             numl_read = numl_read + numl1d2d_read
-            numk_read = numk_read + numl1d2d_read * mesh1_topo_dim !> if mesh1 is 2D mesh we add 2 nodes per contact link
+            numk_read = numk_read + numl1d2d_read * mesh1_topo_dim ! If mesh1 is 2D mesh we add 2 nodes per contact link (mesh2 is always assumed to be 2D mesh already).
          end do
       else
          ! No UGRID, not a problem, code below will fall back to trying old format.


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -11797,9 +11800,23 @@ subroutine unc_write_net_ugrid2(ncid, id_tsp, janetcell, jaidomain, jaiglobal_s)
...
+            L = temp_indices(i)
+            n1 = abs(lne(1, L))
+            n2 = abs(lne(2, L))
+            contacts_2D2D(1:2, i) = [n1, n2]
+         end do
+      end if
 
       ! 1D network geometry
-      if (numl1d > 0) then
+      if (numl1d > 0 .and. numl1d > n2d2dcontacts) then
```
arthurvd:
```suggestion
      if (numl1d > n2d2dcontacts) then
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -12108,10 +12125,17 @@ subroutine unc_write_net_ugrid2(ncid, id_tsp, janetcell, jaidomain, jaiglobal_s)
 
          !define 1d2dcontacts only after mesh2d is completly defined
          if (n1d2dcontacts > 0) then
-            ierr = ug_def_mesh_contact(ncid, id_tsp%meshcontacts, trim(contactname), n1d2dcontacts, id_tsp%meshids1d, id_tsp%meshids2d, UG_LOC_NODE, UG_LOC_FACE, start_index)
+            ierr = ug_def_mesh_contact(ncid, id_tsp%meshcontacts_1D2D, trim(contactsname)//'1D2D', n1d2dcontacts, id_tsp%meshids1d, id_tsp%meshids2d, UG_LOC_NODE, UG_LOC_FACE, start_index)
```
arthurvd: This reveals a bug that was already present: we now start supporting multiple contact variables in a single .nc file, but until now the module `m_save_ugrid_state` assumed only a single contact variable, see:
```fortran
   character(len=255) :: network1dname, mesh2dname, mesh1dname, contactname
```
That name gets read (in a loop!) in `unc_read_net_ugrid()`:
```fortran
ierr = ionc_get_contact_name(ioncid, im, contactname)
```
So, upon repeated reading/writing you would be appending `1D2D1D2D1D2D`, etc.
I think the solution is to change `contactname` -> `contact1d2d_name` and add a new `contact2d2d_name`. and properly set and use that in our unc_read and write  subroutines.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -12634,20 +12667,28 @@ subroutine unc_read_net_ugrid(filename, numk_keep, numl_keep, numk_read, numl_re
...
+            kn(2, numl_last + l) = numk_last + l
 
-            contact1d2didx(1, contactnlinks + L) = mesh1indexes(L)
-            contact1d2didx(2, contactnlinks + L) = mesh2indexes(L)
-            if (nodesOnBranchVertices == 1) then
-               kn(1, numl_last + l) = mesh1dUnmergedToMerged(mesh1indexes(l))
-            else
-               kn(1, numl_last + l) = mesh1indexes(l)
+            if (mesh1_topo_dim == 2) then
+               XK(numk_last + ncontacts + l) = xface(mesh1indexes(l))
```
arthurvd: I understand where this code comes from (one net node used to be added for each contactlink, now it can be 2 net nodes for a single link), but it's a bit counterintuitive that you're first adding all *first* points for 2d2d link to the net node arrays, and after that the list of *second* points for 2d2d links. Could you change that both first and second points per contact link are added consecutively? It probably means changing the `numk_last = numk_last + ncontacts` update to incremental `numk_last = numk_last + 1` inside the `L` loop.
By the way, notice how confusing a lower case `do l=` loop is: the GitHub font has a not super obviously visible difference between `+ l` and `+ l` ;-)

src/utils_lgpl/io_netcdf/packages/io_netcdf/src/io_ugrid.F90
```diff
@@ -4169,6 +4169,86 @@ function ug_get_contacts_count(ncid, contactids, ncontacts) result(ierr)
...
+      icolon = index(contact_attr, ':')
+      if (icolon == 0) then
+         call check_ug_error(UG_SOMEERR, 'Invalid contact attribute format: missing first colon'//error_message)
+         ierr = UG_SOMEERR
+         return
+      end if
+      mesh1_name = trim(adjustl(contact_attr(1:icolon - 1)))
+
+      ! Skip to second mesh name
+      istart = icolon + 2 ! skip space
```
arthurvd: Use +1 and adjustl instead: the space after `:` is not always there/optional. Only the space between first `mesh1:face` and second `mesh2:node` part is optional (and may be multiple spaces).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
FlorisBuwaldaDeltares: I thought it was unnecessary but you are right, repeated reading and writing does not work this way. I will change it

PR #430, opened 2025-12-05
UNST-9429: Freezing point is wrongly computed for background salinity

thomas-pijls:
# What was done 

The logic to compute the freezing point with salinity would only run when salinity was actively modeled. The logic was changed to also compute the freezing point when a background salinity was present, but no salinity was modeled (i.e. a constant salinity). 

A test case for this needs to be added to the testbench.
 

# Evidence of the work done 

# Tests 

- [x]	Add test to testbench

# Documentation  

- [x]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/extract_constituents.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/extract_constituents.f90
@@ -139,17 +139,17 @@ contains
          if (use_salinity_freezing_point) then
             if (isalt > 0) then ! if salinity is modeled, use local salinity to determine freezing point
                do kk = 1, ndx
-                  k = ktop(kk)
+                  k = ktop(kk) ! only the top layer is checked for freezing point
                   freezing_point_temperature = real(freezing_temperature(real(constituents(isalt, k), fp)), dp)
                   if (constituents(itemp, k) < freezing_point_temperature) then
                      constituents(itemp, k) = freezing_point_temperature
                      cells_with_min_limit = cells_with_min_limit + 1
                   end if
                end do
-            else if (isalt == 0) then ! if salinity is not modeled, use background salinity to determine freezing point
+            else ! if salinity is not modeled, use background salinity to determine freezing point
                freezing_point_temperature = real(freezing_temperature(backgroundsalinity), dp)
                do kk = 1, ndx
-                  k = ktop(kk)
+                  k = ktop(kk) ! only the top layer is checked for freezing point
                   if (constituents(itemp, k) < freezing_point_temperature) then
                      constituents(itemp, k) = freezing_point_temperature
                      cells_with_min_limit = cells_with_min_limit + 1


adrimourits:
If adding a testcase to the testbench is NOT included in this issue, then I prefer:
- Create a new issue for adding a testcase (Erik is going to do that)
- Link issue UNST-9429 to the newly created issue
- In "Definition of Done -checklist", mention the new issue under "Follow up issues created"
- In this pull request description: replace the checked "add testcase to testbench" (that suggests it is done as part of this issue) by "To Do: add testcase to testbench: issue-reference"

To avoid all these references and new issues etcetera: Preferably, adding a testcase IS part of this issue (if needed).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/extract_constituents.f90
```diff
@@ -135,15 +135,27 @@ subroutine extract_constituents()
...
-            do kk = 1, ndx
-               k = ktop(kk)
-               freezing_point_temperature = real(freezing_temperature(real(constituents(isalt, k), fp)), dp)
-               if (constituents(itemp, k) < freezing_point_temperature) then
-                  constituents(itemp, k) = freezing_point_temperature
-                  cells_with_min_limit = cells_with_min_limit + 1
-               end if
-            end do
+
+         if (use_salinity_freezing_point) then
```
adrimourits: Can you keep the comment about "only at surface limit to freezing point" around here?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_transport/extract_constituents.f90
```diff
@@ -135,15 +135,27 @@ subroutine extract_constituents()
...
+            if (isalt > 0) then ! if salinity is modeled, use local salinity to determine freezing point
+               do kk = 1, ndx
+                  k = ktop(kk)
+                  freezing_point_temperature = real(freezing_temperature(real(constituents(isalt, k), fp)), dp)
+                  if (constituents(itemp, k) < freezing_point_temperature) then
+                     constituents(itemp, k) = freezing_point_temperature
+                     cells_with_min_limit = cells_with_min_limit + 1
+                  end if
+               end do
+            else if (isalt == 0) then ! if salinity is not modeled, use background salinity to determine freezing point
```
adrimourits: I propose to remove `if (isalt == 0) then`: it avoids confusion about "else branch missing".

adrimourits (changes requested):

adrimourits:
> If adding a testcase to the testbench is NOT included in this issue, then I prefer:
> 
> * Create a new issue for adding a testcase (Erik is going to do that)
> * Link issue [UNST-9429](https://issuetracker.deltares.nl/browse/UNST-9429) to the newly created issue
> * In "Definition of Done -checklist", mention the new issue under "Follow up issues created"
> * In this pull request description: replace the checked "add testcase to testbench" (that suggests it is done as part of this issue) by "To Do: add testcase to testbench: issue-reference"
> 
> To avoid all these references and new issues etcetera: Preferably, adding a testcase IS part of this issue (if needed).

I see did a testcase IS added as part of this issue. Please skip the related remarks. Sorry for the confusion.

PR #432, opened 2025-12-05
Copy each dependency dll into a separate build folder for the tests t…

harmenwierenga:
…o prevent parallel copying conflicts
 

# Evidence of the work done
Before:
before
After:
after
Tests still pass:
tests_pass

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link UNST-9490


FlorisBuwaldaDeltares (approved):
perfect, thanks!

harmenwierenga:
> Do note that I also saw build problems on Linux when building multiple test projects in the same folder

Oh that would point to a different cause, because we only copy the dependency binaries to the build folder on windows. Linux uses the rpath for resolving the test dependencies I think

PR #447, opened 2025-12-16
DEVOPSDSC-850: Remove team-city folder

M-D-J:
# What was done 
A leaked credential was found in src/engines_gpl/dflowfm/scripts/team-city/example_testcases.py. This credentials is not valid anymore and the team-city folder can be removed since it is not used.

# Issue link
DEVOPSDSC-850


robin-deltares (approved):
The folder is unused and can be removed.

PR #459, opened 2025-12-23
Fm/task/unst 9384 1 d2 d flow1 d solver

FlorisBuwaldaDeltares:
# What was done 

- e.g. Introduces a new functionality on energy losses at bridge peirs 
- e.g. … 
 

# Evidence of the work done 

- [ x]	Video/figures \
Identical results for long culvert
image
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -509,6 +510,7 @@ subroutine loadModel(filename)
       ! set administration arrays and fill cross section list. So getbobs for 1d can be called.
       call timstrt('Initialise 1d administration', timerHandle)
       call initialize_1dadmin(network, network%numl, numl)
+      call set_default_1d2d_crs(network)
```
FlorisBuwaldaDeltares: should this call maybe be part of "initialize_1dadmin"? it's a bit small to get a full call in the main unstruc_model routine

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,92 @@
...
+
+      ! Now create a cross-section instance
+      if (network%crs%count + 1 > network%crs%size) then
+         call realloc(network%crs)
+      end if
+      icrs = network%crs%count + 1
+      call finalizeCrs(network, network%crs%cross(icrs), idef, icrs)
+      temp_line2cross = network%adm%line2cross
+      deallocate (network%adm%line2cross)
+      allocate (network%adm%line2cross(numl1d, 3))
```
FlorisBuwaldaDeltares: isn't there a realloc for this?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,92 @@
...
+      network%adm%line2cross(1:size(temp_line2cross, 1), :) = temp_line2cross
```
FlorisBuwaldaDeltares: this is just "keepexisting = .true." behavior if i'm not mistaken

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,92 @@
...
+      do L = 1, numl1d
+         if (kn(3, L) == 5 .and. all(network%adm%line2cross(L, :)%c1 == -1)) then !> don't overwrite long culvert lin2cross
```
FlorisBuwaldaDeltares: right now this seems to only be for type 5 links, should we update the name of the crs and function name to reflect this?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
```diff
@@ -0,0 +1,164 @@
...
+                if (col == 0) then
+                    kn(:, link_index) = [up_node_index, bottom_left_node_index, LINK_2D]
+                    link_index = link_index + 1
+                end if
+            end do
+        end do
+
+        ! Initializes node, face and flow geometry stuff.
+        call setnodadm(0)
+        call findcells(0)
```
FlorisBuwaldaDeltares, line 89: I think these two routines print a lot of information to the console, any way to prevent that?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_getprof_1d.f90
```diff
@@ -0,0 +1,686 @@
...
+        use Timers, only: timini, timon
+        use m_partitioninfo, only: jampi
+
+        call timini() ! Initialize timers (otherwise `flow_geominit` crashes)
+        timon = .false. ! Disable timers because we're running unit tests.
+        jampi = 0 ! Disable MPI because we're running unit tests.
+    end subroutine
+
+    !> Initialize a simple rectangular cross section for testing
+    subroutine setup_network_rectangular_cross_section(network, width, height, friction_type, friction_value)
```
FlorisBuwaldaDeltares, line 25: I think all those cross section setup routines are pretty elegant! however they will all break if we ever change something about the cross_section implementation.. i'm almost thinking, isn't it almost easier to mock a .crs file and use dflowfm's own crs reader to initialize the cross_section? or is that much more complex than it sounds?

FlorisBuwaldaDeltares (commented):
I can't approve as it is my own PR, but great work! very good that we're getting proper unit tests for a pretty hard-to-test routine (getprof_1d)

Otherwise a few tiny comments

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
rene-deltares: I think that makes sense now that we're going to use the 1D code to handle the 1D2D links. I'll try to move it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: I don't think so, but I'll check. The type of `line2cross` is `type(t_chainage2cross), dimension(:, :)`

Update: I couldn't find an `interface realloc` for `t_chainage2cross`. Is it a good idea to add yet another one? I don't really like all these `realloc` implementations scattered all over the codebase. The ones for primitive types and multidimensional arrays are all concentrated in a single place, which I think is quite handy.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: Yes, but if I understand correctly we're also going to make a default cross section for type 3 and 7 links. We're going to have to do that before we can remove the 'profile1D' stuff from `getprof_1d` I think.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
rene-deltares: Yes, there is! The deltares_common logging module has an option to turn off stdout/stderr writing. I'll try to turn it off

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_getprof_1d.f90
rene-deltares: In my experience creating the stuff in memory is always easier and more flexible. Either with a test helper or a factory function. If you have to refactor the cross section type you just have to also refactor this helper, but as long as the 'creation code' is not duplicated in every test, but confined to a single helper routine, then that's no big deal right. The nice thing about helper routines is: You reuse them, so you have then one place to refactor, instead of 'all the tests'. And you can't forget it, because the compiler will give an error.

You know how I feel about file IO in unit tests: I avoid it whenever possible. Using scratch files can sometimes be a solution, so you don't create actual files. But even this approach result in brittle elaborate setup code compared to just creating the types.

FlorisBuwaldaDeltares: Yes, I don't like file IO either, in some other unit tests I just created the property tree in memory:
image
What i like is that in that way, the only way your test breaks is if the public api changes, as that's what you're calling in your unit test.

But i'm not opposed to test helpers, if you prefer them that's alright with me!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
FlorisBuwaldaDeltares: ok, you could implement the realloc interface for this type if you feel like it, but we could also leave the code as it is, either is fine by me

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
rene-deltares: Moving this inside `initialize_1dadmin` has "some feet in the earth". `initialize_1dadmin` is in `utils_gpl` and `set_default_1d2d_crs` uses some stuff from the dflowfm (because it loops over the 1D links and checks the link types). So I think we'll get circular dependencies, right? That is, unless I pass whatever information `set_default_1d2d_crs` needs through the parameters.

FlorisBuwaldaDeltares: oh no.. ok then we can't put it in, initialize_1dadmin is flow1d core and that should have zero coupling or FM knowledge.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1239,6 +1237,41 @@ subroutine addlongculvertcrosssections(network, branchId, csdefId, zpl, iref)
...
+         pCrs => network%crs%cross(icrs)
+         write (kchar, '(I0)') flowlink
+         pCrs%csid = trim(branchId)//'_1D2D_'//trim(kchar)
+         call finalizeCrs(network, pCrs, iref, icrs)
+         network%adm%line2cross(flowlink, :)%c1 = icrs
+         network%adm%line2cross(flowlink, :)%c2 = icrs
+         network%adm%line2cross(flowlink, :)%f = 1.0_dp
+         network%adm%line2cross(flowlink, :)%distance = 0.0_dp
+      end if
+
```
KoenBussemakerDeltares, line 1273: What happens if this fails? Is a silent failure acceptable? Should we communicate the result back and deal with it at the call site?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
rene-deltares: Done by  disabling logging in the get_prof1d tests.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
rene-deltares: Ok, yeah, I tried breaking it apart and moving things, but in the end I wasn't really happy with it. I ended up just refactoring `set_default_1d2d_crs` a little bit. Maybe at some point the 1D2D link cross sections should be taken into account in initialize_1dadmin proper.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: I broke the reallocation off into a separate subroutine `realloc_line2cross`. I'm not sure if we should include it in the public interface now, so it's kind of custom for this module. I think in the end the better solution is to allocate space for the 1D2D link administration stuff in the channel 1D initialization routines properly.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1239,6 +1237,41 @@ subroutine addlongculvertcrosssections(network, branchId, csdefId, zpl, iref)
...
+   !> add special 1D2D crossection for the longculvert and add it to the line2cross array
+   subroutine add_longculvert_1D2D_crosssection(network, flowlink, branchId, csdefId)
+      use precision, only: dp
+      use m_hash_search
+      use m_readCrossSections
+      use m_network
+      type(t_network), intent(inout) :: network !< Network structure
+      integer, intent(in) :: flowlink !< Flowlink number on which to place the cross section. Should be 1D2D link belonging to the long culvert
+      character(len=IdLen), intent(in) :: branchId !< Branch id on which to place the cross section
```
KoenBussemakerDeltares: Use snake case: branch_id, csdef_id

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,146 @@
...
+   subroutine add_cross_section(network, idef, icrs)
+      use m_network, only: t_network
+      use m_readcrosssections, only: finalizeCrs
+      use m_CrossSections, only: realloc
+
+      type(t_network), intent(inout) :: network !< The 1D channel network object.
+      integer, intent(in) :: idef !< The cross section definition index.
+      integer, intent(out) :: icrs !< The index of the new cross section instance.
+
+      associate(cross_sections => network%crs)
```
KoenBussemakerDeltares, line 100: Can we rename `network%crs` to `network%cross_sections`?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,146 @@
...
+      if (associated(line2cross)) then
+         if (all(new_size == ubound(line2cross))) then
+            return ! Array is already the right size. Nothing to do
+         end if
+      end if
+
+      ! Reallocation required
+      allocate(temp(1:new_size(1), 1:new_size(2)), stat=stat_)
+      if (stat_ /= 0) then
+         if (present(stat)) then
```
KoenBussemakerDeltares: I would make `stat` non-optional, both to force the caller to deal with it (or explicitly ignore it there) and to make this code easier (you'd no longer need `stat_` and these `present` checks)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
```diff
@@ -0,0 +1,146 @@
...
+         call finalizeCrs(network, cross_sections%cross(icrs), idef, icrs)
+      end associate
+   end subroutine add_cross_section
+
+   !> Re-allocate and reshape the `line2cross` to have `new_size`. If reallocation is necessary
+   !! then copy the old elements to the new array.
+   subroutine realloc_line2cross(line2cross, new_size, stat)
+      use m_GlobalParameters, only: t_chainage2cross
+
+      type(t_chainage2cross), dimension(:,:), pointer, intent(inout) :: line2cross !< Link to cross sections administration.
```
KoenBussemakerDeltares, line 114: Is it line-to-cross or line2-cross? It's both ambiguous and a bit 2000-era-childish. I'd go for `line_to_cross`.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
```diff
@@ -0,0 +1,164 @@
...
+        nump1d2d = 0
+
+        ! Reset flow administration
+        if (allocated(xz)) then
+            deallocate(xz)
+        end if
+        ndx = 0
+    end subroutine cleanup_network_data
+
+end module m_network_helpers
```
KoenBussemakerDeltares, line 179: Suggestion: create a derived type that does the cleanup of the network in the destructor. This concept is very commonly used in C++ (It's call RAII if you're interested). The nice thing is that this works for happy flow as well as early returns due to e.g. errors. which lowers the risk of one failing test causing a cascade of other test failures due to a corrupted global state.

Here's a sketch of the idea (made by copilot, I didn't fully check it)

```
module TestGridNetworkModule
  implicit none

  type :: TestGridNetwork
  contains
    final :: destroy
  end type TestGridNetwork

  interface TestGridNetwork
    module procedure :: create_TestGridNetwork
  end interface TestGridNetwork

contains

  function create_TestGridNetwork() result(obj)
    type(TestGridNetwork) :: obj
    ! Set up the grid here
  end function create_TestGridNetwork

  subroutine destroy(this)
    type(TestGridNetwork), intent(inout) :: this
    ! Clean up and reset globals
  end subroutine destroy

end module TestGridNetworkModule
```

Now you can just create a TestGridNetwork instance in your test body and just let it go out of scope, cleanup will happen automagically.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
rene-deltares: I guess this routine silently does nothing if the branch or cross section can't be found. In that case I think we should ERROR out because that should not happen. Communicating an error code back to the callsite and logging the error there is a good option I think.

rene-deltares: Oh my god, I actually think testcase e02_f002_c102 runs into this silent failure. because the `branch_id` is all zero bytes, and `hashsearch` returns -1.
I think the effect is that the long culvert is treated like a streetinlet. How bizarre...

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: Well, we can, of course, but I don't think I should do it in this PR. It would introduce rename changes in many, many places in the codebase. And looking at the `t_network` type, there are other obvious contenders for renaming: https://github.com/Deltares/Delft3D/blob/8ce85add34926b40abd7dd4b12dde67d55ae9324/src/utils_gpl/flow1d/packages/flow1d_core/src/Network.f90#L81

So, I'm not against this, but I would pitch it as another renaming refactoring a la the jahis/jamap configs.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: I'm 2cool4school, but the name is a reference to the 1d administration type: https://github.com/Deltares/Delft3D/blob/8ce85add34926b40abd7dd4b12dde67d55ae9324/src/utils_gpl/flow1d/packages/flow1d_core/src/Network.f90#L67
I can rename it in this subroutine, but I would pitch a renaming refactor for this to the team as well

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
rene-deltares: Ok, in the case of 2D2D links (longculverts consisting of only one link between two 2D cells), there is no branch (1D concept AFAIK), but in that case the structure has a 'contactId' instead. I added some code to the call site of add_longculvert_1d2d_crosssection to use the branchId in the case of 1D2D links, and the contactId in the case of 2D2D links. I pass it as 'link_id' and it only affects the cross section identifier. The cross section definition is the only thing that absolutely cannot be missing.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
rene-deltares: Forcing people to check OOM errors, that's hardcore! :D
This subroutine is modelled after the auto-generated realloc subroutines, but stripped down a bit, so I kept the optional stat argument.

This subroutine is private and there's one call-site, so that's OK, but all I can do in this case is log an error and exit the program.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
rene-deltares: I've been meaning to try this out, so I will try. The only "conceptual" objection I have to this is that such a derived type does not 'own' the globals that it is initializing and destroying, because they're globals. But I think it already helps a lot that we don't have to call 'cleanup' routines ourselves in every possible code path.

rene-deltares: So I got something that works. I initially wanted to create factory functions that return the `t_grid_helper` instance. But when I did that the destructor was called twice, because Fortran apparently makes a shallow copy of a type when you return it from a function, which is then destroys, and then when the copy goes out of scope, destroy is called again. I'm pretty happy-ish with how it is now. The destructor is called exactly once!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
KoenBussemakerDeltares: Fair enough, let's not do it in this PR

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
KoenBussemakerDeltares: Like the other comment, if it changes too much and clutters up this PR, let's not do it here

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/network_helpers.f90
KoenBussemakerDeltares: I like it, it's quite clean. The type does indeed not own the globals but IMO the pro's outweigh the cons. Technically there should only ever be one `t_grid_helper` at one type, an instance counter could help with that, but that already feels like overengineering.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/default1d2d.f90
KoenBussemakerDeltares: Ugh, I just wish Fortran had a decent exception model. The current solution is fine then.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
KoenBussemakerDeltares: I'm very happy that I left this comment haha.

Your call site changes look plausible but you might want to check with Floris if this is the correct thing to do, I can't judge that. This comment is adequately resolved so feel free to mark this thread as resolved :)

FlorisBuwaldaDeltares: nice bugfix! call site changes look fine one question.. do we even need link_id in this function anymore?
I see it is only used to define the cross section csid, but if that id is not read anywhere else in the codebase it is not actually necessary?
The variable description `branch id or contact id on which to place the cross section` seems to imply more is happening than only an id being filled

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_longculverts_data.f90
```diff
@@ -72,16 +74,24 @@ module m_longculverts_data
...
 
+   !> Returns whether or not this longculvert consists of only one 2D2D link.
+   pure function is_2D2D(self) result(res)
+      class(t_longculvert), intent(in) :: self
+      logical :: res
+      res = .false.
+      if (newculverts .and. allocated(self%netlinks)) then
+         res = size(self%netlinks) == 1
+      end if
+   end function is_2D2D
```
FlorisBuwaldaDeltares, line 85: I still don't like the "self" argument, but as far as type-bound procedures go this is a nice one :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
rene-deltares: You're right that 'link_id' is only a string used to set the id of the new cross section. I doubt that that id has any effect on later on besides maybe being read and printed in a log or something. 
I do think it's nice to have the 'contact/branch id' in there so you can more easily tie the name of the cross section to the name of the branch/contact in structures.ini. Otherwise it's just a link number or the empty string.

KoenBussemakerDeltares (approved):

PR #701, opened 2026-03-10
DELWAQ-1244: Fix to agrhyd when layer information is missing in the input waqgeom-file

arjenmarkusdeltaresnl:
If the waqgeom file does not contain layer information, use a default layering. Also correct a small but crucial mistake (number of layers needs to be passed to the waqgeom component)

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


michelle-deltares (changes requested):
Hi Arjen,

I do have some questions. I would think that the number of layers is only missing if it is not present in the input. Your fall back might work, but I think this will only happen with older versions of FM. It would be handy if you extract the hyd- and waqgeom-file from your large zip, and place it in the maintenance folder for this issue.

also, one test seems to fail.

groetjes,
Michelle

src/utils_lgpl/io_hyd/packages/io_hyd/src/read_hyd_init.f90
```diff
@@ -108,6 +108,11 @@ subroutine read_hyd_init(hyd)
             endif
          endif
 
+         ! Fallback: the waqgeom file has no layer information?
```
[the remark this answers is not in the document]
arjenmarkusdeltaresnl: But then you would have to make the difference in the rest of the code. The way things are filled in works even in this case.

michelle-deltares: Maybe we can now remove the fallback option?

src/utils_lgpl/io_hyd/packages/io_hyd/src/read_hyd_init.f90
```diff
@@ -200,4 +205,27 @@ subroutine read_hyd_init(hyd)
   990 write(lunrep,*) 'error allocating memory:',ierr_alloc
       write(lunrep,*) 'hyd%num_exchanges:',hyd%num_exchanges
       call stop_with_error()
+
+      contains
+
+      ! Fallback if the waqgeom file holds no layer information
+      ! - assume uniform sigma layers
+      subroutine fillin_layers( waqgeom, num_layers )
```
michelle-deltares: If we remove the fallback option, we also don't need this new subroutine.

michelle-deltares (changes requested):
Code looks good, but I suggest to remove the fallback option too. As discussed, don't try to add information that might not be correct (e.g. z-sigma, uneven layer distributions).

Also, you will need to update from main (see teams: https://teams.microsoft.com/l/message/19:hgw82j1vhftgEHFZiYkDI0Iy8iTXMdG4oprfl-litpU1@thread.tacv2/1776079013724?tenantId=15f3fe0e-d712-4981-bc7c-fe949af215bb&groupId=13e11e46-4789-469c-ab6c-82a96e94a3fd&parentMessageId=1776079013724&teamName=HMWESS&channelName=General&createdTime=1776079013724)

michelle-deltares (approved):

PR #745, opened 2026-03-26
Added first implementation of preCICE adapter and builder

erwin-mulder:
Added first implementation of preCICE adapter and builder.

# What was done 
 Introduces the first simple version of the preCICE adapter and builder objects. 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9549


adrimourits (commented):
Follow-up review will be needed. Especially once the calls from the D-Flow FM code are added.

erwin-mulder (commented):
Answered questions and added facade and calls to it and the builder and adapter in new commit.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
```diff
@@ -337,6 +339,14 @@ integer function flowinit() result(iresult)
       !call update_statistical_output(out_variable_set_map%configs,dts)
       !call update_statistical_output(out_variable_set_clm%configs,dts)
 
+      ! Build and initialize precice adapter if enabled.
+      if (precice_adapter_is_enabled()) then
+         fm_precice_adapter => precice_adapter_get_adapter()
```
harmenwierenga, line 344: What is the reason that there are two if checks, one for checking if precice is enabled and one for checking if the adapter is associated? Should it be an error if the adapter is not associated?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
```diff
@@ -337,6 +339,14 @@ integer function flowinit() result(iresult)
...
+         if (ASSOCIATED(fm_precice_adapter)) then
```
harmenwierenga: Lower case

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/facade.F90
```diff
@@ -0,0 +1,54 @@
...
+   public :: precice_adapter_enable
+   public :: precice_adapter_is_enabled
+   public :: precice_adapter_get_builder
+   public :: precice_adapter_get_adapter
+   public :: precice_adapter_interface_t
+   public :: precice_adapter_builder_t
+   
+   logical :: is_enabled = .false.
+   class(precice_adapter_builder_t), allocatable, target :: builder_instance
+   class(precice_adapter_interface_t), allocatable, target :: adapter_instance
```
harmenwierenga, line 17: We talked about these this morning, and you explained that you made this facade to introduce globals and make it easier to introduce this code in FM, which does not pass around any data and accesses all data through globals everywhere. Did you consider introducing a type that can be passed around, and how much impact would that have?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/interface.F90
```diff
@@ -0,0 +1,41 @@
...
+
+   interface
+      subroutine initialize_function_interface(self)
+         import precice_adapter_interface_t
+         implicit none(type, external)
+
+         class(precice_adapter_interface_t), intent(inout) :: self
+      end subroutine initialize_function_interface
+
+      subroutine update_function_interface(self, timestep)
```
harmenwierenga, line 23: There is a single update function now. Are there scenarios where we need to read the preCICE output from a coupled model as input for a time step, but it matters what the time is of that read? As in, does it matter that it happens at the end of the previous time step or at the beginning of the next one?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
```diff
@@ -0,0 +1,104 @@
...
+   type :: precice_adapter_builder_t
+      character(kind=c_char, len=250) :: configfile ! preCICE XML config file path
+      character(kind=c_char, len=20) :: name ! Participant name.
+      integer(kind=c_int) :: my_rank = 0_c_int
+      integer(kind=c_int) :: numranks = 1_c_int
+      integer(kind=c_int) :: comm = 0_c_int
+      logical :: is_comm_set = .false.
+      character(kind=c_char, len=20) :: mesh_name ! mesh name
+      integer(kind=c_int) :: mesh_size = 0_c_int ! mesh size (number of points)
+      real(kind=c_double), dimension(:), pointer :: mesh_coordinates_x ! mesh X coordinates
```
harmenwierenga: I think that these can be allocatable instead of pointer

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+
+      character(kind=c_char, len=*), intent(in) :: configfile
+      character(kind=c_char, len=*), intent(in) :: name
+      logical, intent(in) :: is_comm_set
+      integer(kind=c_int), intent(in) :: comm
+      integer(kind=c_int), intent(in) :: my_rank
+      integer(kind=c_int), intent(in) :: numranks
+      character(kind=c_char, len=*), intent(in) :: meshname
+      integer(kind=c_int), intent(in) :: mesh_size
+      real(KIND=c_double), dimension(:), intent(in), pointer :: mesh_coordinates_x
```
harmenwierenga: lower case kind

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_modelinit.F90
```diff
@@ -615,6 +617,17 @@ integer function flow_modelinit() result(iresult) ! initialise flowmodel
       call timstop(handle_extra(36)) ! End remainder
       call writesomeinitialoutput()
 
+      ! Geometry and initial data available, set up mesh(es) for precice adapter
+      ! TODO: Move name and config to more appropropriate places (also get names from config)
```
harmenwierenga, line 621: There is still a TODO here. Where should these land?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+   use precision, only: dp
+   use, intrinsic :: iso_c_binding, only: c_int, c_char, c_double
+
+   implicit none(type, external)
+
+   private
+   public :: precice_adapter_t
+
+   type, extends(precice_adapter_interface_t) :: precice_adapter_t
+      character(kind=c_char, len=250) :: configfile
```
harmenwierenga: Can you make the fields that do not have a constant name allocatable instead? I see configfile and fm.

Nitpick: configfile -> config_file, meshname -> mesh_name

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+      character(kind=c_char, len=20) :: name = "fm"
+      character(kind=c_char, len=13) :: meshname = "fm_flow_nodes"
```
harmenwierenga: Can you make the fields that are really fixed 'parameter'?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+      character(kind=c_char, len=10) :: bed_levels_name = "bed_levels"
+      character(kind=c_char, len=12) :: water_levels_name = "water_levels"
+      character(kind=c_char, len=12) :: water_depths_name = "water_depths"
+      integer(kind=c_int), dimension(:), allocatable :: vertex_ids
+      logical :: is_comm_set = .false.
+      integer(kind=c_int) :: comm
```
harmenwierenga: comm -> communicator

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+      integer(kind=c_int) :: my_rank = 0_c_int
+      integer(kind=c_int) :: numranks = 1_c_int
+      real(kind=c_double), dimension(:), allocatable :: mesh_coordinates ! Mesh coordinates: x1,y1,x2,y2,...,xN,yN
```
harmenwierenga: mesh_coordinates -> cell_coordinates_2d (?). There will be more meshes, so mesh can be confusing

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+
+   function precice_adapter_constructor(configfile, name, is_comm_set, comm, my_rank, numranks, meshname, &
+                                        mesh_size, mesh_coordinates_x, mesh_coordinates_y) &
+      result(adapter_instance)
+      use precision, only: dp
+      use, intrinsic :: iso_c_binding, only: c_int, c_char, c_double
+
+      implicit none(type, external)
+
+      character(kind=c_char, len=*), intent(in) :: configfile
```
harmenwierenga: Could you consistently break words by underscores? There are more names that are glued together

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+         adapter_instance%mesh_coordinates(2 * i) = mesh_coordinates_y(i)
+      end do
+
+   end function precice_adapter_constructor
+
+   subroutine precice_adapter_destructor(self)
+      type(precice_adapter_t), intent(inout) :: self
+
+      if (allocated(self%mesh_coordinates)) then
+         deallocate (self%mesh_coordinates)
```
harmenwierenga: I wonder if this is necessary, when would the allocatables be deallocated automatically?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+      call precicef_write_data(self%meshname, self%water_depths_name, &
+                               size(self%vertex_ids), self%vertex_ids, &
+                               hs, len(self%meshname), len(self%water_depths_name))
+
+      ! Actually advance time
+      call precicef_get_max_time_step_size(max_timestep)
+      if (max_timestep < timestep) then
+         call precicef_advance(max_timestep)
+         ! ... and now what?
+         call precicef_advance(timestep - max_timestep) ! This works?
```
harmenwierenga: What is the goal here? We could also rely on preCICE to hand back control to FM if the full coupling time step is not finished yet. The current code does not actually simulate the time that is larger than max_time_step

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,169 @@
...
+      ! TODO: Read latest state here ?
+   end subroutine precice_adapter_update
+
+   subroutine precice_adapter_finalize(self)
+      use precice, only: precicef_finalize
+      implicit none(type, external)
+      class(precice_adapter_t), intent(inout) :: self
+
+      if (allocated(self%mesh_coordinates)) then
+         deallocate (self%mesh_coordinates)
```
harmenwierenga: This is probably redundant if mesh_coordinates is turned into an allocatable

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
```diff
@@ -0,0 +1,104 @@
...
+   end type
+
+   interface precice_adapter_builder_t
+      procedure :: precice_adapter_builder_constructor
+   end interface
+
+contains
+
+   function precice_adapter_builder_constructor() result(builder_instance)
+      type(precice_adapter_builder_t), pointer :: builder_instance
```
harmenwierenga, line 37: Does this work with allocatable instead?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
```diff
@@ -0,0 +1,104 @@
...
+      self%comm = comm
+   end subroutine builder_set_mpi_comm
+
+   subroutine builder_add_mesh_2d(self, mesh_name, mesh_size, mesh_coordinates_x, mesh_coordinates_y)
+      use precision, only: dp
+      class(precice_adapter_builder_t), intent(inout) :: self
+      character(len=*) :: mesh_name
+      integer(kind=c_int), intent(in) :: mesh_size
+      real(kind=c_double), dimension(:), intent(in), pointer :: mesh_coordinates_x
+      real(kind=c_double), dimension(:), intent(in), pointer :: mesh_coordinates_y
```
harmenwierenga: Can these two be target instead of pointer? They cannot be null. The current function also assumes that those two are not temporaries, should they be intent(inout) in that case so that the compiler can flag that? What we could also do is to build the 2d x1,y1,x2,y2.... array here already, that way we can make a copy and the cell_faces_2d (what is now called mesh_coordinates_x/y) could be held by value instead of being pointers to global data

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/interface.F90
erwin-mulder: None of the current couplings read or write at a higher time resolution than the user time step, so updating to a new state should suffice? Using the theoretically more detailed and/or interpolated present in preCICE is a future improvement at best.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: The issue here is that this branch is actually only hit if for some reason the preCICE coupling window is not exactly an integer number of user time steps. I was not sure how we should handle these cases without introducing a discrepancy in timekeeping. Dflowfm expects the coupling to advance by the user time step and has no way to deal with any deviations from that.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: You might be right here.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/facade.F90
erwin-mulder: Those data structures would be the builder or adapter instances (where relevant).

erwin-mulder: I would wait with that refactor once it is clear where we (want to) use these objects exactly before altering possibly a host of function signatures.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
erwin-mulder: For some reason my VS2022 insists on automatically inserting the upper case keywords if you don't press escape or something.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
erwin-mulder: Not associated is a programming error, really. I would love to be able to simply put an assert here to check if it is associated. If the program is correct, the result should always be associated. Perhaps we can just remove the check?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
erwin-mulder: I did some testing with this and noticed additional calls to constructor- and destructor-functions, which means that there is probably copying going on, which I don't want.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
erwin-mulder: In a previous version, I did in fact create the x1,y1,x2,y2,.. array, but at some point I needed the x and y coordinates separately again for setting up triangulation. Now that we have removed the idea of doing triangulation on the FM side, we can revert to creating a single preCICE-style array here.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Yes.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/interface.F90
harmenwierenga: I meant splitting up the update into a read_input_data and write_output_data for example. We had a short discussion, and we can always make the choice to break them up later if a specific coupling demands that.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
erwin-mulder: Fixed by removing it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/interface.F90
erwin-mulder: As discussed, we'll cross this bridge when we get there. For now, we assume this setup can be made to work by writing/reading in bulk inside the update call. Any initial data reading/writing probably belongs more in the initialize phase/function.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/unstruc_api.F90
erwin-mulder: As discussed, I removed the if statements as it should technically always be true.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
erwin-mulder: Replaced with allocatable combined x1,y1,x2,y2,... preCICE style mesh coordinate array.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
erwin-mulder: These are no longer pointer in the new situation.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/builder.F90
harmenwierenga: Fortran makes me sad sometimes

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
harmenwierenga: Use the MessageHandling module to stop with an error if this situation occurs. Change the check to see if max_time_step is very close to time_step

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/flow_finalize_usertimestep.f90
```diff
@@ -170,6 +172,13 @@ subroutine flow_finalize_usertimestep(iresult)
          call update_fourier(merge(dt_user, ti_his, md_fou_step == 0))
       end if
 
+      ! Communicate trough precice_adapter if anabled.
+      if (precice_adapter_is_enabled()) then
+         print *, "[DEBUG] Updating preCICE adapter."
```
harmenwierenga: Use MessageHandling

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_modelinit.F90
erwin-mulder: Wherever we retrieve these settings? Form an adapter config file or the command line. To be determined.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: In the end none are going to remain fixed, but made the items set in the constructor allocatable

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed, maximum difference is now set to 1e-5

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: compiler complains about trying to make fields parameter.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: I am thinking we may need to change this into some sort of a list of meshes? I think we may have multiple meshes of different types in the end. Or should we name each one individually?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/flow_finalize_usertimestep.f90
erwin-mulder: Fixed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
harmenwierenga: I think that naming each one individually would be more readable

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -0,0 +1,153 @@
...
+   use precision, only: dp
+   use, intrinsic :: iso_c_binding, only: c_int, c_char, c_double
+
+   implicit none(type, external)
+
+   private
+   public :: precice_adapter_t
+
+   type, extends(precice_adapter_interface_t) :: precice_adapter_t
+      character(kind=c_char, len=250), allocatable :: config_file
```
harmenwierenga: Ah, you made them allocatable, but they are still of fixed length instead of assumed length (len=*)

harmenwierenga (commented):
There are still two small points open, we can talk about whether to address them or to merge now

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: Fixed with len=: instead of len=*.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: renamed mesh related items (including setter build function to a relevant variant of `cell_center_mesh_*_2d`.

harmenwierenga (approved):

PR #753, opened 2026-03-30
Add Submit option to reuse existing verschilanalyse output

robin-deltares:
This pull request introduces a new mechanism to optionally skip running model executions and instead reuse previously archived outputs in the Verschilanalyse pipeline. The changes add a `run_models` parameter throughout the TeamCity configuration and pipeline scripts, allowing users to control whether models are executed or existing outputs are used.

**Parameterization and Pipeline Control:**

* Added a `run_models` checkbox parameter to the TeamCity build configuration (`StartVerschilanalyse`) and propagated it through the environment and command-line arguments, enabling users to decide whether to run models or reuse existing output.

**Pipeline Logic Updates:**

* Updated `start_verschilanalyse.sh` to branch execution based on the `RUN_MODELS` flag: if true, models are executed and output is uploaded; if false, existing output is downloaded and reused. Job dependencies are adjusted accordingly.

**New Script for Output Reuse:**

* Added `download_current_output.sh`, a Slurm job script that downloads and unpacks archived model outputs from S3, supporting the reuse workflow.

### DIFF
--- a/ci/teamcity/Delft3D/verschilanalyse/StartVerschilanalyse.kt
+++ b/ci/teamcity/Delft3D/verschilanalyse/StartVerschilanalyse.kt
@@ -44,6 +44,15 @@ object StartVerschilanalyse : BuildType({
         param("model_filter", "")
         param("json_configs_path", "config")
         checkbox(
+            "run_models",
+            "true",
+            display = ParameterDisplay.NORMAL,
+            label = "Run models on H7",
+            description = "Run models on Slurm before running Verschillentool. Disable to reuse existing output at current_prefix.",
+            checked = "true",
+            unchecked = "false",
+        )
+        checkbox(
             "send_email",
             "true",
             display = ParameterDisplay.NORMAL,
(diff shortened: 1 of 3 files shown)


ahmad-el-sayed (commented):
My comments are completely optional. I think the usage of a switch is a bit more elegant but absolutely not necessary.

ahmad-el-sayed (changes requested):

ahmad-el-sayed (approved):
@robin-deltares  Approving because it works and it is handy to have ASAP. To revisit and refactor/improve next sprint.

ahmad-el-sayed (commented):
A few thing to consider while working on DEVOPSDSC-950:

1. Both download scripts do not check if zip files have been found prior to unzipping. this can happen if syncing fails, die to a wrong current or reference minio path.
2. it is safer to set sbatch dependencies when downloading the current data. Is it stands this can take place simultaneously with the download of the reference data. Luckily the run docker containers do not share common volumes.

PR #782, opened 2026-04-08
All/feature/unst 9649 ff2 nf writer

harmenwierenga:
# What was done 

- Create data structures for the data that needs to be written to the FF2NF files
- Create a writer class that accepts the data and then writes an xml file from it
- Properly indent the data in the XML file
- Add unit tests for all xml writing functionality
- Refactor some xml parsing utilities such that they can be reused by the tests
- Add the core logic to the coupling_steps for writing the FF2NF files
- Skip writing a copy of the C-SUMO settings XML tree to the FF2NF file, we can add it if C-SUMO really needs it

# Evidence of the work done 

- [ ]	Video/figures \
- [x]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link UNST-9746


src/tools_gpl/pre_c_sumo/include/parsing_types.hpp
```diff
@@ -0,0 +1,26 @@
...
+     */
+    struct [[nodiscard]] ParseError
+    {
+        std::string message{};
+    };
+
+    /**
+     * @brief A 2-D coordinate pair (x, y).
+     */
+    struct [[nodiscard]] Point2D
```
erwin-mulder: Is adding [[nodiscard]] required here if we also have the habit of making all our functions [[nodiscard]] too? Feels a bit like wearing suspenders and a belt.

src/tools_gpl/pre_c_sumo/include/monadic_utils.hpp
erwin-mulder, line 1: This whole construct has a very high probability of becoming 'that thing we never touch'.

erwin-mulder (approved):
Looks good, only some comments/questions regarding choices made.

src/tools_gpl/pre_c_sumo/include/parsing_types.hpp
harmenwierenga: I think that adding it to the functions is the nicest indeed, relying on the type versions can be inconsistent. Some types don't have it, like std::vector, so functions returning a vector need it. Plus it belongs to the function, I will remove it here again.

src/tools_gpl/pre_c_sumo/include/monadic_utils.hpp
harmenwierenga: I agree that the macros are complex. I do think that it makes the rest of our code much cleaner, so I personally think that it's worth it. If the language ever gets a try construct for std::expected or std::optional, then we can get rid of the macros.
I am less afraid of the concepts (IsExpected etc.) and predicates (is_valid etc.), since they are pretty standard C++ and also make the rest nice and clean.

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
```diff
@@ -0,0 +1,103 @@
...
+    struct FF2NFConfig
+    {
+        std::string ff2nf_filename;
+        std::string wait_for_file;
+        std::string ff_run_directory;
+        std::string run_id;
+        std::string unique_id;
+        int subgrid_model_nr{};
+        double current_time_seconds{};
+        std::vector<std::string> constituent_names;
```
AabhasSrivastava, line 55: can we make a note that we have similar way of addressing constituents in both directions of data?

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
```diff
@@ -0,0 +1,108 @@
...
+    {
+        std::string message;
+    };
+
+    /**
+     * @brief Far-field layer at a single horizontal point.
+     */
+    struct FarFieldLayer
+    {
+        double depth_from_surface{}; ///< Depth of the layer from the water surface [m].
```
adrimourits: The (FarField and NearField) applications share the same coordinate system: x, y AND z. I expect here z_coordinate instead of depth_from_surface.

src/tools_gpl/pre_c_sumo/include/parsing_types.hpp
```diff
@@ -0,0 +1,26 @@
...
+    };
+
+    /**
+     * @brief A 2-D coordinate pair (x, y).
+     */
+    struct Point2D
+    {
+        double x{};
+        double y{};
+    };
```
AabhasSrivastava, line 23: Why can't Point2D be used as constituent of FarfieldPoint2D?

AabhasSrivastava: not saying this is incorrect, but if we work with points (functions) might be easier to just keep to one type consistently. I would also then change it properly on the Endpoint data structure.

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
```diff
@@ -0,0 +1,108 @@
...
+        double x_velocity{};         ///< X velocity in this layer [m/s].
+        double y_velocity{};         ///< Y velocity in this layer [m/s].
+    };
+
+    /**
+     * @brief Far-field state at a single horizontal point (diffuser, intake, or ambient).
+     *
+     * @note density and constituents are depth-averaged scalars (per node, not per layer);
```
adrimourits: I don't think this repeating of depth-averaged scalars is intended. Can you move them to FarFieldLayer?

src/tools_gpl/pre_c_sumo/include/monadic_utils.hpp
```diff
@@ -43,14 +43,22 @@ namespace monadic_utils::internal
 
 } // namespace monadic_utils::internal
```
adrimourits, line 53: Can you add some documentation? What is it, why is it needed.

src/tools_gpl/pre_c_sumo/src/parsing_utils.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    std::filesystem::path normalizePath(std::string path)
+    {
+        std::replace(path.begin(), path.end(), '\\', '/');
+        path.erase(std::find_if_not(path.rbegin(), path.rend(), [](const char c) { return c == '/'; }).base(),
+                   path.end());
+        return std::filesystem::path(std::move(path));
+    }
+} // namespace parsing_utils
```
adrimourits: Missing: return at and of file.

src/tools_gpl/pre_c_sumo/test/test_FF2NF_writer.cpp
adrimourits, line 1: Nice setup of the tests. It gives confidence.

src/tools_gpl/pre_c_sumo/src/FF2NF_writer.cpp
adrimourits, line 1: I'm a bit worried about the organization of the subroutines/methods: There are a lot floating around, executing their task on different hierarchical levels. Can you think about a way to structure this? This doesn't need to be resolved as part of this issue.

adrimourits (commented):
My comments are mostly generic; you can neglect them in this issue.

src/tools_gpl/pre_c_sumo/src/parsing_utils.cpp
harmenwierenga: I fixed this yesterday I think

src/tools_gpl/pre_c_sumo/include/parsing_types.hpp
harmenwierenga: They did not interact before, but I can also reuse it, I have added the parsing_utils::Point2D here now.

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
harmenwierenga: The Delft3D4 code does write densities and constituents per layer, but the bmi code does not. I will change it to reflect the intention of writing depth-varying densities and constituent concentrations

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
harmenwierenga: Done

src/tools_gpl/pre_c_sumo/include/monadic_utils.hpp
harmenwierenga: Done

src/tools_gpl/pre_c_sumo/src/FF2NF_writer.cpp
harmenwierenga: We can refactor the more generic functions into a separate utilities library when we need to read/write more xml files outside of preC-SUMO. We agreed that we will leave the structure as-is for now

src/tools_gpl/pre_c_sumo/include/FF2NF_writer.hpp
AabhasSrivastava: Note, maybe in the documentation, code comment or Jira issue/epic.I'll resolve this conversation for now.

AabhasSrivastava: added in JIRA issue UNST-9756

adrimourits (approved):

PR #790, opened 2026-04-10
UNST-9842 set the stack size of OpenMP helper threads to 20MB, just l…

harmenwierenga:
…ike the main thread on windows

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link UNST-9842

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm-cli_exe/src/net_main.F90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm-cli_exe/src/net_main.F90
@@ -204,9 +204,7 @@ program unstruc
 
    call iset_jaopengl(md_jaopengl)
 
-#ifdef _OPENMP
-      ierr = init_openmp(md_numthreads, jampi)
-#endif
+   ierr = init_openmp(md_numthreads, jampi)
 
    call START_PROGRAM()
    call resetFullFlowModel()
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/init_openmp.F90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/init_openmp.F90
@@ -66,6 +66,12 @@ contains
 #endif
 
 #ifdef _OPENMP
+      ! Set OpenMP worker thread stack size to 20 MB.
+      ! Must be called before the first parallel region.
+      ! On Windows, the default (4 MB for Intel OpenMP) is too small for Fortran
+      ! subroutines with large automatic arrays, causing random stack overflows.
+      call kmp_set_stacksize_s(20 * 1024 * 1024)
+
       if (mpion == 1 .and. maxnumthreads == 0) then
          ! If MPI is on for this model, *and* no user-define numthreads was set, then disable OpenMP.
          openmp_threads = 1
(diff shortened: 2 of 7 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/init_openmp.F90
```diff
@@ -66,6 +66,12 @@ integer function init_openmp(maxnumthreads, mpion) result(iresult)
 #endif
 
 #ifdef _OPENMP
+      ! Set OpenMP worker thread stack size to 20 MB.
+      ! Must be called before the first parallel region.
+      ! On Windows, the default (4 MB for Intel OpenMP) is too small for Fortran
+      ! subroutines with large automatic arrays, causing random stack overflows.
+      call kmp_set_stacksize_s(20 * 1024 * 1024)
```
FlorisBuwaldaDeltares: this should be a parameter, and ideally one inherited from CMake if possible (where we also set the stack size for the main executable)

FlorisBuwaldaDeltares (changes requested):
functionally correct but I think if we don't inherit the value from somewhere it is easy to forget for future developers that the omp threads have a separate stacksize. Would it be possible to inherit the cmake value?

src/engines_gpl/dflowfm/packages/dflowfm-cli_exe/src/net_main.F90
```diff
@@ -204,9 +204,7 @@ program unstruc
 
    call iset_jaopengl(md_jaopengl)
 
-#ifdef _OPENMP
```
harmenwierenga, line 207: _OPENMP was only defined on windows, so init_openmp was never called on linux through the CLI.

src/engines_gpl/dflowfm/packages/dflowfm_lib/src/unstruc_bmi.F90
```diff
@@ -290,14 +291,13 @@ integer(c_int) function initialize(c_config_file) result(c_iresult) bind(C, name
          jampi = 0
       end if
 
-#ifdef _OPENMP
-      ierr = init_openmp(md_numthreads, jampi)
```
harmenwierenga, line 294: This was never called, because _OPENMP was not defined on windows nor on linux (proved by the absence of the ierr declaration, it could not even compile if _OPENMP were defined).

I also moved it out of the ifdef MPI block, it has nothing to do with mpi.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/init_openmp.F90
harmenwierenga: Done

harmenwierenga: Done. One small gotcha can be that this parameter is in bytes, while the environment variable OMP_STACKSIZE has a default unit of KB (but there, you can specify the unit if you want). I have also wrapped it in a macro to check for Intel, because the function is an Intel extension.

FlorisBuwaldaDeltares (approved):
nice improvements!

PR #792, opened 2026-04-10
All/feature/unst 9819 dsle constituents

FlorisBuwaldaDeltares:
# What was done 

- added constituent calculation to DSLE
- constituents are registered by name, defined by dimr_xml
- enabled temperature output by treating it as a passive constituent
- added bmi unit tests and sealock unit tests
- phase-wise computation separately, updates concentrations after each phase-wise update based on volumes
- old computation (mixing fraction * concentrations) remains valid for cycle-average computations

# Evidence of the work done 

- [ ]	Video/figures \
Before:

image

After:

image

Tracer results:

image

- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
p:\dflowfm\maintenance\JIRA\09000-09999\09765\e119_f001_c011_3D_pw_2sl_uic_cbc_with_tracer\
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dsle/packages/tests/test_sealock.c
```diff
@@ -397,6 +504,361 @@ static void test_sealock_delta_time_ok__diff_eq_delta_time__not_ok(void) {
...
+  // Boundary conditions.
+  lock.parameters.head_lake = 0.0;
+  lock.parameters.head_sea = 1.0;
+  lock.parameters3d.salinity_lake[0] = 0.0;
+  lock.parameters3d.salinity_sea[0] = 30.0;
+
+  // Restore salinity state (sealock_init may have modified it).
+  lock.phase_state.salinity_lock = 15.0;
+  lock.phase_state.saltmass_lock = 15.0 * lock.parameters.lock_length * lock.parameters.lock_width *
+                                   (1.0 - lock.parameters.lock_bottom);
```
sumihar: This 1.0 here is water level in the lock, I suppose. Perhaps we can use lock.phase_state.head_lock here instead?

src/engines_gpl/dsle/packages/tests/test_sealock.c
```diff
@@ -397,6 +504,361 @@ static void test_sealock_delta_time_ok__diff_eq_delta_time__not_ok(void) {
...
+
+  // Slot 1 mirrors salinity exactly.
+  lock.constituent_lock[1] = lock.phase_state.salinity_lock;                     // 15.0
+  lock.parameters3d.constituent_lake[1][0] = lock.parameters3d.salinity_lake[0]; // 0.0
+  lock.parameters3d.constituent_sea[1][0] = lock.parameters3d.salinity_sea[0];   // 30.0
+
+  // Act + Assert: after each phase, constituent_lock[1] must equal salinity_lock.
+  for (int i = 0; i < 4; i++) {
+    TEST_ASSERT_EQUAL(SEALOCK_OK, sealock_update(&lock, times[i]));
+    TEST_ASSERT_DOUBLE_WITHIN(1e-9, lock.phase_state.salinity_lock, lock.constituent_lock[1]);
```
sumihar, line 1004: This is a very smart approach: making sure that constituent is calculated the same way as salinity, given identical initial and boundary conditions of the two. Nice!

I think we still need to add the following checks:

1. phase step_flush_doors_closed, i.e. flushing when doors are closed. In the input csv file, it is identified as phase -2 or -4
2. phase 2 and 4, where flushing is involved. The test example doesn't contain this scenario (https://gitlab.com/deltares/libzsf/-/blob/master/examples/python/lockages.csv?ref_type=heads), but the code step_phase_2 and step_phase_4 in dsle.c contain flushing. When flushing is involved, it becomes somewhat more complicated. For example, in step_phase_2:
             double mt_sea_2_flushing =
		            volume_flush_refresh * sal_lock_2a + volume_flush_passthrough * p->salinity_lake;
		     double mt_to_sea_2b = mt_sea_2_flushing;
		     double saltmass_lock_2b = saltmass_lock_2a + mt_from_lake_2b - mt_to_lake_2b - mt_to_sea_2b;
This shows that volume water that goes to sea has two components: volume_flush_refresh and volume_flush_passthrough, each with its own salinity. In this case, we can't assume that all water that goes to sea has a salinity that equals to the lock salinity.

sumihar (commented):
Hi Floris,
I have a few comments. Please have a look.
Many thanks,
Julius

src/engines_gpl/dsle/packages/tests/test_sealock.c
FlorisBuwaldaDeltares: good catch! I will change it

src/engines_gpl/dsle/packages/tests/test_sealock.c
FlorisBuwaldaDeltares: ah ok.. thanks for checking. I'll investigate it

FlorisBuwaldaDeltares: the flush volume is now (hopefully) correctly taken into account for steps 2, 4 and -1

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -603,6 +714,51 @@ static int sealock_distribute_results(sealock_state_t *lock) {
   return SEALOCK_OK;
 }
 
+static int sealock_distribute_constituent_results(sealock_state_t *lock) {
```
sumihar: - There is no unit test yet available for the routine, hence we can't check if the output is correct. But I guess the results will be different from what you would get from running a cycle average with constituent. It misses here for example the volume_from_lake and volume_from_sea, which are important in determining the average concentration. To be consistent with dsle.c, the formula will be something like: constituent_to = (mt + constituent_from * volume_from) / volume_to; but we need a correct mt (mass transfer) of the constituent over all the four sea-lock phases. I haven't found a handy way of computing it.
- Concentration will be identical over layers (where discharge is non-zero), hence the loop over layers may be omitted.
- This routine is now used only for cycle-average. Perhaps rename to something with cycle-average and run it only for cycle-average mode?

src/engines_gpl/dsle/packages/tests/test_sealock.c
```diff
@@ -834,29 +963,91 @@ static void test_sealock_update__phase_wise__constituent_tracks_salinity_across_
...
+  lock.parameters3d.constituent_lake[1][0] = lock.parameters3d.salinity_lake[0];
+  lock.parameters3d.constituent_sea[1][0] = lock.parameters3d.salinity_sea[0];
+
+  for (int i = 0; i < 2; i++) {
+    TEST_ASSERT_EQUAL(SEALOCK_OK, sealock_update(&lock, times[i]));
+    TEST_ASSERT_DOUBLE_WITHIN(1e-9, lock.phase_state.salinity_lock, lock.constituent_lock[1]);
+  }
+
+  // Sanity: flushing must have reduced salinity meaningfully from initial 15.0.
+  TEST_ASSERT_TRUE(fabs(lock.phase_state.salinity_lock - 15.0) > 1e-6);
```
sumihar, line 1102: Can we use TEST_ASSERT_EQUAL or TEST_ASSERT_DOUBLE_WITHIN here with the correct salinity_lock? It is more sensitive to (unwanted) changes. There may be code changes in the future that cause salinity_lock to be incorrect, but it is still very different from 15.0. In that case, it will incorrectly pass this TEST_ASSERT_TRUE.

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: As far as I understand, we don't need to follow dsle, my formula is also exact. I've added a unit test that demonstrates also that a constituent with the same initial condition as the salinity, follows the salinity exactly for the cycle-average computation.

I don't think we can omit the loop over layers, as we still have to fill the values even if they're identical.

i've renamed the routine and it is now also called for cycle average.

Can you design a unit test that shows divergence between constituent mirror and salinity for my cycle average computation?

src/engines_gpl/dsle/packages/tests/test_sealock.c
FlorisBuwaldaDeltares: I can't get pretty numbers in there, but sure I guess it's better than an assertion that the value changed. On the other hand, this unit tests doesn't test if the salinity value is correct, only that it changed and that the constituent has the same result, so for those purposes it is sufficient

src/engines_gpl/dsle/packages/src/sealock.c
sumihar: I have checked again how cycle-average works. I think in this mode, all information about volume to/from, required to determine final concentration, is also contained in salinity_to and salinity_from. You can indeed reuse the salinity mixing fraction in this case for other constituents.

src/engines_gpl/dsle/packages/tests/test_sealock.c
sumihar: O, oké, the check on values are done on the previous lines indeed. So, this extra check is sufficient in this form.

src/engines_gpl/dsle/packages/tests/test_sealock.c
```diff
@@ -397,6 +504,646 @@ static void test_sealock_delta_time_ok__diff_eq_delta_time__not_ok(void) {
...
+
+  // Slot 1 mirrors salinity exactly.
+  lock.constituent_lock[1] = lock.phase_state.salinity_lock;                     // 15.0
+  lock.parameters3d.constituent_lake[1][0] = lock.parameters3d.salinity_lake[0]; // 0.0
+  lock.parameters3d.constituent_sea[1][0] = lock.parameters3d.salinity_sea[0];   // 30.0
+
+  // Act + Assert: after each phase, constituent_lock[1] must equal salinity_lock.
+  for (int i = 0; i < 4; i++) {
+    TEST_ASSERT_EQUAL(SEALOCK_OK, sealock_update(&lock, times[i]));
+    TEST_ASSERT_DOUBLE_WITHIN(1e-9, lock.phase_state.salinity_lock, lock.constituent_lock[1]);
```
sumihar, line 1004: Hi Floris,
perhaps good to add here a check on "constituent_to", e.g. lock.results3d.constituent_to_sea[1][i]. I have tried it, and it seems that sealock_update() doesn't update the lock results.

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -603,6 +710,51 @@ static int sealock_distribute_results(sealock_state_t *lock) {
   return SEALOCK_OK;
 }
 
+static int sealock_distribute_cycle_average_constituent_results(sealock_state_t *lock) {
+  // Phase-wise mode: constituent outputs are written directly by
+  // sealock_step_constituents using the pre-phase lock concentration.
+  // Results persist between updates when no phase transition occurs.
+  if (lock->computation_mode == phase_wise_mode) {
+    return SEALOCK_OK;
+  }
```
sumihar: This block is no longer required, I suppose? Hence, we can remove those lines. This routine will be called only in cycle-average mode.

sumihar (commented):
Hi, Floris

I think that your approach works fine. Cool that we could do it that way, i.e. adding constituents in the "wrapper" of dsle instead of in the kernel and that there is no need to prepare a list of constituents in dsle module. Very nice!

I still have just two small final comments. Please have a look.

Many thanks,
Julius

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: I have removed it :)

src/engines_gpl/dsle/packages/tests/test_sealock.c
FlorisBuwaldaDeltares: as we discussed, we must compare results3d.constituent to results.salinity and not results3d.salinity, as results3d.salinity is masked and the constituents are not.

src/engines_gpl/dsle/packages/src/log/log.c
```diff
@@ -16,6 +16,9 @@ void log_init(char *name, FILE *fp) {
 void log_set_level(log_level_t lvl) { logger.level = lvl; }
 
 static void vlog(log_level_t lvl, char *fmt, va_list ap) {
+  if (logger.name == NULL) {
```
rene-deltares: Not sure if this is safe. `logger` is a global variable. If it is not initialized I guess `logger.name` is `NULL` by accident because it's stored in the `.bss` section, which is initialized to zeroes when the program is loaded. But I'm not sure if this is true on both windows and linux. Best not rely on it and just make sure that `log_init` is called before ever calling `vlog`.

src/engines_gpl/dsle/packages/src/dimr_bmi.c
```diff
@@ -145,6 +170,18 @@ int set_var(const char *key, void *src_ptr) {
     dest_ptr = &config.locks[lock_index].parameters.temperature_lake;
   } else if (match_key(quantity, "temperature_sea")) {
     dest_ptr = &config.locks[lock_index].parameters.temperature_sea;
+  } else if (strncmp(quantity, "constituent_lake_", 17) == 0) {
+    int c = find_or_register_constituent(&config.locks[lock_index], quantity + 17);
+    if (c < 0)
+      return DIMR_BMI_FAILURE;
```
KoenBussemakerDeltares: Either go for `if (c < 0) return DIMR_BMI_FAILURE` or write out the if over multiple lines, with brackets. The current form is considered the most error-prone one. Imagine you change the code to:

```
if (c < 0)
    do_something();
    return DIMR_BMI_FAILURE;
```

The last line is no longer part of the if statement, although the indenting might make you believe that. This code will always return a `DIMR_BMI_FAILURE`.

In this case I would just go for the inline ifs

src/engines_gpl/dsle/packages/src/dimr_bmi.c
```diff
@@ -91,6 +91,26 @@ void copy_key(const char *src, char *dst) {
 // Checks if a previously retrieved (sub) key matches a defined key.
 inline static int match_key(char *key, char *defined_key) { return !strcmp(key, defined_key); }
 
+// Looks up 'name' in the lock's constituent name registry.
+// Registers it in the next free slot if not found.
+// Returns the constituent index, or -1 when MAX_NUM_CONSTITUENTS is exceeded.
+static int find_or_register_constituent(sealock_state_t *lock, const char *name) {
+  for (unsigned int c = 0; c < lock->num_constituents; c++) {
```
rene-deltares, line 107: ~Just `int c = 0` is fine. Then you don't have to cast.~

rene-deltares: Ah, nevermind. `lock->num_constituents` is also unsdigned, which is consistent with other `num_` struct members. Perhaps change the return type to an `unsigned int` as well, in that case?

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -65,9 +65,16 @@ int sealock_init(sealock_state_t *lock, time_t start_time, unsigned int max_num_
                                    lock->phase_state.salinity_lock, lock->phase_state.head_lock);
   }
 
+  if (status == SEALOCK_OK) {
+    // Register temperature at slot 0 (TEMPERATURE_CONSTITUENT_SLOT) unconditionally.
+    // User tracers arriving later via BMI set_var will occupy slots 1+.
+    lock->constituent_names[TEMPERATURE_CONSTITUENT_SLOT] = "temperature";
```
KoenBussemakerDeltares, line 71: A word of warning here. C doesn't have strings, only character arrays, which are somewhat similar to what you're used to in fortran. But it's a lot easier to shoot yourself in the foot in C!

You might think you are doing a string assignment here. What you are doing here is simply storing a pointer to a so-called string literal. String literals have static storage duration so they will exist for the lifetime of the application, so technically this is ok. However, according to `sealock.h` you store these as non-const `char *`, which means you can modify the memory of the string literal if you wanted to. And that happens to be undefined behavior in C / C++.

To fix this, make `constituent_names` in `sealock.c` a `char *constituent_names[MAX_NUM_CONSTITUENTS];`. Following the rule of thumb that is reading from right to left, you'll see that this means "a pointer to a char that is const".

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -223,6 +230,91 @@ static int sealock_apply_phase_wise_result_correction(sealock_state_t *lock, tim
...
+      c_to_sea = tp->volume_to_sea > DBL_EPSILON
+                     ? (c_lock * (tp->volume_to_sea - tp->volume_flush_passthrough) +
+                        c_lake * tp->volume_flush_passthrough) /
+                           tp->volume_to_sea
+                     : c_lock;
+    }
+
+    lock->constituent_lock[c] = c_lock_new;
+
+    for (unsigned int i = 0; i < to_lake_volumes->num_active_cells; i++)
```
KoenBussemakerDeltares: Wrap these for loops in `{}` and spread them out over multiple lines. Inline if statements are often fine but inline for loops tend to be hard to read.

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -470,6 +567,16 @@ static int sealock_collect_layers(sealock_state_t *lock) {
     lock->parameters.salinity_sea = sealock_collect(sea_volumes, lock->parameters3d.salinity_sea);
   }
 
+  // Temperature always occupies slot 0. Inject the scalar temperature_lake/sea values
+  // (kept current via set_var) into every layer of the constituent arrays so that
+  // sealock_step_constituents_phase_wise and sealock_distribute_cycle_average_constituent_results see consistent data.
+  for (unsigned int i = 0; i < lock->from_lake_volumes.num_volumes; i++) {
+    lock->parameters3d.constituent_lake[TEMPERATURE_CONSTITUENT_SLOT][i] =
```
KoenBussemakerDeltares: Too much line wrapping. Monitors are wide these days.

src/engines_gpl/dsle/packages/src/dimr_bmi.c
```diff
@@ -145,6 +170,18 @@ int set_var(const char *key, void *src_ptr) {
     dest_ptr = &config.locks[lock_index].parameters.temperature_lake;
   } else if (match_key(quantity, "temperature_sea")) {
     dest_ptr = &config.locks[lock_index].parameters.temperature_sea;
+  } else if (strncmp(quantity, "constituent_lake_", 17) == 0) {
```
rene-deltares: These string length constants make me a feel a bit uneasy. I'd prefer `strlen` and relying on the compiler to replace it by a constant

Since we have a nice `match_key` function, I was thinking, maybe add a `match_key_prefix` function that returns the name of the constituent. Something like this perhaps?
```c
static char * match_key_prefix(char * quantity, char *prefix) {
    int length = strlen(prefix);
    if (strncmp(quantity, prefix, length) == 0) {
        return &quantity[length];
    }
    return NULL;    
}
```

Then you can use it like this:
```c
} else if (constituent = match_key_prefix(quantity, "constituent_lake_")) {
    int c = find_or_register_constituent(&config.locks[lock_index], constituent);
```
Although maybe people object to the assignment in the condition.

src/engines_gpl/dsle/packages/src/log/log.c
FlorisBuwaldaDeltares: I really don't like log_init being mandatory to call, the reason i changed this because the previous people who worked on this code forgot it and the log was flooded with unnecessary debug logging even in release. I've now added a default constructor, is that ok?

src/engines_gpl/dsle/packages/src/dimr_bmi.c
FlorisBuwaldaDeltares: will do!

src/engines_gpl/dsle/packages/src/dimr_bmi.c
FlorisBuwaldaDeltares: ok!

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: okay, it's now a const char * in the header :) it's a bit abracadabra for me but copilot says that we now have compiler safety instead of relying on legacy behaviour

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: disagree with you here, simple loop from 0 to num_active_cells, we have similar stuff in our fortran do loops?

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: ah well spotted, indeed!

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: oh wait you just meant the brackets, yes sorry i've added them

src/engines_gpl/dsle/packages/src/sealock.c
KoenBussemakerDeltares: "Trust me bro" ;)

C memory management is not something you get the hang of overnight.

src/engines_gpl/dsle/packages/src/dimr_bmi.c
FlorisBuwaldaDeltares: i don't object :) i've implemented your suggestion

KoenBussemakerDeltares: Since you aren't modifying anything (and you don't want to) this should be:

`static const char * match_key_prefix(const char * quantity, const char *prefix) { ...`

FlorisBuwaldaDeltares: fixed!

src/engines_gpl/dsle/packages/src/dimr_bmi.c
```diff
@@ -341,13 +411,16 @@ int get_var_shape(char *key, int dims[DIMR_BMI_MAXDIMS]) { // dims -> int[6]
     source_len = config.locks[lock_index].to_sea_volumes.num_volumes;
   } else if (match_key(quantity, "salinity_to_sea")) {
     source_len = config.locks[lock_index].to_sea_volumes.num_volumes;
-  } else if (match_key(quantity, "water_volume_from_lake")) {
+  } else if (match_key(quantity, "water_volume_to_sea")) {
+    source_len = config.locks[lock_index].to_sea_volumes.num_volumes;
+  } else if (strncmp(quantity, "constituent_lake_", 17) == 0 ||
```
rene-deltares: I'd replace these with `match_key_prefix(quantity, "whatever")` still

src/engines_gpl/dsle/packages/src/dimr_bmi.c
```diff
@@ -91,6 +91,35 @@ void copy_key(const char *src, char *dst) {
...
+      return c;
+    }
+  }
+  if (lock->num_constituents >= MAX_NUM_CONSTITUENTS-1) {
+    log_error("find_or_register_constituent: exceeded MAX_NUM_CONSTITUENTS (%d) registering '%s'\n",
+              MAX_NUM_CONSTITUENTS, name);
+    return -1;
+  }
+  unsigned int c = lock->num_constituents++;
+  lock->constituent_names[c] = strdup(name); // owns the string, safe after keystr goes out of scope
```
rene-deltares, line 118: We're kind of leaking the copied string here. But I'm guessing it is fine because `lock` is global, so it never goes out of scope and is only destroyed when the program ends. So I guess I'll let it slide.

rene-deltares (changes requested):
Small changes requested, but otherwise looks good to me although I'm not sure what the code is doing, so I'm leaning on the other reviewers a little bit.

sumihar (approved):
I agree with the approach taken here: using variable result generated by dsle kernel to reconstruct exchanges of temperature and other constituents. Unit tests have been developed to show that the approach works.

src/engines_gpl/dsle/packages/src/dimr_bmi.c
FlorisBuwaldaDeltares: oops

FlorisBuwaldaDeltares: forgot about that one, fixed!

rene-deltares (approved):
Looks good to me

PR #805, opened 2026-04-14
DELWAQ-1242 Retrieving Latitude from D-Flow FM when using online DELWAQ

michelle-deltares:
# What was done 

- For spherical models, the latitude of faces/cells is made available as a segment function in integrated FM-Delwaq calculations when Latitude is mentioned in the sub-file

# Evidence of the work done 

# Tests 
- [x] New test added: c116_latitude_from_fm_to_waq

# Documentation  
- [x]	Documentation updated \

# Issue link
https://issuetracker.deltares.nl/browse/DELWAQ-1242


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_wq_processes.f90
```diff
@@ -121,6 +121,7 @@ module m_fm_wq_processes
    real(hp), allocatable, dimension(:, :) :: waqoutputs !< waq outputs, dim(noout,Ndkx)
 
    integer :: isfsurf !< pointer to surface         segment function
+   integer :: isflat !< pointer to latitude         segment function
```
arjenmarkusdeltaresnl: I suggest using the name "isflatitude" - now I read the name as "is flat", whereas it is not zero, if the model is NOT flat.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waq/fm_wq_processes.f90
```diff
@@ -492,6 +494,22 @@ module subroutine fm_wq_processes_ini_proc()
...
 
+      icon = index_in_array(clatitude, coname_sub)
+      if (icon > 0) then
+         if (jsferic == 0) then
+            call mess(LEVEL_INFO, '''face (cell) latitude'' not connected, because model is not spherical.')
+            isflat = 0
+         else
+            num_spatial_time_fuctions = num_spatial_time_fuctions + 1
+            isflat = num_spatial_time_fuctions
+            call realloc(sfunname, num_spatial_time_fuctions, keepExisting=.true., fill='latitude')
```
arjenmarkusdeltaresnl, line 505: Why not reuse the paramter clatitude?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_wq_processes.f90
michelle-deltares: Are you telling me that earth isn't flat ;-).
I was in doubt, and I did notice it had become 'flat', but didn't notice the consequence of the combination with 'is'. :-D. I'll update this.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waq/fm_wq_processes.f90
michelle-deltares: Oh, that is a nice suggestion. That could be don on other locations as well, but I think not always... I think I had some issues with string comparison, which I resolved by using a parameter.
I think this has to do with the function call... Since I don't want to use the parameter only for this one, and also do not want to change it everywhere else, I leave it like it is. I'll focus on the weird fact that two unrelated testcases are now failing. Restart runs, while the 'normal' version have no issue....

arjenmarkus (approved):
LGTM

arjenmarkus:
I have no idea why this still needs an approval, as I approved it just now myself.

arjenmarkusdeltaresnl (approved):
LGTM

PR #815, opened 2026-04-17
All/bugfix/unst 9861 xmltree memory leaks

harmenwierenga:
# What was done

- Fix small issue where inquire accesses trailing whitespace in Fortran code
- Update Dimr::GetInstance to use proper singleton pattern with static local variable instead of an allocated instance that is never deallocated
- Only delete Dimr::log when it owns it, not when it refers to the DimrExe::log
- Inside xmltree, use std::string over allocated char*. Make the fields private that can be private. 
- Delete log before deleting clock, since log references clock.
- Fix bug where workingDir was assigned to a pointer to pointer cast to a char * (garbage): newComp->workingDir = (char*)&curPath;
- Make workingDir an owned and allocated char *, so that it can always be deleted safely
- Free the components[i].processes, which were leaking
- Free the computeTimes when allocated, and ensure that they are nullptr otherwise. We cannot easily use a managed type, because dimr_control_block is always malloced.
- Store the charDataStr (xmltree) in a state variable of the library, instead of in global data, so that there is no global std::string with constructors/destructors that can cause issues
- Properly delete all children of the XML nodes

# Evidence of the work done 

- [ ]	Video/figures \
- [x]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link UNST-9861


adrimourits (approved):
Nice!

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -184,20 +189,31 @@ Dimr::~Dimr(void)
     // to do:  (void) FreeLibrary(handle);
     freeLibs();
 
-    log->Write(DEBUG, my_rank, "dimr shutting down normally");
+    if (logIsOwned)
```
KoenBussemakerDeltares: Why do you want to do this conditionally? Shutdown is happening normally even when the log is not owned.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -184,20 +189,31 @@ Dimr::~Dimr(void)
...
 
 #ifndef _WIN32
     free(exeName);
 #else
     delete[] exeName;
 #endif
 
+    if (logIsOwned)
+    {
+        delete log;
```
KoenBussemakerDeltares: Consider a `std::unique_ptr` to store the logger instance. You can use it to check for ownership by testing against null, and you don't have to worry about cleanup.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1659,7 +1679,7 @@ void Dimr::scanUnits(XmlTree* rootXml)
     // Scan
     for (int i = 0; i < rootXml->children.size(); i++)
     {
-        if (strcmp(rootXml->children[i]->name, "component") == 0)
+        if (rootXml->children[i]->name == "component")
```
KoenBussemakerDeltares, line 1684: This is incorrect, == will do a pointer comparison in this case (`name` is a char*, and the string literal will resolve to one as well). Use `strcmp` or put `std::string("component")` on the right hand side to trigger the use of `std::string`'s equality operator.

If you want to be bold you can try changing the name from a char to a std::string, but you might just get into more trouble.

There are many other places where this happens too, make sure you check thoroughly.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1841,37 +1861,39 @@ void Dimr::scanComponent(XmlTree* xmlComponent, dimr_component* newComp)
...
-        log->Write(INFO, my_rank, "         workingDir is set to %s", newComp->workingDir);
+        log->Write(INFO, my_rank, "         workingDir is set to %s", rawWorkingDir);
     }
     else
     {
-        newComp->workingDir = workingDirElement->charData;
+        rawWorkingDir = workingDirElement->charData.c_str();
     }
     // Is workingDir a valid relative path?
     char* combinedPath = new char[MAXSTRING];
```
KoenBussemakerDeltares, line 1883: Consider using std::filesystem to fix this messy path handling. I also understand if you want to limit the scope of this PR, it is up to you.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -472,26 +444,13 @@ void XmlTree::ExpandEnvironmentVariables() { return this->ExpandEnvironmentVaria
...
-        orgstr = this->charData;
-        instr = orgstr;
-        outstr = SubstEnvVar(instr) + '\0';
-        delete[] this->charData;
-        this->charData = new char[outstr.length()];
-        outstr.copy(this->charData, outstr.length());
+        this->charData = SubstEnvVar(this->charData);
     }
 
     for (int i = 0; i < children.size(); i++)
```
KoenBussemakerDeltares, line 456: This file is an excellent reminder why we should not write our own xml parsers

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -173,54 +168,30 @@ XmlTree::XmlTree(XmlTree* parent, const char* name)
...
-    else
-        ppn = parent->pathname;
-
-    int pathlen = strlen(ppn) + strlen(name) + 2;
-    if (pathlen > this->maxPathname)
-        throw Exception(Exception::ERR_XML_PARSING, " XML pathname for node \"%s\" is too long", name);
-
-    this->name = new char[strlen(name) + 1];
-    strcpy(this->name, name);
+    const std::string& ppn = (parent == NULL) ? std::string() : parent->pathname;
```
KoenBussemakerDeltares: If somebody ever has to remove the `const` it will lead to a dangling reference. I know this is guaranteed to work in it's current form, but IMO it's better to just make it a copy for clarity sake.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -84,22 +84,26 @@
     #define strdup _strdup
 #endif
 
+struct ParseState
+{
+    XmlTree** curnode;
+    std::string charDataStr;
```
KoenBussemakerDeltares: Don't use the type in the name

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -117,35 +121,33 @@ XmlTree::XmlTree(FILE* input)
...
-        while (IS_WHITESPACE(*end)) end--;
-        *++end = '\0';
-
-        int len = strlen(begin);
-        (*curnode)->charData = new char[len + 1];
-        memcpy((*curnode)->charData, begin, len);
-        (*curnode)->charData[len] = '\0';
-        (*curnode)->charDataLen = len + 1;
-        CharDataLen = 0;
+        // Trim leading whitespace
```
KoenBussemakerDeltares: Isn't there a nice boost function that can do this?

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -282,30 +253,31 @@ int XmlTree::Lookup(const char* pathname, int instance,
...
     //  (think of a backwards dirname/basename)
 
-    char* path = new char[strlen(pathname) + 1];
-    strcpy(path, pathname);
-    char* remainder = strchr(path, '/');
-    if (remainder == NULL)
-        remainder = (char*)"";
-    else
-        *remainder++ = '\0';
+    std::string path = pathname;
```
KoenBussemakerDeltares, line 266: Again, consider using `std::filesystem` for this

KoenBussemakerDeltares (commented):
I left a few remarks. Nice improvements in general!

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: I removed logIsOwned.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Although I don't think it's wrong (In "StringA == charB", charB will be automatically converted to a string), I changed this into "StringA.compare(charB)" for more clarity.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Sorry, I go for limiting the scope.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
adrimourits: Waarvan akte.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
adrimourits: Changed into `std::string`

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
adrimourits: Changed into `charData`

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
adrimourits: Yes, boost::trim looks exactly as what we need here.
Boost was not used as a dependency yet by DIMR.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
adrimourits: Sorry, I go for limiting the scope.

adrimourits:
Hi @harmenwierenga , I spent about a day now trying to get this branch working, but it just doesn't work on Linux. I give up. I'll assign this issue back to you.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: The Dimr instance will own the log only if DimrExe did not call set_dimr_logger (or BmiDimrSetLogger). A unique_ptr implies unique ownership. I can look into making it a std::variant<std::unique_ptr<Log>,Log*>.

harmenwierenga: I opted for a separate ownedLog_ unique_ptr that manages ownership and a separate Log* log that points to the correct one.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: Kept Adri's changes, all logIsOwned checks were overly cautious

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: XmlTree::name is std::string, so it was correct indeed. I personally find == clearer than .compare == 0, so I will stick to the original == code. Operator== is also overloading for (std::string, const char*), so there is no efficiency difference either

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: I changed it to use std::filesystem.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
harmenwierenga: I think that the new boost addition was a bit complicated, because dimr was not using boost and our current third party libs system requires some additional hard coded search paths for finding the boost package. Since the only usage here is a trim function, I decided that it would be cleaner to refactor the three uses here into a local trim function

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
harmenwierenga: These are not file paths, they are xml paths.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: I added the checks back: Dimr is destructed at static time, there is no guarantee that DimrExe's log is still alive, so we can only log if Dimr owns the log.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: I changed it back. Tests keep breaking, first also locally in debug, but after some fixes only on teamcity's release build in combination with mpi. I hope that the partial revert restores the tests

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
KoenBussemakerDeltares: Sounds like a (potential) static order destruction fiasco, which are notoriously annoying to fix. It's ok to leave it this way for now.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
KoenBussemakerDeltares: Ah yes, it was changed to a string in this PR. Then ignore my earlier comments, == is fine. I must have looked against the main-version XmlTree.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
KoenBussemakerDeltares: It was worth a try :)

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
KoenBussemakerDeltares: Hopefully that will be a thing of the past when we start using conan, but for now this is a fair compromise.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
KoenBussemakerDeltares: Check, ignore my comments then

harmenwierenga:
Solving the memory leaks appears to cause dimr crashes (probably static destruction order fiasco). A reduced PR is green, will continue there: https://github.com/Deltares/Delft3D/pull/845

PR #837, opened 2026-04-29
All/feature/unst 9650 couple correct water depth between fm and c sumo

erwin-mulder:
# What was done 

 Couple correct water depth between fm and c sumo, which is read from the 2d FM mesh

# Evidence of the work done 

- [ ]	Video/figures \
- [x]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link


adrimourits (approved):
Looks good to me.
At some point either Harmen or Koen should do some reviewing too, but merging in main is fine.

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
```diff
@@ -4,12 +4,68 @@
...
+     * in preCICE communication buffers in O(1) time.
+     */
+    Mesh getMesh2d(const std::string_view csumo_mesh_name, const CSumoSettingsReader& csumo_settings)
+    {
+        const int dimensions = 2;
+        Mesh mesh = {};
+        mesh.name = csumo_mesh_name;
+        for (const DiffuserSettings& d : csumo_settings.diffusers())
+        {
+            const std::size_t diffuser_index = mesh.coordinates.size() / dimensions;
```
AabhasSrivastava: I was a bit confused with the naming. Based on code it looks like diffuser start index in the whole flat array- right?

AabhasSrivastava: I guess that is also because diffuser is an entity, but also a point within that entity - maybe something to look into.

AabhasSrivastava (approved):

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
erwin-mulder: I renamed the variable to `diffuser_index_mapping` so make it somewhat more obvious it is part of the mapping and not the index to the diffuser itself.

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
```diff
@@ -17,6 +20,29 @@ namespace pre_c_sumo
      * preC-SUMO library. They handle timestepping control, configuration and settings
      * files parsing and the conversion/communication of NF/FF data.
      */
+    // TODO?: Move/fold into class(es)?
+    const std::string water_levels_id = "sea_surface_height";
+    const std::string bed_levels_id = "sea_floor_depth_below_geoid";
+    const std::string water_depth_id = "sea_floor_depth_below_sea_surface";
+
+    typedef struct
```
harmenwierenga: typedef struct and the name at the end is not necessary, change to
`struct DiffuserMapping {};`

Same for Mesh

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
```diff
@@ -42,21 +106,21 @@ namespace pre_c_sumo
 
         participant.initialize();
         double coupling_time_step;
```
harmenwierenga: Is coupling_time_step used outside of the while loop? Otherwise its declaration can be moved closer to its use

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
```diff
@@ -17,6 +20,29 @@ namespace pre_c_sumo
      * preC-SUMO library. They handle timestepping control, configuration and settings
      * files parsing and the conversion/communication of NF/FF data.
      */
+    // TODO?: Move/fold into class(es)?
+    const std::string water_levels_id = "sea_surface_height";
```
harmenwierenga: I don't like global std::strings, because they have constructors/destructors, and the order of calling these at startup and shutdown is not defined. That means that no other globals are allowed to use them (otherwise there is undefined behaviour). Instead, if we use constexpr std::string_view, they simply point to the value that is baked into the binary, there is no heap memory, and no chance of undefined behaviour

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
```diff
@@ -37,7 +63,7 @@ namespace pre_c_sumo
      * Blocking receive of farfield data via preCICE.
      * The demo implementation only logs a message.
      */
-    void receiveFFData();
+    void receiveFFData(precice::Participant& participant, Mesh& csumo_2d_mesh, const double coupling_time_step);
```
harmenwierenga: const at the value level does nothing in a function declaration, it is only meaningful in the function definition:

```
// This declaration
int f_1(const int x);
// is the same as this declaration
int f_2(int x);
// These declarations are separate though, because a ref or pointer to a const int is part of the signature
int g_1(const int& x);
int g_2(int& x);
```

So you can remove the const from const double coupling_time_step in the declaration, because it is meaningless to a user of the function. It is meaningful on the definition side, where it indicates that the function does not modify it.

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
```diff
@@ -17,6 +20,29 @@ namespace pre_c_sumo
...
+        size_t number_of_ambient_points;
+        size_t first_ambient_point_index;
+    } DiffuserMapping;
+
+    typedef struct
+    {
+        std::string name;
+        std::vector<double> coordinates;
+        std::vector<int> vertex_ids;
+        std::vector<DiffuserMapping> forward_map;
```
harmenwierenga, line 42: What does forward_map mean?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -30,24 +30,47 @@ namespace pre_c_sumo
...
     {
         // TODO: obtain these from the far-field model / coupling state
-        const double current_time_seconds = 0.0;
         const std::string run_id = "FlowFM";
-        const std::vector<std::string> constituent_names = {"temperature"}; // TODO: derive from settings
+        const std::vector<std::string> constituent_names = {"temperature", "salinity",
+                                                            "tracer"}; // TODO: derive from settings
 
-        for (const auto& [index, diffuser] : csumo_settings.diffusers() | std::views::enumerate)
+        for (const auto&& [index, diffuser] : csumo_settings.diffusers() | std::views::enumerate)
```
harmenwierenga: What made you make this change? const auto&& is probably never useful

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -56,8 +79,14 @@ namespace pre_c_sumo
                 };
             };
 
-            const auto ambient_points =
-                diffuser.ambient_positions | std::views::transform(make_point) | std::ranges::to<std::vector>();
+            std::vector<FarFieldPoint2D> ambient_points{};
+            for (const auto&& [position_index, ambient_point] : diffuser.ambient_positions | std::views::enumerate)
```
harmenwierenga: I personally prefer ranges and algorithms over naked loops. Why did you make this change?

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
```diff
@@ -4,12 +4,68 @@
...
+    }
+
+    /**
+     * @details Construct both preCICE 2d mesh coordinates and forward mappings from settings.
+     * The latter allows us to find the index of values belonging to diffusers, intakes and ambient points
+     * in preCICE communication buffers in O(1) time.
+     */
+    Mesh getMesh2d(const std::string_view csumo_mesh_name, const CSumoSettingsReader& csumo_settings)
+    {
+        const int dimensions = 2;
```
harmenwierenga: constexpr

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
```diff
@@ -4,12 +4,68 @@
...
+        Mesh mesh = {};
+        mesh.name = csumo_mesh_name;
+        for (const DiffuserSettings& d : csumo_settings.diffusers())
```
harmenwierenga: Could you use a better name than 'd'?

erwin-mulder: renamed it to diffuser, also changed p to position in a similar loop.

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
erwin-mulder: fixed

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
erwin-mulder: It is a map that we use to find values for the diffuser-, intake-, ambient points to the (1-dimensional) preCICE value/data buffers.

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
erwin-mulder: done. It was mainly added because of keeping the hpp and cpp prototypes the same.

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
erwin-mulder: Changed it.

src/tools_gpl/pre_c_sumo/include/coupling_steps.hpp
erwin-mulder: Done.

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
erwin-mulder: I added this by request of Adri, I am assuming he is going to expand on this in his work.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
erwin-mulder: Changed it back, there was a problem with compiling this on Linux at some point.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
erwin-mulder: Feel free to refactor this at a later stage. At this point nobody in the team understands the ranges approach.

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
erwin-mulder: Moved the declaration inside the loop.

harmenwierenga (approved):

PR #845, opened 2026-05-04
UNST-9861 Create minimal memory leak fix for xmltree. Fix small bugs …

harmenwierenga:
…in error logging.

# What was done 

This is a safe subset of pull request https://github.com/Deltares/Delft3D/pull/815. Solving all memory leaks using a singleton pattern appears to cause a static destruction order fiasco, so I only focused on keeping some bug fixes and keeping the memory leak fixes in xmltree

- Fix small issue where inquire accesses trailing whitespace in Fortran code
- Inside xmltree, use std::string over allocated char*. Make the fields private that can be private.
- Fix bug where workingDir was assigned to a pointer to pointer cast to a char * (garbage): newComp->workingDir = (char*)&curPath;
- Store the charDataStr (xmltree) in a state variable of the library, instead of in global data, so that there is no global std::string with constructors/destructors that can cause issues
- Properly delete all children of the XML nodes

 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
+++ b/src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
@@ -1386,8 +1386,8 @@ void Dimr::receive_ptr(const char* name, const char* sourceName, int compType, B
     strcat(nameShape, "_shape");
     if (compType != COMP_TYPE_DSLE)
     {
-       //Setting nameshape in DSLE registers it as a constituent, which we do not want.
-       (dllSetVar)(nameShape, shape);
+        // Setting nameshape in DSLE registers it as a constituent, which we do not want.
+        (dllSetVar)(nameShape, shape);
     }
     // Finally: call setvar(name, pointer)
     (dllSetVar)(name, (void*)sourceVarPtr);
@@ -1594,7 +1594,7 @@ void Dimr::scanConfigFile(void)
                         configfile);
 
     // Check version number
-    const char* versionnr = fileversion->charData;
+    const char* versionnr = fileversion->charData.c_str();
     float versionnumber;
     int intRead = sscanf(versionnr, "%f", &versionnumber);
     if (intRead != 1)
@@ -1634,7 +1634,7 @@ void Dimr::scanGlobalSettings(XmlTree* rootXml)
     XmlTree* loggerNcFormat = globalSettings->Lookup("logger_ncFormat");
     if (loggerNcFormat != NULL)
     {
-        int intRead = sscanf(loggerNcFormat->charData, "%d", &(nc_mode));
+        int intRead = sscanf(loggerNcFormat->charData.c_str(), "%d", &(nc_mode));
         if (intRead != 1)
             throw Exception(Exception::ERR_INVALID_INPUT, "logger_ncFormat must contain the value 3 or 4");
         if (nc_mode == 3)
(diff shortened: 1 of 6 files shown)


src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -84,22 +84,73 @@
...
+    };
+
+    void starttag(void* userdata, const XML_Char* name, const XML_Char* attr[])
+    {
+        XmlTree** curnode = static_cast<ParseState*>(userdata)->curnode;
+        XmlTree* node = new XmlTree(*curnode, name);
+        (*curnode)->AddChild(node);
+        *curnode = node;
+
+        for (int i = 0; attr[i] != NULL && attr[i + 1] != NULL; i += 2) node->AddAttrib(attr[i], attr[i + 1]);
```
KoenBussemakerDeltares: I'm generally not a fan of inline fors, they tend to create quite clutter code. Write it out using curly braces.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -2432,8 +2436,9 @@ void Dimr::freeLibs(void)
         DWORD ierr;
```
KoenBussemakerDeltares: ierr can be created inside the if statement, on line 2441

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
```diff
@@ -84,22 +84,73 @@
...
+    {
+        ParseState* state = static_cast<ParseState*>(userdata);
+        XmlTree** curnode = state->curnode;
+
+        if (!state->charData.empty())
+        {
+            std::string trimmed = trim(state->charData);
+            if (!trimmed.empty())
+            {
+                (*curnode)->charData = std::move(trimmed);
```
KoenBussemakerDeltares: Doing this conditionally looks like a premature optimization to me. I'd just always do the move, even if the string is empty. It's likely that std::string's move operator already does a similar check, and if not then it's something you're probably never going to notice in terms of performance.

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
harmenwierenga: Done

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
harmenwierenga: Done

src/engines_gpl/dimr/packages/dimr_lib/src/xmltree.cpp
harmenwierenga: I thought that it would be a logic change, but this is the first place where (*curnode)->charData is filled, so that one is empty. I have removed the if statement.

harmenwierenga:
These changes appear to cause randomly hanging tests that cause on windows:
The build Test (auto-generated)::[fm dsle] #all-testbench: e727e9201eed0ea4cff6f4ebbf59505ba9ceff93 {buildId=7186784} has been running for more than 90 minutes. Terminating...

PR #846, opened 2026-05-04
added parsing of csv file (output from teamcity case list) as input f…

FlorisBuwaldaDeltares:
…or run_testbench.py

# What was done 

- added .csv parser to the --filter argument of testbench.py, so that you can easily run a list of failed testcases downloaded from teamcity
 

# Evidence of the work done 

image

image

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/test/deltares_testbench/src/utils/test_bench_parameter_parser.py
+++ b/test/deltares_testbench/src/utils/test_bench_parameter_parser.py
@@ -3,6 +3,7 @@
 Copyright (C)  Stichting Deltares, 2026
 """
 
+import csv
 import getpass
 import os
 from argparse import ArgumentParser, Namespace
@@ -63,7 +64,7 @@ class TestBenchParameterParser:
         # Additionally, extra TeamCity messages will be produced.
         settings.teamcity = cls.__get_argument_value("teamcity", args) or False
 
-        settings.filter = args.filter
+        settings.filter = cls.__resolve_filter(args.filter)
         # Determine type of run
         settings.run_mode = cls.__get_argument_value("run_mode", args) or ModeType.LIST
         settings.config_file = cls.__get_argument_value("config", args) or "config.xml"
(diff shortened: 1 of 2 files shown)


test/deltares_testbench/src/utils/test_bench_parameter_parser.py
```diff
@@ -63,7 +64,7 @@ def parse_arguments_to_settings(cls) -> CommandLineSettings:
         # Additionally, extra TeamCity messages will be produced.
         settings.teamcity = cls.__get_argument_value("teamcity", args) or False
 
-        settings.filter = args.filter
+        settings.filter = cls.__resolve_filter(args.filter)
```
rene-deltares: I would not jam this together with the '--filter' argument. Add a separate argument for this, I don't know '--teamcity-csv' or something like that

rene-deltares: Also make sure the testbench parser then immediately converts it to a `patlib.Path` object instead of a string. You can immediately validate iwth `.is_file` that it exists, and just abort when it does not.

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
```diff
@@ -73,6 +74,70 @@ def parse_arguments_to_settings(cls) -> CommandLineSettings:
...
+            return filter_arg
+
+        failed_names = cls.__read_failed_tests_from_csv(filter_arg)
+        if not failed_names:
+            return filter_arg
+
+        return f"testcase={','.join(failed_names)}"
+
+    @classmethod
+    def __read_failed_tests_from_csv(cls, csv_path: str) -> List[str]:
```
rene-deltares: Please make this a public method or something that you can write a unit test for. Also, don't pass the path, but the result of `open(path)` (a `TextIO` object). That way you don't have to actually create a file in your test, but you can work with text instead.

Then opening the file can be done before calling this method, and you end up with a method that's much easier to write a unit test for.

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
FlorisBuwaldaDeltares: sure, will do!

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
```diff
@@ -73,6 +74,70 @@ def parse_arguments_to_settings(cls) -> CommandLineSettings:
...
+        with open(csv_path, newline="", encoding="utf-8") as csv_file:
+            reader = csv.DictReader(csv_file)
+            required_columns = {"Test Name", "Status"}
+            if not required_columns.issubset(reader.fieldnames or []):
+                missing = required_columns - set(reader.fieldnames or [])
+                raise ValueError(
+                    f"CSV file '{csv_path}' is missing required columns: {missing}. "
+                    f"Found: {reader.fieldnames}"
+                )
+            for row in reader:
```
rene-deltares: This only works if all of the failed tests are in a single config, the same config that is passed with the `--config` argument. If there's any failing tests in this CSV that's not in that XML I don't know what the behavior is. A crash? A warning?

rene-deltares (changes requested):

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
FlorisBuwaldaDeltares: i'm not sure, but i feel that's out of scope for this test? this whole setup relies on the current implementation of the --filter argument. It is the users responsibility to supply case names in the filter argument that are present in the testbench xml, the only functionality added is parsing the string from a csv file instead of having to manually supply it

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
FlorisBuwaldaDeltares: will do

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
FlorisBuwaldaDeltares: nonexisting cases are ignored, i just tested it

test/deltares_testbench/src/utils/test_bench_parameter_parser.py
```diff
@@ -172,12 +214,22 @@ def __create_argument_parser(cls) -> ArgumentParser:
...
         )
-        parser.add_argument(
+        filter_group = parser.add_mutually_exclusive_group()
+        filter_group.add_argument(
             "--filter",
             default="",
             help="Specify what tests to run based on filter (--list for options)",
             dest="filter",
         )
+        filter_group.add_argument(
```
rene-deltares, line 224: I would be more explicit and mention teamcity in the argument. I think `--teamcity-csv` is a better name and also in the help text, so that it's clear what kind of CSV is meant here: The one from the TeamCity builds. This also avoids confusion with the minio tools csv.

FlorisBuwaldaDeltares: done

rene-deltares (approved):
Danku!

PR #861, opened 2026-05-06
UNST-9874: increase numerical range of time variables in D-WAVES

jreyns:
- Change time variable precision to hp
- Cleanup

TO DO: check if the integer TIME dimension of the wavm file in seconds does not cause more trouble, as that is essentially left alone by casting upscaled tscale back to sp

# Evidence of the work done 

- [ ]	Video/figures \
- [x ]	Clear from the issue description 
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
Test to add @marcio-albernaz 
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ x]	Not applicable 

# Issue link
closes UNST-9874

### DIFF
--- a/src/engines_gpl/wave/packages/io/src/get_params.f90
+++ b/src/engines_gpl/wave/packages/io/src/get_params.f90
@@ -32,6 +32,8 @@ subroutine get_params(tscale, rho, filnam)
 !!--pseudo code and references--------------------------------------------------
 ! NONE
 !!--declarations----------------------------------------------------------------
+    use precision
+   
     implicit none
 !
 ! Local parameters
@@ -40,7 +42,7 @@ subroutine get_params(tscale, rho, filnam)
 !
 ! Global variables
 !
-    real, intent(out)               :: tscale
+    real(hp), intent(out)           :: tscale
     real, intent(out)               :: rho
     character(256)                  :: filnam  ! filename com-file
 !
@@ -72,7 +74,7 @@ subroutine get_params(tscale, rho, filnam)
 !! executable statements -------------------------------------------------------
 !
     rho    = -999.0
-    tscale = -999.0
+    tscale = -999.0_hp
     allocate(rbuff(1))
     wrswch = .false.
     ielem  = 1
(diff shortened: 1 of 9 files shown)


src/engines_gpl/wave/packages/io/src/get_params.f90
```diff
@@ -106,7 +108,7 @@ subroutine get_params(tscale, rho, filnam)
               & elmqty    ,elmunt    ,elmdes    ,elmtps    ,nbytsg    , &
               & elmnms(ielem)        ,celidt    ,wrswch    ,error     ,rbuff     )
     if (error /= 0) goto 200
-    tscale = rbuff(1)
+    tscale = real(rbuff(1),hp)
```
adrimourits, line 111: The data in a Delft3D4 com-file (in nefis format) seems to be single precision. This might mean that the problem is not fixed for Delft3D4. Do you agree?

adrimourits (approved):

src/engines_gpl/wave/packages/io/src/get_params.f90
jreyns: yes, this is what my TO DO comment above refers to. I am not a NEFIS specialist, so a bit hesitant to touch that. I asked @marcio-albernaz to make an equivalent d3d4 model to test this hypothesis. Let's see

hrajagers: Delft3D 4 writes time as integer multiple of `tscale`. D-Flow FM writes time as floating point multiple of seconds. It's not clear to me that they would run into the same issue at the same time.

jreyns: I think we are good on Delft3d4 side. Like Bert says, time is written in seconds, and that gives us a range of 1e14 days since refdate, and no potential for rounding errors with the real(tscale,hp) statement above. All the other upgrades of precision handle propagation of single precision floats

marcio-albernaz: I see that it passed all tests from TeamCity. I will get the binaries from TC and test it with my model, likely this friday.

marcio-albernaz: For FM it seems to fix all issues the I saw before. 
The wave .nc and .dat outputs are correct, the com-file is correct and the swan INPUT file is correct. 

Is it indeed necessary to check a Delft3D-4 analogue run? Asking it since the DAT file from waves is also NEFIS. Making the D3D-4 will take some extra time.

PR #863, opened 2026-05-06
UNST-9921: add tracer diffusion test with explicit timestep + mass balance

willem4:
Add tracer diffusion test with explicit timestep + mass balance

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
e02_f019_c091
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9921


test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
```diff
@@ -4819,6 +4819,24 @@
         </file>
       </checks>
     </testCase>
+    <testCase name="e02_f019_c091_tracer_diffusion_explicit_mass_balance" ref="dflowfm_default"> 
+      <path version="2026-05-06T13:02:27.883000">e02_dflowfm/f022_morphology_2D/c108_settling_mass_balance</path>
```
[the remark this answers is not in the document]
willem4: fix in the later commit

test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
```diff
@@ -4819,6 +4819,24 @@
         </file>
       </checks>
     </testCase>
+    <testCase name="e02_f019_c091_tracer_diffusion_explicit_mass_balance" ref="dflowfm_default"> 
+      <path version="2026-05-06T13:45:30.244000">e02_dflowfm/f019_transport/c091_tracer_diffusion_explicit_mass_balance</path>
+      <maxRunTime>60.0000000</maxRunTime>
+      <checks>
+        <file name="dflowfmoutput/masscon_map.nc" type="netCDF">
+          <parameters>
+               <!-- current error with exact solution is approximately 1.2e-4 -->
```
hrajagers, line 4828: The doc states: "The absolute error in the tracer concentration remains below $1.2 \cdot 10^4$." Looks like a missing minus ...

test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
```diff
@@ -4819,6 +4819,24 @@
...
+      <checks>
+        <file name="dflowfmoutput/masscon_map.nc" type="netCDF">
+          <parameters>
+               <!-- current error with exact solution is approximately 1.2e-4 --> 
+      			<parameter name="mesh2d_tracer" toleranceAbsolute="1.5e-4" />
+          </parameters>
+        </file>
+        <file name="dflowfmoutput/masscon_mass_balances.nc" type="netCDF">
+          <parameters>
+               <!-- current error is approximately -1e-12 -->
```
hrajagers, line 4834: The error is seems to increase linearly over time ... will the error eventually become larger than the initial mass, or will it be limited to a reasonable value as the effect of the diffusion reduces?

hrajagers (commented):
Please correct the minus-sign in the documentation. Can you run a longer model to test whether the mass error stabilizes at some point?

test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
willem4: The same case, but then run for 100000 times longer leads to the following result. So the error in the solution goes to zero, but the cumulative mass error seem to grow due to the rounding error. I cannot explain the jumps in the instantaneous mass error plot (order 10e-12).


comparison_abs
comparison_diff
comparison_diff_time
mass_error
mass_error_instant

test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
willem4: This is adjusted in the documentation

test/deltares_testbench/configs/include/dimr_dflowfm_all_but_validation_cases.xml
hrajagers: Thank you! Informative ... and intriguing. The jumps in the "tracer balance error" match with the changes in slope for the "tracer balance cumulative error" as expected. Indeed, not sure where these jumps come from ... looks like some precision issue. If the total water volume is constant, I would expect that the "tracer balance cumulative error" graph would have the same shape as the "average concentration difference" if the difference is computed as the average concentration in the model compared to the analytically obtained average concentration (which should be constant and equal to the initial concentration = (initial) mass / (initial) water volume).

willem4: I think the mass balance error shows how much mass is created or destroyed at every time step. As these rounding errors remain they add up or subtract down, but I do not think there is a need for it to go to zero or level off (would be nice though ..).

hrajagers: Indeed, but I think that there is neither a reason why the "average concentration difference" should go to zero ... unless all concentration would diffuse through open boundaries to an infinite reservoir of zero concentration, but this model doesn't have open boundaries. So, the total tracer mass shouldn't change, the total water volume shouldn't change, hence the average concentration shouldn't change, and thus any error in these particular quantities since the beginning should equal the error compared to the analytical constant solution for these quantities. I see that we compute the cumulative error by adding up the errors per time step instead of computing the mass per time step and comparing it to the initial mass (corrected for cumulative fluxes which are zero by definition in this case without open boundaries) ...

willem4: I think the wording in the ylabel is wrong: It should be the `average difference in concentration (modelled - exact)`. The final solution is a constant concentration so that is why the error should decrease to zero. Would you like me to adjust the computation of cumulative mass error?

hrajagers: Ok, but I still don't understand why the concentration would be "correct" when the mass has "a slightly larger error" (not to suggest that it's completely wrong). Let's move ahead and pin this result, but I'm interested in revisiting the computation at some point. Maybe when formalizing the D-Morphology validation document.

willem4: Ok let's do that - I have created UNST-10010 for the remaining unkown. Can you resolve and approve this conversation and approve the review if you agree? Thanks

hrajagers:
@willem4 Can you merge main and resolve the merge conflict? The failing case should then also be fixed.

willem4:
> @willem4 Can you merge main and resolve the merge conflict? The failing case should then also be fixed.

Done

hrajagers (approved):
Thank you for adding the test case.

PR #876, opened 2026-05-11
unst-9643: bubble screens output

mihai-pricope:
# What was done 

Output vars for Bubble Screens
*  the bubble screens sinks or sources, subdivided over possible all vertical layers in the computational cells with bubble screens, should not be written anymore to the HIS file.
* Output Air Discharge
* Output BubbleScreen geometry as separate entry in the HIS file (same as Source Sink geometry). Remove Bubble screen from SourceSink geometry.  Check with other series that might be related with the geometry (x-coordinate of source sink, grid (source)sink) , etc)

# Evidence of the work done 

- [x ]	Video/figures 
image

image

image

- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x ]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9643


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -186,6 +189,53 @@ subroutine calculate_sediment_SBC(source_input)
...
+      call allocate_and_associate(source_input, num_normal_source_sink, source_sink_discharge2_out)
+      source_sink_discharge2_out = pack(source_sink_all_discharges(2, :), is_source_sink_normal)
+   end subroutine filter_source_sink_discharge2
+
+   subroutine filter_source_sink_discharge3(source_input)
+      use fm_external_forcings_data, only: num_normal_source_sink, is_source_sink_normal, source_sink_all_discharges
+      real(dp), pointer, dimension(:), intent(inout) :: source_input !< Pointer to source input array for the "SBCX" item, to be assigned once on first call.
+      call allocate_and_associate(source_input, num_normal_source_sink, source_sink_discharge3_out)
+      source_sink_discharge3_out = pack(source_sink_all_discharges(3, :), is_source_sink_normal)
+   end subroutine filter_source_sink_discharge3
```
FlorisBuwaldaDeltares, line 211: I think we can have better names than source_sink_discharge 1-3, don't these 3 columns represent something in the source_sink_all_discharges array?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -186,6 +189,53 @@ subroutine calculate_sediment_SBC(source_input)
       call assign_sediment_transport(SBCX, SBCY, IPNT_SBCX1, IPNT_SBCY1)
    end subroutine calculate_sediment_SBC
 
+   subroutine filter_source_sink_discharge(source_input)
+      use fm_external_forcings_data, only: num_normal_source_sink, is_source_sink_normal, source_sink_all_discharges
+      real(dp), pointer, dimension(:), intent(inout) :: source_input !< Pointer to source input array for the "SBCX" item, to be assigned once on first call.
```
FlorisBuwaldaDeltares: I don't see where this pointer is set? is it SBCX? isn't this a dummy argument that's never used? since the source_input seems to be source_sink_all_discharges(2,:)?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -186,6 +189,53 @@ subroutine calculate_sediment_SBC(source_input)
       call assign_sediment_transport(SBCX, SBCY, IPNT_SBCX1, IPNT_SBCY1)
    end subroutine calculate_sediment_SBC
 
+   subroutine filter_source_sink_discharge(source_input)
```
FlorisBuwaldaDeltares, line 192: as we discussed, let's make a followup issue where we think about an elegant way to eliminate these packing routines. If a customer would ever use source-sinks with statistical output, they would be horribly slow.
If we could somehow sort the source sinks so that the normal ones are indices 1:num_normal_source_sinks and the rest is bubblescreen then we can just take nice pointers to contiguous slices

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -2308,20 +2367,35 @@ subroutine flow_init_statistical_output_his(output_config_set, output_set)
       !
       ! Source-sink variables
       !
-      if (his_write_settings%sourcesink > 0 .and. num_source_sink > 0) then
-         call add_stat_output_items(output_set, output_config_set%configs(IDX_HIS_SOURCE_SINK_PRESCRIBED_DISCHARGE), source_sink_all_discharges(1, :))
+      if (his_write_settings%sourcesink > 0 .and. num_normal_source_sink > 0) then
+         function_pointer => filter_source_sink_discharge
+         call add_stat_output_items(output_set, output_config_set%configs(IDX_HIS_SOURCE_SINK_PRESCRIBED_DISCHARGE), null(), function_pointer)
+         ! call add_stat_output_items(output_set, output_config_set%configs(IDX_HIS_SOURCE_SINK_PRESCRIBED_DISCHARGE), source_sink_all_discharges(1, :))
```
FlorisBuwaldaDeltares: please remove commented code

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
```diff
@@ -3563,8 +3563,30 @@ subroutine reduce_double_array_max(N, var)
...
+
+      integer :: ierror
+
+#ifdef HAVE_MPI
+      call MPI_allreduce(var, dum, N, mpi_logical, mpi_lor, DFM_COMM_DFMWORLD, ierror)
+      var = dum
+#endif
+      return
+    end subroutine reduce_logical_array_or
+
```
FlorisBuwaldaDeltares, line 3589: why not use reduce_int_max with the integers set to either 0 or 1? I find that less confusing than logicals

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
```diff
@@ -5112,6 +5134,77 @@ subroutine gatherv_int_data_mpi_dif(ndata_send, data_send, ndata_gat, data_gat,
 
    end subroutine gatherv_int_data_mpi_dif
 
+!> Gathers real(kind=dp) array data from all processes to root, automatically handling size gathering.
+!! This convenience wrapper handles variable-length arrays by first gathering sizes, then the data.
+!! The output array is only allocated and filled on the root process. Use size(data_all) to get total count.
+   subroutine gatherv_double_auto(data_local, data_all, root, ierror)
```
FlorisBuwaldaDeltares: i don't like this routine name, gaterv_double_auto? can we think of something more intuitive?
Also is root ever something other than 0? otherwise it does not need to be an input argument.

I would perhaps call it merge_array_mpi or something, to signify the array size is the union of all local arrays

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
```diff
@@ -5112,6 +5134,77 @@ subroutine gatherv_int_data_mpi_dif(ndata_send, data_send, ndata_gat, data_gat,
...
+#endif
+      implicit none
+
+      real(kind=dp), dimension(:), intent(in) :: data_local !< Local array to send
+      real(kind=dp), dimension(:), allocatable, intent(out) :: data_all !< Gathered array (allocated on root with total size, dummy size 1 on non-root)
+      integer, intent(in) :: root !< Rank of receiving process
+      integer, intent(out) :: ierror !< Error index
+
+      integer :: i, ndata_total
+      integer, dimension(:), allocatable :: recvcounts, displs
```
FlorisBuwaldaDeltares: I would suggest receive_counts, not recvcount. Also displacements, not displs

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
```diff
@@ -5112,6 +5134,77 @@ subroutine gatherv_int_data_mpi_dif(ndata_send, data_send, ndata_gat, data_gat,
...
+      integer :: ndata_local !< Number of data elements in local array
+
+      ndata_local = size(data_local)
+      ierror = -1
+      
+      if (jampi == 0) then
+         ! Serial case: just copy the data
+         allocate(data_all(ndata_local))
```
FlorisBuwaldaDeltares: allocate statement is unnecessary, an allocatable will be allocated upon assignment

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
```diff
@@ -5112,6 +5134,77 @@ subroutine gatherv_int_data_mpi_dif(ndata_send, data_send, ndata_gat, data_gat,
...
+                      root, DFM_COMM_DFMWORLD, ierror)
+
+
+      ! Step 2: Calculate displacements and total size (root only)
+      if (my_rank == root) then
+         displs(1) = 0
+         do i = 2, ndomains
+            displs(i) = displs(i-1) + recvcounts(i-1)
+         end do
+         ndata_total = sum(recvcounts)
```
FlorisBuwaldaDeltares: maybe a bit more explanations on what these displacements are and why they are necessary?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
FlorisBuwaldaDeltares: doesn't this routine kind of do the same thing as my "reduce_cells" implementation? can that not be reused?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_structures.f90
```diff
@@ -1057,6 +1069,152 @@ subroutine get_geom_coordinates_of_structure(istrtypein, i, nNodes, x, y, maskLo
...
+            call mess(LEVEL_ERROR, 'Error in gatherv_double_auto in fill_geometry_bubblescreens')
+         end if
+
+         if (my_rank == 0) then
+            geomXBubbleScreen(global_idx:global_idx + size(collectGeomXBubbleScreens) - 1) = collectGeomXBubbleScreens
+            geomYBubbleScreen(global_idx:global_idx + size(collectGeomYBubbleScreens) - 1) = collectGeomYBubbleScreens
+            global_idx = global_idx + size(collectGeomXBubbleScreens)
+         end if
+      end do
+   end subroutine fill_geometry_bubblescreens
```
FlorisBuwaldaDeltares: as we discussed, I think this routine can be eliminated by saving the already reduced bubblescreen source-sink x's and y's during add_bubblescreen_source_sinks

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
```diff
@@ -677,35 +688,33 @@ subroutine unc_write_his(tim) ! wrihis
...
+                  integer :: num_points
+                  num_points = source_sink_max_xy_points(num_source_sink)
+                  call realloc(tm_source_sink_x, [num_normal_source_sink, max_source_sink_polyline_points])
+                  call realloc(tm_source_sink_y, [num_normal_source_sink, max_source_sink_polyline_points])
+                  j = 1
+                  do i = 1, num_source_sink
+                     if (is_source_sink_normal(i)) then
+                        tm_source_sink_x(j, 1:num_points) = source_sink_x(i, 1:num_points)
+                        tm_source_sink_y(j, 1:num_points) = source_sink_y(i, 1:num_points)
+                        j = j + 1
```
FlorisBuwaldaDeltares, line 703: this seems complex, can it not also be a pack() operation with is_source_sink_normal as mask?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
```diff
@@ -2501,6 +2501,51 @@ subroutine setup(iresult)
...
+      call realloc(is_source_sink_normal, num_source_sink, fill=.false.)
+
+      do i = 1, size(bubblescreens)
+         associate (bubblescreen => bubblescreens(i))
+            do j = 1, bubblescreen%num_flowcells
+               sidx = bubblescreen%source_sink_indices(j)
+               if (jampi == 1 .and. allocated(idomain)) then
+                  flownode_nr = bubblescreen%flowcell_indices(j)
+                  if (idomain(flownode_nr) == my_rank) then ! Check if flow cell is owned by current partition
+                     is_source_sink_normal(sidx) = .true.
```
FlorisBuwaldaDeltares: I think we need a comment here explaining why we need to make it an inverse map, because for a bubblescreen source sink i'd expect is_source_sink_normal = .false.

FlorisBuwaldaDeltares (commented):
nice implementation! I do have a few comments, many of which we already discussed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
mihai-pricope: Nope, as far as I understood from your code :) (allocate_and_associate), this will allocate the buffer source_sink_discharge2_out and the point the source_input to that buffer. 

and source_input is not the source of the data for this method (like in EC module) ... but the source for the statistical_output which in this method is actually the target.

At least this is my understanding.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
mihai-pricope: as far as I understand this are the 1st/2nd constituents. But you never know if there is salinity or temperature .. because if there is only one it can be anything

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
mihai-pricope: I fin booleans somehow less confusing that integers for 0/1. So you need to explain why 0/1 are less confusing than booleans/logicals. 

Except maybe if we need to avoid a reduce logical. But I am confused

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
mihai-pricope: source_sink_x, source_sink_y are a 2D matrix. This is some sort of a matrix_pack .... based a diagonal num_points size. But I don't know how to write this more concise. Can pack do this on a 2d matrix?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
mihai-pricope: https://issuetracker.deltares.nl/browse/UNST-9951

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
FlorisBuwaldaDeltares: ok! I won't ask you to change that

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/partition.F90
FlorisBuwaldaDeltares: yeah you are right, i was just unfamiliar with the booleans, but I do agree it is better

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
FlorisBuwaldaDeltares: hmm maybe it can be simpler but on second thought it's not a trivial loop. let's leave it as it is

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
FlorisBuwaldaDeltares: ah yes, the function pointer business is a bit weird but it seems ok :)

FlorisBuwaldaDeltares (approved):
looks good to me :)

PR #878, opened 2026-05-11
UNST-8945 Refactor Rhol select-case statements according to Transptype

ZetaTam-dl:
# What was done 

- Refactored the select-case statements in exporting the sediment transport variables in the his/map writers routine.
- Added a new function to get the `rhol` locally.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
[UNST-8945](https://issuetracker.deltares.nl/browse/UNST-8945)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -10833,14 +10662,7 @@ subroutine unc_write_map_filepointer(imapfile, tim, jaseparate) ! wrimap
...
                do l = 1, stmpar%lsedsus
-                  select case (stmpar%morpar%moroutput%transptype)
-                  case (0)
-                     rhol = 1.0_dp
-                  case (1)
-                     rhol = stmpar%sedpar%cdryb(sedtot2sedsus(sedtot2sedsus(l)))
-                  case (2)
-                     rhol = stmpar%sedpar%rhosol(sedtot2sedsus(sedtot2sedsus(l)))
-                  end select
+                  rhol = get_rhol(stmpar, sedtot2sedsus(sedtot2sedsus(l)))
```
hrajagers: Same as above.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_sediment.f90
```diff
@@ -217,5 +217,22 @@ subroutine deallocgrains() ! for all fractions:
          deallocate (sigsed)
       end if
    end subroutine deallocgrains
+   
+   
+   pure function get_rhol(stmpar, idx) result(rhol)
```
hrajagers, line 224: Add `!> function description` and `!< variable description`.

One might argue that `rhol` isn't the most descriptive name. A name such as `get_sediment_density` might look more attractive, but such a routine name would easily cause errors when applied anywhere outside the scope of output transport quantities. What `rhol` represents is actually sometimes a dry bed density, sometimes a solid density, and sometimes just plain one. So, `rhol` is the conversion factor relating the internal mass-based transport quantities to the unit of the transport quantities on the output file which might be in mass as well, or solid or bed volume. A function name that explains all that is too long, so let's keep the current name for the time being and add a bit of function description for context.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -6997,14 +6997,7 @@ subroutine unc_write_map_filepointer_ugrid(mapids, tim, jabndnd) ! wrimap
...
                do l = 1, stmpar%lsedsus
-                  select case (stmpar%morpar%moroutput%transptype)
-                  case (0)
-                     rhol = 1.0_dp
-                  case (1)
-                     rhol = stmpar%sedpar%cdryb(sedtot2sedsus(l))
-                  case (2)
-                     rhol = stmpar%sedpar%rhosol(sedtot2sedsus(l))
-                  end select
+                  rhol = get_rhol(stmpar, l)
```
hrajagers: I see that you dropped `sedtot2sedsus` here. I think that is fine ... but check the next comment.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -7234,14 +7178,7 @@ subroutine unc_write_map_filepointer_ugrid(mapids, tim, jabndnd) ! wrimap
...
                do l = 1, stmpar%lsedsus
-                  select case (stmpar%morpar%moroutput%transptype)
-                  case (0)
-                     rhol = 1.0_dp
-                  case (1)
-                     rhol = stmpar%sedpar%cdryb(sedtot2sedsus(l))
-                  case (2)
-                     rhol = stmpar%sedpar%rhosol(sedtot2sedsus(l))
-                  end select
+                  rhol = get_rhol(stmpar, sedtot2sedsus(l))
```
hrajagers, line 7181: Here you kept `sedtot2sedsus` factor ... check next comment.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -7311,14 +7227,7 @@ subroutine unc_write_map_filepointer_ugrid(mapids, tim, jabndnd) ! wrimap
...
             do l = 1, stmpar%lsedsus
-               select case (stmpar%morpar%moroutput%transptype)
-               case (0)
-                  rhol = 1.0_dp
-               case (1)
-                  rhol = stmpar%sedpar%cdryb(sedtot2sedsus(l))
-               case (2)
-                  rhol = stmpar%sedpar%rhosol(sedtot2sedsus(l))
-               end select
+               rhol = get_rhol(stmpar, l)
```
hrajagers: Here you removed `sedtot2sedsus` conversion again. See next comment.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -10760,14 +10617,7 @@ subroutine unc_write_map_filepointer(imapfile, tim, jaseparate) ! wrimap
...
                do l = 1, stmpar%lsedsus
-                  select case (stmpar%morpar%moroutput%transptype)
-                  case (0)
-                     rhol = 1.0_dp
-                  case (1)
-                     rhol = stmpar%sedpar%cdryb(sedtot2sedsus(sedtot2sedsus(l)))
-                  case (2)
-                     rhol = stmpar%sedpar%rhosol(sedtot2sedsus(sedtot2sedsus(l)))
-                  end select
+                  rhol = get_rhol(stmpar, sedtot2sedsus(sedtot2sedsus(l)))
```
hrajagers: Looks like a correct translation of the old code: two nested `sedtot2sedsus` conversions, but I think that indexing `sedtot2sedsus` twice doesn't make sense.  I think that the only reason why this works is because `sedtot2sedsus(L) = L` by construction: the bed load fractions are always in the last part of the sediment array. I think that the `sedtot2sedsus` transformation isn't actually needed anywhere in the code. @jreyns  Do you know of any reason why we would need `sedtot2sedsus` conversion?

hrajagers (commented):
I think that we're moving in the right direction, but the change showed that we're using a conversion array `sedtot2sedsus` in some places. And sometimes even applied twice. I think that we should remove that, but I've asked Johan for a second opinion.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
jreyns: @hrajagers the double sedtot2sedsus is a mistake I think, tracing back code to 2017. Funny that this never went wrong. The sedtot2sedsus was introduced when we did not have guaranteed suspended fractions at the beginning of the fraction numbering, for loops over the suspended fractions that needed to be mapped to arrays dimensioned on the toal amount of fractions. I t would not be bad to keep it, as if we ever decide to change this order, everything still works as before.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_sediment.f90
ZetaTam-dl: Descriptions of the function and variables added.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
ZetaTam-dl: removed double `sedtot2sedsus` and aligned all mapping as a single `sedtot2sedsus(l)`

hrajagers:
The failure of the `Test (Linux)` seems unrelated to this pull request. The updated D-Flow FM code looks fine. However ... it would be nicer if we could move the `get_rhol` routine to a place inside `src/utils/morphology` such that we can also use it in the equivalent output routines of Delft3D-FLOW: `wrsedm` and `wrsedh`. Unfortunately, `stmtype` is defined in `rdstm.f90` which is not used in Delft3D-FLOW ... need to think about this more carefully to come up with a good solution moving forward.

hrajagers (approved):
Since this is an *fm* branch, we should limit the code changes to D-Flow FM (including D-Morphology) only. Hence, we shouldn't change the Delft3D-FLOW code. So, I approve this change, but do recommend/request a follow-up pull request using an *all* branch to adjust the Delft3D-FLOW code in the same way and to move the `get_rhol` routine to re-usable location inside `utils_gpl/morphology/packages/morphology_io`.

PR #879, opened 2026-05-12
DELWAQ-1257 Adding two new Bubblescreen FM-WAQ testcases

michelle-deltares:
# What was done 

I tested two cases of an implementation of bubble screens in FM that result in extra mixing. We wanted to make sure that the mixing is also resulting in extra mixing in WAQ, and that the water and mass balances are conserving.

# Evidence of the work done 

# Tests 
- [x] New tests added:
- e02_f029_c400_bubblescreen_box_model
- e02_f029_c401_bubblescreen_nieuwe_meer

# Documentation  
- [x]	Documentation updated 
Added test case documentation

# Issue link
https://issuetracker.deltares.nl/browse/DELWAQ-1257


arjenmarkusdeltaresnl (changes requested):
I am not happy with two configuration files that are essentially doing the same thing. I am not convinced that the bubble screen is positioned in the best possible way.
See the JIRA issue for details.

arjenmarkusdeltaresnl (approved):

PR #885, opened 2026-05-15
All/task/UNST-8667 FM blocks swan trial 2

adrimourits:
# What was done 

- Introduced: component-related-parameter `mpi_barrier_sleep`
- Default is zero, except for D-Waves on Linux: 500
- Barrier: if `mpi_barrier_sleep` is zero: MPI_Barrier
   else MPI_IBarrier with sleep(mpi_barrier_sleep ms)
- Note the usage of `mpi_barrier_sleep` of the previous update, since the barrier is at the start of the loop
- Note that `mpi_barrier_sleep` should be zero when a loop is skipped
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated: e109_f02_c011c011_weir_timeseries_mpi \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 
- [x] TODO

# Issue link
[UNST-8667](https://issuetracker.deltares.nl/browse/UNST-8667)


adrimourits:
Hi @rene-deltares , I think this is more or less the code for component-dependent-mpi_barrier_sleep.
What is your opinion about it?

adrimourits:
Hi @rene-deltares , Can you find some time to review my code changes? I expect you will need about 30 minutes. My advise is to start with reading the newly added section 8.5.2 (pdf attached to the issue) of the manual. I will ask Erik to review the manual update. I will ask Marcio to have a look at the manual and do a test.

adrimourits:
Hi @harmenwierenga , a long time ago in a galaxy far away, you reviewed my first trial, see https://github.com/Deltares/Delft3D/pull/287 . Would you be so kind to have a look at this second trial? I asked Rene for a review too, Erik to have a look at the manual update and will ask Marcio to check the manual and do a test.

test/deltares_testbench/configs/dimr/dimr_dflowfm_drtc_dwaves_lnx64_parallel.xml
```diff
@@ -0,0 +1,107 @@
...
+            The Windows workaround causes on Linux to need to define program "dimr_seq", being a full copy of program "dimr".
+        -->
+        <program name="dimr_seq" logOutputToFile="true" programStringRemoveQuotes="true">
+            <location ref="reference_engines" type="reference">
+                <from>lnx64</from> 
+            </location>
+            <location ref="local" type="check">
+                <from>teamcity_artifacts/lnx64</from> <!-- just the first part of the directory structure, wil be file by TeamCity -->
+            </location>
+            <path>bin/run_dimr.sh</path>
```
harmenwierenga, line 78: Why does windows require OMP_NUM_THREADS=1 and linux not?

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -2441,6 +2484,35 @@ void Dimr::freeLibs(void)
     }
 }
 
+//------------------------------------------------------------------------------
+void Dimr::barrier(MPI_Comm comm, bool use_mpi, int mpi_barrier_sleep)
```
harmenwierenga: All three can be const here

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -2441,6 +2484,35 @@ void Dimr::freeLibs(void)
...
+{
+    if (use_mpi)
```
harmenwierenga: I personally prefer an early return over an if statement that spans the whole function:
```
if (!use_mpi)
{
    return;
}
```

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -2441,6 +2484,35 @@ void Dimr::freeLibs(void)
...
+    {
+        if (mpi_barrier_sleep == 0)
+        {
+            int ierr = MPI_Barrier(comm);
```
harmenwierenga: The ierr variable is never used. This will cause warnings with proper compiler settings

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1847,6 +1860,35 @@ void Dimr::scanComponent(XmlTree* xmlComponent, dimr_component* newComp)
...
+#endif
+        }
+        else
+        {
+            newComp->mpi_barrier_sleep = 0;
+        }
+    }
+    else
+    {
+        newComp->mpi_barrier_sleep = atoi(mpiBarrierSleep->charData);
```
harmenwierenga: Can we use std::from_chars instead of atoi? [from_chars](https://en.cppreference.com/cpp/utility/from_chars)

src/engines_gpl/dimr/packages/dimr_lib/include/dimr.h
```diff
@@ -106,6 +106,7 @@ class Dimr
     void printComponentVersionStrings(Level);
 
     void freeLibs(void);
+    void barrier(MPI_Comm, bool, int);
```
harmenwierenga: Why don't we give names to the parameters here? I know that it is not technically necessary, but it is not very idiomatic

src/engines_gpl/dimr/packages/dimr_lib/include/dimr_control_block.h
```diff
@@ -13,7 +13,8 @@ struct dimr_control_block
     dimr_control_block* subBlocks;
     int masterSubBlockId; // Identifying the unique subBlock of this controlBlock acting as the master
     int type;
-    dimr_unit unit; // pointer to the actual units (in case of no further subblocks)
+    int mpi_barrier_sleep; // Time to sleep in milliseconds while a component is updating
```
harmenwierenga, line 16: I think that the name does not reflect what it is, namely a duration. Would `mpi_sleep_duration` be better? I also wonder about the inclusion in dimr_control_block. Do the other fields of dimr_control_block, such as type and tStart, represent actual fields in the dimr_config.xml file's control blocks? Is there a different type that is passed around that can carry state like this?
Perhaps I would prefer a reference back to its owning component, so that the sleep duration can be found there.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1009,10 +1008,10 @@ void Dimr::runParallelUpdate(dimr_control_block* cb, double tStep)
...
             // Sync all partitions to execute the same component
             // This ensures that all flow calculations are finished before a wave calculation is started
-            if (use_mpi)
-            {
-                int ierr = MPI_Barrier(MPI_COMM_WORLD);
-            }
+            // This barrier is at the start of the loop. That means the mpi_barrier_sleep of the previous component is
+            // used. The component->mpi_barrier_sleep is copied to cb->mpi_barrier_sleep after the component is
+            // executed.
+            barrier(MPI_COMM_WORLD, use_mpi, cb->mpi_barrier_sleep);
```
harmenwierenga, line 1016: Is MPI_COMM_WORLD properly defined when use_mpi is false?

I mentioned adding an early return in the barrier function, but perhaps I prefer the if statement here, the way it was before. [It makes me think of this post](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html)

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1009,10 +1008,10 @@ void Dimr::runParallelUpdate(dimr_control_block* cb, double tStep)
         {
             // Sync all partitions to execute the same component
             // This ensures that all flow calculations are finished before a wave calculation is started
-            if (use_mpi)
-            {
-                int ierr = MPI_Barrier(MPI_COMM_WORLD);
-            }
+            // This barrier is at the start of the loop. That means the mpi_barrier_sleep of the previous component is
+            // used. The component->mpi_barrier_sleep is copied to cb->mpi_barrier_sleep after the component is
+            // executed.
```
harmenwierenga, line 1015: Is this what we want? This comment looks like a review comment that indicates that something needs changing 😄

test/deltares_testbench/configs/dimr/dimr_dflowfm_drtc_dwaves_lnx64_parallel.xml
adrimourits: Good catch. Synchronized.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/include/dimr.h
adrimourits: Done.

src/engines_gpl/dimr/packages/dimr_lib/include/dimr_control_block.h
adrimourits: The most-correct-name would be `mpi_barrier_sleep_duration`. Knowing that it's an integer, I can not imagine that confusion will pop-up if we drop the `_duration` part.
In the control_block: replacing it with a reference to the related component doesn't work, because it has to be set to zero if this control_block is skipped (dimr.cpp line 1291).

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Yes, MPI_COMM_WORLD is properly defined when use_mpi is false.

I read the post, understand the arguments, but in this case I prefer the if statement inside method barrier: the calling location (runParallelUpdate) is very complicated. Moving everything (including needed sanity checks) related to the barrier inside method barrier makes runParallelUpdate a little cleaner.

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: What we want is preCICE :-) . Or at least move the barrier to the end of the loop. But I don't have the guts for that. For sure out of scope for this issue.

adrimourits:
Hi @harmenwierenga , Can you re-review please?

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1847,6 +1862,46 @@ void Dimr::scanComponent(XmlTree* xmlComponent, dimr_component* newComp)
...
+#endif
+        }
+        else
+        {
+            newComp->mpi_barrier_sleep = 0;
+        }
+    }
+    else
+    {
+        // Parse mpiBarrierSleep using std::from_chars (no locale, no heap)
```
harmenwierenga: I don't think that the comment really helps here. It could be refactored into a function if you like the name (`newComp->mpi_barrier_sleep = parseSleepDuration()` or something)

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
```diff
@@ -1847,6 +1862,46 @@ void Dimr::scanComponent(XmlTree* xmlComponent, dimr_component* newComp)
...
+        const char* s = mpiBarrierSleep->charData;
+        int value = 0;
+        auto res = std::from_chars(s, s + std::strlen(s), value);
+        if (res.ec == std::errc() && res.ptr != s)
```
harmenwierenga: The res.ptr != s is redundant, the res.ec (error code) will already be something non-default when it could not parse anything. I suggest `if (res.ec == std::errc{})`, or use the operator bool, `if(res.ec)`, although that may be a bit more difficult to read

harmenwierenga (commented):
Small comments

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done

src/engines_gpl/dimr/packages/dimr_lib/src/dimr.cpp
adrimourits: Done

adrimourits:
@harmenwierenga , my PR is green! Can you please resolve conversations and re-re-review?

harmenwierenga (approved):

PR #896, opened 2026-05-22
UNST-9885: create atmospheric stability module

sumihar:
# What was done 
- A module has been created to compute scaling parameters (u_star, t_star, q_star, etc) following Matlab prototype provided by Firmijn
- Unit tests have been developed comparing the results to the ones generated by the Matlab scripts; the remaining differences have been discussed with Firmijn and considered acceptable
- A routine has been added to compute momentum and heat fluxes based on the scaling parameters
- A set of getter routines have been developed to get arrays of scaling parameters and fluxes

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [X] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [X]	Documentation updated \
To be updated in UNST-9893 
- [ ]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9885


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+   public :: get_z0_momentum
+   public :: get_z0_heat
+   public :: get_z0_humidity
+   public :: get_monin_obukhov_length
+   public :: get_richardson_number
+   
+   real(kind=dp), parameter :: HEIGHT_WIND_VELOCITY = 10.0_dp ! sensor height for wind velocity [m]
+   real(kind=dp), parameter :: HEIGHT_TEMPERATURE = 2.0_dp ! sensor height for temperature [m]
+   real(kind=dp), parameter :: HEIGHT_HUMIDITY = 2.0_dp ! sensor height for specific humidity [m]
+   real(kind=dp), parameter :: CONST_GRAVITY = 9.80665_dp ! gravitational acceleration [m/s2]
```
FlorisBuwaldaDeltares, line 47: let's not introduce extra constants and just take the gravity defined in m_flowparameters

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+   real(kind=dp), parameter :: CONST_KAPPA = 0.4_dp ! von Kármán constant [-]
```
FlorisBuwaldaDeltares: we already have a von Karman constant i think in m_flowparameters

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+      real(kind=dp) :: richardson_number = 0.0_dp      !< Bulk Richardson number [-].
+      real(kind=dp) :: z0_momentum = 0.0_dp            !< Momentum roughness [m].
+      real(kind=dp) :: z0_heat = 0.0_dp                !< Heat roughness [m].
+      real(kind=dp) :: z0_humidity = 0.0_dp            !< Humidity roughness [m].
+   end type t_scales
+
+   !> Fluxes data type.
+   type :: t_fluxes
+      real(kind=dp) :: wind_stress_u = 0.0_dp
+      real(kind=dp) :: wind_stress_v = 0.0_dp
```
FlorisBuwaldaDeltares, line 86: are wind stresses fluxes? it seems strange to lump these together

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+   subroutine compute_scaling_parameters(wind_velocity_u, wind_velocity_v, air_temperature, dew_point_temperature, &
+                                  air_pressure, charnock, surface_temperature, result, options)
+      real(kind=dp), intent(in) :: wind_velocity_u         !< U wind component [m/s].
+      real(kind=dp), intent(in) :: wind_velocity_v         !< V wind component [m/s].
+      real(kind=dp), intent(in) :: air_temperature         !< Air temperature [K].
+      real(kind=dp), intent(in) :: dew_point_temperature   !< Dew-point temperature [K].
+      real(kind=dp), intent(in) :: air_pressure            !< Air pressure [Pa].
+      real(kind=dp), intent(in) :: charnock                !< Charnock parameter [-].
+      real(kind=dp), intent(in) :: surface_temperature     !< Surface temperature [K].
+      type(t_scales), intent(out) :: result                !< Computed scaling outputs.
```
FlorisBuwaldaDeltares: I would suggest a function that returns result

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+      integer :: iteration
+      integer, parameter :: MINIMUM_ITERATION = 5
+      integer, parameter :: MAXIMUM_ITERATION = 50
+      real(kind=dp), parameter :: EPSILON = 1.0e-4_dp
+
+      if (present(options)) then
+         runtime_options = options
+      else
+         runtime_options = t_options()
+      end if
```
FlorisBuwaldaDeltares: is this function ever called without options? I would rather always pass options, and having options be a default t_options() object if the user does not want to specify options much

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+
+      ! Initialize scaling parameters.
+      u_star = 0.04_dp * delta_wind_speed
+      t_star = 0.04_dp * delta_temperature
+      q_star = 0.04_dp * delta_specific_humidity
+      psi_momentum = 0.0_dp
+      psi_heat = 0.0_dp
+      psi_humidity = 0.0_dp
+      obukhov_length = 0.0_dp
+      richardson_number = 0.0_dp
```
FlorisBuwaldaDeltares, line 170: if the scaling parameters are initialized like this, why is the use of a scaling_parameters data type?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
+module m_atmospheric_stability
```
FlorisBuwaldaDeltares, line 1: I think the module could use with some explanation of what is happening here to add context, something like:


```suggestion
module m_atmospheric_stability
!! Computes bulk air-sea exchange fluxes (wind stress, latent and sensible heat)
!! using Monin-Obukhov Similarity Theory.
!!
!! The algorithm iteratively solves for the turbulence scaling parameters (u*, t*, q*)
!! by updating roughness lengths (Charnock + smooth flow) and stability corrections
!! (psi functions) until convergence. From these, surface fluxes are derived via:
!!
!!   tau   = rho * u*^2
!!   Qh    = rho * Cp * u* * t*
!!   Qe    = rho * Lv * u* * q*
!!
!! Inputs required:
!!   - Wind velocity components (u, v)
!!   - Air temperature
!!   - Dew-point temperature (used to derive vapor pressure and specific humidity)
!!   - Air pressure at surface
!!   - Sea surface temperature
!!   - Charnock parameter (wave-state dependent roughness)
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+         t_star_new = CONST_KAPPA * delta_temperature / log_denominator_heat
+         q_star_new = CONST_KAPPA * delta_specific_humidity / log_denominator_humidity
+
+         convergence_error = sqrt((u_star_new - u_star)**2 + (t_star_new - t_star)**2 + (q_star_new - q_star)**2)
+
+         u_star = u_star_new
+         t_star = t_star_new
+         q_star = q_star_new
+
+         if (iteration >= MINIMUM_ITERATION .and. convergence_error <= EPSILON) then
```
FlorisBuwaldaDeltares, line 242: this should be an OR! we should iterate until the convergence error is sufficiently small, or maximum iterations is reached

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+
+         u_star_new = CONST_KAPPA * delta_wind_speed / log_denominator_momentum
+         t_star_new = CONST_KAPPA * delta_temperature / log_denominator_heat
+         q_star_new = CONST_KAPPA * delta_specific_humidity / log_denominator_humidity
+
+         convergence_error = sqrt((u_star_new - u_star)**2 + (t_star_new - t_star)**2 + (q_star_new - q_star)**2)
+
+         u_star = u_star_new
+         t_star = t_star_new
+         q_star = q_star_new
```
FlorisBuwaldaDeltares: why bother with a u_star and a u_star new? why not just use result%u_star everywhere?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+   real(kind=dp), parameter :: CONST_Cpd = 7/2 * CONST_Rd ! specific heat capacity at constant pressure for dry air [J/(kg K)]
+   real(kind=dp), parameter :: CONST_E0 = 611.21_dp ! water vapour saturation pressure over water (Pa) at triple point temperature (Tt) [Pa]
+   real(kind=dp), parameter :: CONST_TT = 273.16_dp ! triple point temperature [K]
+   real(kind=dp), parameter :: OBUKHOV_LENGTH_LIMIT = 1.0e4_dp ! limit of Obukhov length to prevent numerical issues [m]
+   
+   !> Optional switches controlling physical parameterizations.
+   type :: t_options
+      logical :: include_free_convection = .false.  !< Include free convection.
+      logical :: include_fqsat = .false.            !< Include reduction factor of saturation specific humidity due to salt.
+      logical :: include_monin_obukhov = .true.     !< Include atmospheric stability in computing scaling parameters.
```
FlorisBuwaldaDeltares: there is a description of the options, but there should also be information on why it's an option. In what scenario do you want to include free convection, and why would you want to exclude it?

If an option should always be on, it shouldn't be an option. So every option needs to justify its existence

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+             end if
+             
+            log_denominator_momentum = log(HEIGHT_WIND_VELOCITY) - log(z0_momentum) - psi_momentum
+            log_denominator_heat = log(HEIGHT_WIND_VELOCITY) - log(z0_heat) - psi_heat
+            log_denominator_humidity = log(HEIGHT_WIND_VELOCITY) - log(z0_humidity) - psi_humidity
+         else
+            log_denominator_momentum = log(HEIGHT_WIND_VELOCITY) - log(z0_momentum)
+            log_denominator_heat = log(HEIGHT_TEMPERATURE) - log(z0_heat)
+            log_denominator_humidity = log(HEIGHT_HUMIDITY) - log(z0_humidity)
+         end if
```
FlorisBuwaldaDeltares, line 224: these expressions are identical save for the psi's, it would be cleaner to initialize them out of the if-statement, and then only add the psi's when relevant. Then it is immediately obvous that include_monin_obukhov adds psi's to the log denominators

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+
+   !> Compute roughness lengths following an ECMWF-style parameterization.
+   pure subroutine compute_roughness_lengths(u_star, charnock, z0_momentum, z0_heat, z0_humidity)
+      real(kind=dp), intent(in) :: u_star, charnock          !< u* [m/s], Charnock [-].
+      real(kind=dp), intent(out) :: z0_momentum, z0_heat, z0_humidity !< Roughness [m].
+      real(kind=dp), parameter :: ALPHA_M = 0.11_dp ! roughness length coefficient for momentum
+      real(kind=dp), parameter :: ALPHA_H = 0.40_dp ! roughness length coefficient for heat
+      real(kind=dp), parameter :: ALPHA_Q = 0.62_dp ! roughness length coefficient for humidity
+
+      z0_momentum = ALPHA_M*CONST_NU_AIR/u_star + charnock*u_star*u_star/CONST_GRAVITY
```
FlorisBuwaldaDeltares: Image

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+            log_denominator_momentum = log(HEIGHT_WIND_VELOCITY) - log(z0_momentum)
+            log_denominator_heat = log(HEIGHT_TEMPERATURE) - log(z0_heat)
+            log_denominator_humidity = log(HEIGHT_HUMIDITY) - log(z0_humidity)
+         end if
+
+         u_star_new = CONST_KAPPA * delta_wind_speed / log_denominator_momentum
+         t_star_new = CONST_KAPPA * delta_temperature / log_denominator_heat
+         q_star_new = CONST_KAPPA * delta_specific_humidity / log_denominator_humidity
+
+         convergence_error = sqrt((u_star_new - u_star)**2 + (t_star_new - t_star)**2 + (q_star_new - q_star)**2)
```
FlorisBuwaldaDeltares: Image

I agree with copilot that it seems strange to simply sum 3 variables that have different scales

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+      real(kind=dp) :: stability_correction            !< Stability correction [-].
+      real(kind=dp), parameter :: coef_a = 1.0_dp
+      real(kind=dp), parameter :: coef_b = 2.0_dp/3.0_dp
+      real(kind=dp), parameter :: coef_c = 5.0_dp
+      real(kind=dp), parameter :: coef_d = 0.35_dp
+      real(kind=dp), parameter :: coef_unstable = 16.0_dp
+      real(kind=dp) :: clipped_zeta, unstable_factor, psi_unstable, psi_stable
+
+      clipped_zeta = min(stability_parameter, 5.0_dp)
+      unstable_factor = (1.0_dp - coef_unstable*clipped_zeta)**0.25_dp
```
FlorisBuwaldaDeltares: if coef_unstable*clipped_zeta > 0 we are calculating an imaginary number here and unstable_factor will be NaN. unstable factor should only be calculated when it is a real number

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+                delta_specific_humidity = humidity_at_wind_height - surface_humidity
+             end if
+             
+            log_denominator_momentum = log(HEIGHT_WIND_VELOCITY) - log(z0_momentum) - psi_momentum
+            log_denominator_heat = log(HEIGHT_WIND_VELOCITY) - log(z0_heat) - psi_heat
+            log_denominator_humidity = log(HEIGHT_WIND_VELOCITY) - log(z0_humidity) - psi_humidity
+         else
+            log_denominator_momentum = log(HEIGHT_WIND_VELOCITY) - log(z0_momentum)
+            log_denominator_heat = log(HEIGHT_TEMPERATURE) - log(z0_heat)
+            log_denominator_humidity = log(HEIGHT_HUMIDITY) - log(z0_humidity)
```
FlorisBuwaldaDeltares, line 223: do we have a guarantees that z0_momentum =/ HEIGHT_WIND_VELOCITY etc? these denominators cannot be 0 or we get NaN's

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+
+      number_of_elements = size(fluxes)
+      allocate(wind_stress_u(number_of_elements))
+      allocate(wind_stress_v(number_of_elements))
+
+      do index = 1, number_of_elements
+         wind_stress_u(index) = fluxes(index)%wind_stress_u
+         wind_stress_v(index) = fluxes(index)%wind_stress_v
+      end do
+   end subroutine get_wind_stress
```
FlorisBuwaldaDeltares, line 337: I really dislike all these getters, why not just access module variables directly?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+      real(kind=dp) :: u_star = 0.0_dp                 !< Friction velocity [m/s].
+      real(kind=dp) :: t_star = 0.0_dp                 !< Temperature scale [K].
+      real(kind=dp) :: q_star = 0.0_dp                 !< Humidity scale [kg/kg].
+      real(kind=dp) :: w_star = 0.0_dp                 !< Gustiness velocity [m/s].
+      real(kind=dp) :: monin_obukhov_length = 0.0_dp   !< Obukhov length [m].
+      real(kind=dp) :: richardson_number = 0.0_dp      !< Bulk Richardson number [-].
+      real(kind=dp) :: z0_momentum = 0.0_dp            !< Momentum roughness [m].
+      real(kind=dp) :: z0_heat = 0.0_dp                !< Heat roughness [m].
+      real(kind=dp) :: z0_humidity = 0.0_dp            !< Humidity roughness [m].
+   end type t_scales
```
FlorisBuwaldaDeltares, line 81: the AI says the outputs should be u*, t* and q*, I see many other outputs. How will this be used in Dflowfm? we don't have to output variables we will never use

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -0,0 +1,643 @@
...
+   !> Compute turbulence scaling parameters for a point value.
+   subroutine compute_scaling_parameters(wind_velocity_u, wind_velocity_v, air_temperature, dew_point_temperature, &
+                                  air_pressure, charnock, surface_temperature, result, options)
+      real(kind=dp), intent(in) :: wind_velocity_u         !< U wind component [m/s].
+      real(kind=dp), intent(in) :: wind_velocity_v         !< V wind component [m/s].
+      real(kind=dp), intent(in) :: air_temperature         !< Air temperature [K].
+      real(kind=dp), intent(in) :: dew_point_temperature   !< Dew-point temperature [K].
+      real(kind=dp), intent(in) :: air_pressure            !< Air pressure [Pa].
+      real(kind=dp), intent(in) :: charnock                !< Charnock parameter [-].
+      real(kind=dp), intent(in) :: surface_temperature     !< Surface temperature [K].
```
FlorisBuwaldaDeltares, line 105: it seems strange to have all the inputs be individual variables but then to lump the outputs in a t_scales object. I'd either make this a compute_scaling_parameters(input, output) or also unpack the output object. Mixing the strategies does not make sense

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_atmospheric_stability.f90
```diff
@@ -0,0 +1,166 @@
...
+   use assertions_gtest
+   use precision, only: dp
+   use precision_basics, only: comparereal
+   use m_atmospheric_stability
+   implicit none
+
+contains
+
+   !$f90tw TESTCODE(TEST, test_atmospheric_stability, test_compute_scales_and_fluxes_moninobukhov_true, test_compute_scales_and_fluxes_moninobukhov_true,
+   subroutine test_compute_scales_and_fluxes_moninobukhov_true() bind(C)
```
FlorisBuwaldaDeltares, line 11: where did you get these numbers from? the idea of a unit test is usually that the result is trivially to verify by hand. If this unit test would ever fail, it would require difficult debugging to figure out what calculation exactly is going wrong. However if this is the only unit test we can write it is better than nothing I suppose.

FlorisBuwaldaDeltares (changes requested):
Hi @sumihar , nice work but I do have a few comments. Let me know if you have any questions!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Description is added.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: We agreed the following, during our Teams video call just now:
- modify variable gravity in m_GlobalParameters, from 9.81d0 to 9.80665d0
- use that gravity in m_atmospheric_stability

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Now use m_physcoef, only: vonkarw

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: More description is added.

In the first step, only option include_monin_obukhov will be supported via mdu. The other two options include_free_convection and include_fqsat are intended for the future; the relevant codes have been programmed just to anticipate.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: As we discussed in our video call, the other output variables will be needed in his/map output files, as diagnostic variables.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Wind stresses are momentum fluxes.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: My idea was to prepare the inputs as how dflow-fm will provide them (as separate arrays); the output variables are not yet definitive; lumping them in a t_scales object makes it easier to extend, without modifying the function definition.

We agreed to postpone this point until we reached the point to implement / connect this module in dflow-fm.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Suggestion is applied.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Suggestion is applied. 

We also agreed to discuss this further when implementing this module in dflow-fm, if necessary.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: We agreed to leave it as it is now.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Notice also that, besides the use of psi, they differ also in sensors' height for temperature and humidity. When include_monin_obukhov = true, all the sensor heights are converted to HEIGHT_WIND_VELOCITY. Otherwise, the original HEIGHT_TEMPERATURE and HEIGHT_HUMIDITY are used in determining the log_denominator.

Hence, we can not implement the suggestion.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: The loop is written as a do loop with MAXIMUM_ITERATION. It requires also a MINIMUM_ITERATION. Hence the stopping criteria.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: The getters are helper functions to prepare the output variables, from arrays of derived data type to fortran arrays. 

We agreed to discuss this further when we are implementing the module in dflow-fm.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_atmospheric_stability.f90
sumihar: The numbers are copied from input and output used in Firmijn's Matlab prototype. Three cases are chosen that represent situations with low and high wind speed. Here the unit tests are intended to show / check whether the results match the results of Firmijn's prototype. Also to identify early if future code changes affect the results.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: A threshold is introduced: SMALL_NUMBER = 10e-8, to prevent this potential issue from happening; it's  used as e.g.

log_denominator_momentum = sign(max(abs(log_denominator_momentum), SMALL_NUMBER), log_denominator_momentum)

Firmijn has been contacted if he has suggestions.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Suggestion has been applied.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: The following has been introduced:

inverse_u_star = 1.0_dp / sign(max(abs(u_star), SMALL_NUMBER), u_star)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Good catch! The calculations have been moved to inside the respective if-block.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: Relative error has been used here. 

Note that tests have been done using the Matlab prototype: the use of relative error makes the scales of the 3 elements more comparable to each other. However, it leads to slightly different results. This will be discussed with Firmijn.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -39,20 +57,22 @@ module m_atmospheric_stability
    real(kind=dp), parameter :: CONST_E0 = 611.21_dp ! water vapour saturation pressure over water (Pa) at triple point temperature (Tt) [Pa]
    real(kind=dp), parameter :: CONST_TT = 273.16_dp ! triple point temperature [K]
    real(kind=dp), parameter :: OBUKHOV_LENGTH_LIMIT = 1.0e4_dp ! limit of Obukhov length to prevent numerical issues [m]
+   real(kind=dp), parameter :: SMALL_NUMBER = 1.0e-8_dp
```
FlorisBuwaldaDeltares, line 60: usually we call small number some form of epsilon, but I guess small number is pretty descriptive too haha

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
FlorisBuwaldaDeltares: oh i didn't notice! okay, I guess then this is as clear as it can be. Thanks for checking

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
FlorisBuwaldaDeltares: I see it, then this is ok :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_atmospheric_stability.f90
FlorisBuwaldaDeltares: I guess it is good to test whether the fortran implementation matches matlab!

src/utils_gpl/flow1d/packages/flow1d_core/src/GlobalParameters.f90
```diff
@@ -35,7 +35,7 @@ module m_GlobalParameters
    
    public
    
-   double precision                 :: gravity                       = 9.81d0
+   double precision                 :: gravity                       = 9.80665d0
```
FlorisBuwaldaDeltares: I hope this doesn't make the whole testbench red. If it does, we could consider reverting this change with a comment that the atmospheric stability module uses a fixed, more precise earth gravity

FlorisBuwaldaDeltares (approved):
I julius, i think for now this is good! if we want to change things later when we're connecting things to FM that's fine. For now we have a module that explains what it calculates with some tests that verify the result. Nice work!

src/utils_gpl/flow1d/packages/flow1d_core/src/GlobalParameters.f90
hrajagers: Note that we have also `ag` defined in `m_physcoef` as `9.81_dp`. That value is probably used in more places of the 2D/3D code than this value defined in `flow1d_core` ... Would love to have only one definition for gravity ...

hrajagers: By the way, gravity fluctuates around the earth. Wikipedia indicates: "At different points on Earth's surface, the free fall acceleration ranges from 9.764 to 9.834 m/s2" ... and continues: "A conventional [standard value](https://en.wikipedia.org/wiki/Standard_gravity) is defined exactly as 9.80665 m/s² (about 32.1740 ft/s²)." That's the value adopted here, but the question whether these last decimals are useful when applied around the world ...

sumihar: Hi @hrajagers,
thanks for your input. I agree that it is important to have one gravity definition. I have registered this point in UNST-9970 to follow up again later.

sumihar: It turns out that there are 43 failed tests in Windows and 33 in Linux due to this change in gravity.

I will revert the change and introduce a local CONST_GRAVITY = 9.80665 as required by Firmijn. In the follow up issue UNST-9970, we will consolidate this.

FlorisBuwaldaDeltares (approved):

PR #900, opened 2026-05-22
UNST-9952 fix time reading from mdu

jreyns:
# What was done 

Moved the [time] read block in readMDU up in the routine, behind the geometry block. Some output fields use tstart_user and tstop_user, which was modified by the [time] read after fields  were already set based on the m_flowtimes defaults.
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ x]	Clear from the issue description 
- [ ]	Not applicable 

# Tests 
- [ x] Tests updated \
e02_f042_c01 and c02
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9952


FlorisBuwaldaDeltares (commented):
hi @jreyns, your issue description states some input fields already use the tstart and tstop variables before they are defined, which ones are those exactly? I'd like the order in which mdu arguments are read not to matter, so move towards a reading step and a separate parsing/initialization step based on what was read. 

additionally, it is not clear to me what this change has to do with sedtrails

jreyns:
> hi @jreyns, your issue description states some input fields already use the tstart and tstop variables before they are defined, which ones are those exactly? I'd like the order in which mdu arguments are read not to matter, so move towards a reading step and a separate parsing/initialization step based on what was read.
> 
> additionally, it is not clear to me what this change has to do with sedtrails

Hi @FlorisBuwaldaDeltares, I was asked to debug why a Western Scheldt run that should generate sedtrails-related nc output did not. My analysis showed that the (sedtrails_dt,sedtrails_start,sedtrails_stop) triplet in the [output] block was set using the tstart_user and tstop_user default values from m_flowtimes. The updated values were not read yet, as the [time] block in readMDU is read after the [output] block. So this should affect all [output] items that call 

`call set_time_interval(ti_st_array, ti_sts, ti_st, ti_ste, tstart_user, tstop_user, success)`

FlorisBuwaldaDeltares (approved):
after your explanation, i agree with the changes. I still think all set_time_window calls should move out of readmdufile, but i won't ask you to refactor that in this MR.

Just a few failing cases in your pipeline, good luck with fixing them!

PR #904, opened 2026-05-26
Fix bug with setting the "trigger.type" parameter in the windows environment build configurations.

rene-deltares:
# What was done 

The windows build environment configurations set a parameter based on what kind of trigger is used. So if the 'scheduled' trigger is used, a parameter 'trigger.type' gets value 'schedule', and when the 'git' trigger is used, the parameter gets value 'vcs'. But it turns out this was not working properly. The build parameter is created, but it's value is empty. See: https://dpcbuild.deltares.nl/buildConfiguration/Delft3D_WindowsBuildEnvironmentI24/7264047?buildTab=buildParameters&buildParametersTab=parameters

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


harmenwierenga (approved):
Nice improvements!

PR #397, opened 2025-11-20
UNST-9434 merge branch sedmor

jreyns:
# What was done 
Add various functionality for storm impact modeling to main

# Evidence of the work done 

- [ ]	Video/figures \
<add video/figures if applicable> 
- [x ]	Clear from the issue description 
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
<add testcase numbers if applicable, Issue number>
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
to do: [UNST-9452](https://issuetracker.deltares.nl/browse/UNST-9452) 
- [ ]	Not applicable 

# Issue link

PR #907, opened 2026-05-27
sedmor branch: morphology bugfixes and minor improvements

hrajagers:
# Implement sedmor-branch morphology improvements:
- `apply_sediment_bc`: bugfix for 3D boundary conditions for suspended sediment
- `fm_erosed`: bugfix: cohesive sediment should use layer `kmaxlc`, not layer `1`

- `morphology_data_module`, `rdsed`, `erosed`, `z_erosed`, `fm_erosed`, `m_fm_erosed`, `inipointers_erosed`: introduction of _diffusionCal_ (`seddif_cal`) and _diffusionScaling_ (`difparam`) to calibrate the suspended sediment diffusion coefficient in vertical direction over the whole water column and below aks, respectively (note: keywords moved from [sediment] block in mdu-file to [sedimentOverall] block in sed-file. Applied in both D-Flow FM and Delft3D-FLOW).
 
- `fm_erosed`: add `_dp` specifier, use `epshu`  instead of `epshs` for minimum layer thickness
- `fm_erosed`: check on `deltas` _only in case of waves_ **(this change affects the test bench cases e02_f039_c03, c04, c05, c06, c11)**
- `fm_fallve`: comment added indicating that `kmx==1` case hasn't been dealt with yet
- `m_fm_bott3d`: ignore near bed suspended load flux below `aksu` reference height if `aksu` is really small (<0.1nm)

- `unstruc_model`: maintenance: sediment keywords adjusted to match `lowerCamelCase` convention

# What was done 

- Reverted commit https://github.com/Deltares/Delft3D/commit/702fda86df29738e11dc58235cd5d18e879bf4a5 to revive all changes originally implemented in #397.
- Subdivided all these changes into logical sets of changed files and reverted those individually, such that they can be revived one by one. The final code after these actions is the same code as what we started with.
- Reverted one commit related to bugfixes of and small improvements to the morphology code.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
New references for e02_f039_c03, c04, c05, c06, c11
- [ ]	Not applicable 

# Documentation  
- [x]	Documentation updated \
New keywords added to sed-file documentation
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -1215,15 +1215,20 @@ subroutine fm_erosed()
                                  &  aks_ss3d, sourse(nm, l), sour_im(nm, l),              &
                                  &  sinkse(nm, l))
```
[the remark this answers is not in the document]
hrajagers: @jreyns This comment makes sense to me. What do you think? Shall I move it in front of the `soursin_3d` call.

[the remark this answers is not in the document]
jreyns: i think it will blow up the sourse and sour_im terms without compensating sink, and probably give more bed level change in erosive conditions. We would have to test on the analytical rouse tests. I would not move it yet, but make an issue and investigate the results.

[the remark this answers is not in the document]
hrajagers: I expect that moving `diffusionCal` should work fine, but `diffusionScaling` should be left after `soursin_3d`. Anyway, I think that it's a good idea to investigate this in a new issue: UNST-10025. The test case should also be added to the test bed.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -992,9 +996,9 @@ subroutine fm_erosed()
                   end if
                end if
                !
-               kmaxsd = 1 ! for mud fractions kmaxsd points to the grid cell at the bottom of the water column
-               thick0 = max(thicklc(kmaxsd) * h0, epshs)
-               thick1 = max(thicklc(kmaxsd) * h1, epshs)
+               kmaxsd = kmaxlc ! for mud fractions kmaxsd points to the grid cell at the bottom of the water column
+               thick0 = max(thicklc(kmaxsd) * h0, epshu)
+               thick1 = thicklc(kmaxsd) * h1
```
V-Chavarrias, line 998: These code lines are repeated below. It may be overdoing it, but it could be useful to have a two-lines function to set this. Then, if it is changed in the future, we make sure it is changed at both locations. Feel free to ignore this.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -490,7 +494,7 @@ subroutine fm_erosed()
             do k = kb, kt
                zcc = 0.5_dp * (zws(k - 1) + zws(k)) ! cell centre position in vertical layer admin, using absolute height
                kmxvel = k
-               if (zcc >= (bl(kk) + maxdepfrac * hs(kk)) .or. zcc >= (bl(kk) + deltas(kk))) then
+               if (zcc >= (bl(kk) + maxdepfrac * hs(kk)) .or. (jawave /= NO_WAVES .and. zcc >= (bl(kk) + deltas(kk)))) then
```
V-Chavarrias, line 494: It matches the description, but I do not understand why is this needed.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -1215,15 +1218,20 @@ subroutine fm_erosed()
                                  &  aks_ss3d, sourse(nm, l), sour_im(nm, l),              &
                                  &  sinkse(nm, l))
                   !
+                  if (seddif_cal > 0.0_fp) then
```
V-Chavarrias, line 1218: Is the default `seddif_cal` equal to 1? do we need the if-clause?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
V-Chavarrias, line 692: `]epshu sedthresh]` What is this supposed to say? `epshu<h< sedthresh` ?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -1215,15 +1218,20 @@ subroutine fm_erosed()
...
                   ! Impose relatively large vertical diffusion
                   ! coefficients for sediment in layer interfaces from
                   ! bottom of reference cell downwards, to ensure little
                   ! gradient in sed. conc. exists in this area.
-
-                  difbot = 10.0_fp * ws(kmxsed(nm, l) - 1, l) * thick1
-                  do kk = kb - 1, kmxsed(nm, l) - 1
-                     seddif(l, kk) = difbot
-                  end do
+                  if (difparam > 0.0_fp) then
```
V-Chavarrias, line 1226: Idem as above. Is it better to have a default of 1 and no if-clause?

V-Chavarrias: What if `difparam=0`? Then `seddif` will have the value it had before. Is the allocated 0? Or filled in between?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_fallve.f90
```diff
@@ -185,6 +185,7 @@ subroutine fm_fallve()
          end if
          !
          ! loop over the interfaces in the vertical
+         ! NOTE: kmx==1 case is not handled yet
          !
          if (kmx > 0) then ! 3D
```
V-Chavarrias, line 190: Here `kmx>0` is used to identify 3D while below `kmx>1` is used for the same end. It is written in the function that the case `kmx=1` is not handled, but I am not sure this is the best solution. 

Would it be an option to consider 3D for now `kmx>1` and hence tream `kmx=1` as a 2D case? 

Also, I would propose a parameter `THREE_D` or similar computed with `if (kmx>1)` (if you agree) and use that for all switches. Then, if we change our mind in the future it becomes easy and flawless to change.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/m_fm_bott3d.f90
```diff
@@ -427,6 +427,10 @@ subroutine fm_suspended_sand_correction()
                            aksu = ac1 * aks(k1, l) + ac2 * aks(k2, l)
                         end if
                         !
+                        if (aksu < 1e-10_dp) then
```
V-Chavarrias: Introduce parameter.

src/engines_gpl/flow2d3d/packages/flow2d3d_kernel/src/compute_sediment/erosed.f90
```diff
@@ -1270,15 +1274,21 @@ subroutine erosed(nmmax     ,kmax      ,icx       ,icy       ,lundia    , &
                                &  rhosol(l)         ,caks_ss3d      ,ws(nm,kmaxsd,l)    , &
                                &  aks_ss3d          ,sourse(nm,l)   ,sour_im(nm,l)      , &
                                &  sinkse(nm,l) )
+                !
```
V-Chavarrias, line 1277: Same comments as for FM.

src/engines_gpl/flow2d3d/packages/flow2d3d_kernel/src/compute_sediment/z_erosed.f90
```diff
@@ -1261,15 +1265,21 @@ subroutine z_erosed(nmmax     ,kmax      ,icx       ,icy       ,lundia    , &
                                &  rhosol(l)         ,caks_ss3d      ,ws(nm,kmxsed(nm,l),l)      , &
                                &  aks_ss3d          ,sourse(nm,l)   ,sour_im(nm,l)      , &
                                &  sinkse(nm,l) )
+                !
```
V-Chavarrias, line 1268: Same comments. 

The shape of the arrays is different, but I wonder whether there is a technical solution to share the code between FM and D3D. I.e., a function or subroutine `apply_seddif_calibration` or similar that is called by both.

src/utils_gpl/morphology/packages/morphology_data/src/morphology_data_module.f90
```diff
@@ -1288,6 +1289,8 @@ subroutine nullsedpar(sedpar)
       sedpar%tfloc = 1e-10_fp
       sedpar%d_micro = 1e-4_fp
       sedpar%ustar_macro = 0.067_fp
+      sedpar%seddif_cal = 0.0_fp
```
V-Chavarrias, line 1292: See comment about default.

V-Chavarrias (commented):
Nice work. Thanks.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/apply_sediment_bc.f90
```diff
@@ -47,114 +47,70 @@ subroutine apply_sediment_bc()
...
+      integer :: kk !< 3D boundary index
+      integer :: Lb !< bottom-most 3D flow link index
+      integer :: Lt !< top-most 3D flow link index
+      integer :: LLL !< 2D boundary flow link index
 
-      ! New approach: default Neumann, unless time series available
-      !
-      ! Find sand fractions
+      ! default Neumann boundary condition applied for all suspended sediments
       do ll = 1, lsed ! sediment-fraction index
```
jreyns, line 67: Maybe a good time to rename this pointer to lsedsus. I have noticed that non-mor people have used lsed as a local counter in various places. This can easily cause crashes if not careful.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
```diff
@@ -172,6 +172,8 @@ subroutine fm_erosed()
       real(fp) :: delr
       real(fp) :: di50
       real(fp) :: difbot
+      real(fp), pointer :: seddif_cal
```
jreyns: Move these and the pointering below to m_fm_erosed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_fallve.f90
hrajagers: Handling 3D one layer as 2D is probably logical, but the value still needs to be specified in the correct place. I accept the code as is, which identifies the issue, to avoid scope creep. I have created UNST-10038 to address this throughout the FM code.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
jreyns: If you don't add jawave it always evaluates to true, as deltas==0

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
jreyns: epshu<h<=sedthr

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
jreyns: seems overkill to me

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
hrajagers: This case is fundamentally different than the `seddif_cal` case. The default is `10` by the way, so by default this process is active and has been active since the first introduction of this term over 25 years ago. The `difparam=1` case is fundamentally different from the `don't do this` case (now linked to `difparam <= 0`. If we don't do this, the `seddif` value remains equal to the original `seddif` value, optionally multiplied by `seddif_cal` above.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
jreyns: yes, because I want to be able to switch it off if I need to

jreyns (commented):
Thanks Bert, just a few remarks

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
hrajagers: The `seddif_cal=1` and `seddif_cal<=0` cases should give the same results. Just using `1` instead of `0` to switch this off, makes the code conceptually simpler. However, you would do `kmx` computations instead of one logical evaluation. If you try to get maximum performance, this might implementation be slightly better.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
hrajagers: This feels like an overkill since we would have to pass 5 input arguments and 2 output arguments, and it would obfuscate what is happening here ...

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/m_fm_bott3d.f90
hrajagers: I know this is indeed the current rule and fully agree with it for thresholds that are really reusable, but here ... sigh ... I'm really wondering how much compute time we actually save by adding this check. I have introduced a local parameter `AKSU_THRESHOLD_FOR_SCORR_COMPUTATION`.

src/engines_gpl/flow2d3d/packages/flow2d3d_kernel/src/compute_sediment/erosed.f90
hrajagers: Same reaction as for FM.

src/engines_gpl/flow2d3d/packages/flow2d3d_kernel/src/compute_sediment/z_erosed.f90
hrajagers: Same reaction.

Code could be shared between `erosed` and `z_erosed`, but let's first determine where most of the sediment simulation time is spend before cutting the code into pieces.

src/utils_gpl/morphology/packages/morphology_data/src/morphology_data_module.f90
hrajagers: See reaction where applied.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
hrajagers: pointers defined in `m_fm_erosed`, set in `inipointers_erosed`, and used here.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/apply_sediment_bc.f90
hrajagers: I have created UNST-10039 for this request. I would like to move forward with the merging of your branch, and this "minor" change will probably require more code adjustments than I anticipate.

hrajagers (commented):
Replies added and code updated.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_sediment/fm_erosed.f90
V-Chavarrias: My question is why is that a problem.

V-Chavarrias (approved):
Thanks.

PR #908, opened 2026-05-27
DEVOPSDSC-969: Add system libexpat.so to dimrset/lib (quick fix)

robin-deltares:
So people can merge while we find a permanent solution.

https://issuetracker.deltares.nl/browse/DEVOPSDSC-969


rene-deltares (approved):
Urgent quick fix

PR #909, opened 2026-05-27
All/task/devopsdsc 933 fix windows 2025 build

BartSteurDeltares:
Created a new dockerfile for the build container using 2025 base image and intel 2025 suite. Now the windows build can run on the windows server 2025 agents.

### DIFF
--- a/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i25
+++ b/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i25
@@ -1,0 +1,54 @@
+FROM containers.deltares.nl/mcr-proxy/windows/server:ltsc2025
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'Continue'; $verbosePreference='Continue';"]
+
+RUN reg add "HKLM\SYSTEM\CurrentControlSet\Control\FileSystem" /v "LongPathsEnabled" /t REG_DWORD /d 1 /f
+
+SHELL ["cmd", "/S", "/C"]
+
+ADD vs_Community.exe C:/
+
+RUN (start /w vs_community.exe --passive --wait --noUpdateInstaller --norestart --nocache \
+    --installPath "C:\\Program Files\\Microsoft Visual Studio\\17\\Community" \
+    --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended \
+    --add Microsoft.VisualStudio.Component.Windows11SDK.26100 \
+    --add Microsoft.VisualStudio.Component.VC.ATLMFC \
+    --add Microsoft.VisualStudio.Component.VC.CLI.Support \
+    --remove Microsoft.VisualStudio.Component.Windows10SDK.10240 \
+    --remove Microsoft.VisualStudio.Component.Windows10SDK.10586 \
+    --remove Microsoft.VisualStudio.Component.Windows10SDK.14393 \
+    --remove Microsoft.VisualStudio.Component.Windows81SDK \
+    || IF "%ERRORLEVEL%"=="3010" EXIT 0)
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'Continue'; $verbosePreference='Continue';"]
+
+# Install the Intel toolkits
+ADD intel-fortran-compiler-2025.3.3.16_offline.exe C:/
+RUN Start-Process -FilePath "C:\intel-fortran-compiler-2025.3.3.16_offline.exe" -ArgumentList "--silent", "-a", "--action", "install", "--eula=accept", "-p=NEED_VS2022_INTEGRATION=1", "-s" -Wait
+
+ADD intel-onemkl-2025.3.1.10_offline.exe C:/
+RUN Start-Process -FilePath "C:\intel-onemkl-2025.3.1.10_offline.exe" -ArgumentList "--silent", "-a", "--action", "install", "--eula=accept", "-p=NEED_VS2022_INTEGRATION=1", "-s" -Wait
+
+ADD intel-mpi-2021.17.2.93_offline.exe C:/
+RUN Start-Process -FilePath "C:\intel-mpi-2021.17.2.93_offline.exe" -ArgumentList "--silent", "-a", "--action", "install", "--eula=accept", "-s" -Wait
+
+# RUN Remove intel and visual studio executables
+RUN Remove-Item C:\*.exe"
+
+ADD cmake-3.31.0-rc1-windows-x86_64.msi C:\\cmake-3.31.0-rc1-windows-x86_64.msi
+RUN Start-Process msiexec -Wait -ArgumentList '/i C:\\cmake-3.31.0-rc1-windows-x86_64.msi /qn'
+RUN setx /M PATH $($Env:PATH + ';C:\\Program Files\\CMake\\bin')
+RUN Remove-Item C:\\cmake-3.31.0-rc1-windows-x86_64.msi
+
+ENV ONEAPI_ROOT="C:\\Program Files (x86)\\Intel\\oneAPI\\"
+ENV VS2022INSTALLDIR="C:\\Program Files\\Microsoft Visual Studio\\17\\Community\\"
+
+ADD python-3.12.7-amd64.exe C:\\python-3.12.7-amd64.exe
+RUN Start-Process python-3.12.7-amd64.exe -ArgumentList '/quiet InstallAllUsers=1 PrependPath=1' -Wait
+RUN Remove-Item C:\\python-3.12.7-amd64.exe
+RUN New-Item -ItemType SymbolicLink -Path 'C:\\Program Files\\Python312\\python3.exe' -Target 'C:\\Program Files\\Python312\\python.exe'
+
+ADD Git-2.47.0-64-bit.exe C:\\Git-2.47.0-64-bit.exe
+RUN Start-Process -FilePath "C:\\Git-2.47.0-64-bit.exe" -ArgumentList "/SILENT" -Wait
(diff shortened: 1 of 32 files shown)


ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i25
```diff
@@ -0,0 +1,54 @@
...
+
+ADD cmake-3.31.0-rc1-windows-x86_64.msi C:\\cmake-3.31.0-rc1-windows-x86_64.msi
+RUN Start-Process msiexec -Wait -ArgumentList '/i C:\\cmake-3.31.0-rc1-windows-x86_64.msi /qn'
+RUN setx /M PATH $($Env:PATH + ';C:\\Program Files\\CMake\\bin')
+RUN Remove-Item C:\\cmake-3.31.0-rc1-windows-x86_64.msi
+
+ENV ONEAPI_ROOT="C:\\Program Files (x86)\\Intel\\oneAPI\\"
+ENV VS2022INSTALLDIR="C:\\Program Files\\Microsoft Visual Studio\\17\\Community\\"
+
+ADD python-3.12.7-amd64.exe C:\\python-3.12.7-amd64.exe
```
rene-deltares, line 46: I would take another look at `Dockerfile-dhydro-vs2022-i24`. Recently we've stopped using this python installer. We also install `conan` in the build container now, which we will need very soon. It should be a straightforward change.

ci/teamcity/Delft3D/windows/buildInteracter.kt
```diff
@@ -62,13 +62,15 @@ object WindowsBuildDflowfmInteracter : BuildType({
...
-                call C:/set-env-vs2022.cmd
+                call "C:\\Program Files (x86)\\Intel\\oneAPI\\setvars.bat" --force
+                call "C:\\Program Files\\Microsoft Visual Studio\\17\\Community\\Common7\\Tools\\VsDevCmd.bat" -arch=amd64 -host_arch=amd64
 
                 cmake ./src/cmake -G %generator% -T fortran=%intel_fortran_compiler% -D CMAKE_BUILD_TYPE=%build_type% -D CONFIGURATION_TYPE:STRING=%product% -B build_%product% -D CMAKE_INSTALL_PREFIX=build_%product%/install -D ENABLE_CODE_COVERAGE=%enable_code_coverage_flag%
 
                 cd build_%product%
 
-                cmake --build . -j --target install --config %build_type%
+                cmake --build . -j --config %build_type%
```
rene-deltares, line 72: I wouldn't split this `cmake --build` command up in two commands. Every build requires a `--target`. If you leave it out, the default is just `--target all` if I'm not mistaken. We need `--target install`, so better to specify it that way.

ci/teamcity/Delft3D/windows/build2D3DSinglePrecission.kt
```diff
@@ -84,21 +84,21 @@ object WindowsBuild2D3DSP : BuildType({
...
-                call C:/set-env-vs2022.cmd
+                call "C:\\Program Files (x86)\\Intel\\oneAPI\\setvars.bat" --force
+                call "C:\\Program Files\\Microsoft Visual Studio\\17\\Community\\Common7\\Tools\\VsDevCmd.bat" -arch=amd64 -host_arch=amd64
 
                 cmake ./src/cmake -G %generator% -T fortran=%intel_fortran_compiler% -D CMAKE_BUILD_TYPE=%build_type% -D CONFIGURATION_TYPE:STRING=flow2d3d -B build_flow2d3d -D CMAKE_INSTALL_PREFIX=build_flow2d3d/install -D ENABLE_CODE_COVERAGE=%enable_code_coverage_flag%
 
                 cd build_flow2d3d
 
-                cmake --build . -j --target install --config %build_type%
+                cmake --build . -j --config %build_type%
```
rene-deltares, line 94: Repeat comment: I wouldn't split this cmake --build command up in two commands. Every build requires a --target. If you leave it out, the default is just --target all if I'm not mistaken. We need --target install, so better to specify it that way.

ci/teamcity/Delft3D/windows/build.kt
```diff
@@ -75,11 +75,16 @@ object WindowsBuild : BuildType({
...
             scriptContent = """
-                call C:/set-env-vs2022.cmd
+                call "C:\\Program Files (x86)\\Intel\\oneAPI\\setvars.bat" --force
+                call "C:\\Program Files\\Microsoft Visual Studio\\17\\Community\\Common7\\Tools\\VsDevCmd.bat" -arch=amd64 -host_arch=amd64
+
                 cmake ./src/cmake -G %generator% -T fortran=%intel_fortran_compiler% -D CMAKE_BUILD_TYPE=%build_type% -D CONFIGURATION_TYPE:STRING=%product% -B build_%product% -D CMAKE_INSTALL_PREFIX=build_%product%/install -D ENABLE_CODE_COVERAGE=%enable_code_coverage_flag%
                 if %%errorlevel%% neq 0 exit /b %%errorlevel%%
 
-                cmake --build ./build_%product% -j --target install --config %build_type%
+                cmake --build ./build_%product% -j --config %build_type%
```
rene-deltares, line 84: Repeat comment: I wouldn't split this cmake --build command up in two commands. Every build requires a --target. If you leave it out, the default is just --target all if I'm not mistaken. We need --target install, so better to specify it that way.

ci/teamcity/Delft3D/windows/build.kt
```diff
@@ -75,11 +75,16 @@ object WindowsBuild : BuildType({
         script {
             name = "Build"
             scriptContent = """
-                call C:/set-env-vs2022.cmd
+                call "C:\\Program Files (x86)\\Intel\\oneAPI\\setvars.bat" --force
```
rene-deltares, line 78: I am pretty sure we cannot switch our builds to use the oneapi 2025 container immediately. The biggest reason is that we will want to build all of our third-party library dependencies with the oneapi 2025 toolchain as well. We are working on that currently with our conan setup. I suspect we will get different packages depending on if you're using oneapi 2024 or oneapi 2025 in the near future.

I would remove the changes to `build.kt`, `build2D3DSinglePrecission.kt` and `buildInteracter.kt` for now and just keep using the oneapi2024 build container there. And instead I would suggest to make it possible first to build and push an intel oneapi 2025 container in the 'build environment' build configuration. We will need that one to build and push the conan packages in the near future.

ci/teamcity/Delft3D/windows/buildEnvironment-i25.kt
```diff
@@ -8,7 +8,7 @@ import jetbrains.buildServer.configs.kotlin.triggers.schedule
 import Delft3D.template.*
 import Delft3D.step.*
 
-object WindowsBuildEnvironmentI24 : BuildType({
+object WindowsBuildEnvironmentI25 : BuildType({
```
rene-deltares, line 11: Is it possible to make this name generic: just `WindowsBuildEnvironment`. We have a parameter for the `container.tag` so ideally we would be able to use this configuration to build both the oneapi 2024 and the oneapi 2025 images. But there are other minor differences, like the path to the `toolchain` directory on the P-drive, the path to the Dockerfile. But it seems that's about the extent of it.

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i25
```diff
@@ -0,0 +1,54 @@
...
+RUN Start-Process python-3.12.7-amd64.exe -ArgumentList '/quiet InstallAllUsers=1 PrependPath=1' -Wait
+RUN Remove-Item C:\\python-3.12.7-amd64.exe
+RUN New-Item -ItemType SymbolicLink -Path 'C:\\Program Files\\Python312\\python3.exe' -Target 'C:\\Program Files\\Python312\\python.exe'
+
+ADD Git-2.47.0-64-bit.exe C:\\Git-2.47.0-64-bit.exe
+RUN Start-Process -FilePath "C:\\Git-2.47.0-64-bit.exe" -ArgumentList "/SILENT" -Wait
+RUN Remove-Item C:\\Git-2.47.0-64-bit.exe
+RUN setx /M PATH $($Env:PATH + ';C:\\Program Files\\Git\\usr\\bin')
```
rene-deltares, line 54: The other Dockerfile has this section: 

```
# Copy the setup script into the container
COPY set-env-vs2022.cmd C:\\set-env-vs2022.cmd

# Set the default command to run the setup script
CMD ["cmd", "/S", "/C", "C:\\set-env-vs2022.cmd"]
```

I think this `set-env-vs2022.cmd` file might not work for the `oneapi 2025` build container? Is that why you left it out? In any case, to avoid having to change the `build.kt` build steps in the future. I would just write slightly different `set-env-vs2022.cmd` scripts. One for the oneapi-2024 container, and one for the new oneapi-2025 container. And then just add that script to this container like this:
```
# Copy the setup script into the container
COPY set-env-vs2022-oneapi2025 C:\\set-env.cmd

# Set the default command to run the setup script
CMD ["cmd", "/S", "/C", "C:\\set-env.cmd"]
```

...Or something like that.

src/engines_gpl/wave/packages/wave/wave_exe.cmake
```diff
@@ -2,7 +2,7 @@
 set(executable_files ${src_path}/wave_exe.f90) # Because the .dll and the .exe are defined in the same directory, retrieve the relevant files for the executable alone
 
 # Define icon
-set(icon_file resource/wl.ico)
+set(icon_file resource/wave_icon.rc)
```
rene-deltares, line 5: That's weird... Does d-hydro also build with oneapi-2024 with this change? Or does the build then stop working? If the build still works in oneapi-2024 with this change then we might as well put that in a separate PR already.

rene-deltares (changes requested):

PR #913, opened 2026-05-27
UNST-9958: Add windows oldext config to the dimr_testbench_table

rene-deltares:
# What was done 

Added the oldext config to the testbench table. First of all to see how many of those testcases fail
 
UPDATE: 
The oldext config that I updated is basically an old copy of the `dimr_dflowfm_all_but_validation_cases.xml`
The test cases in `dimr_dflowfm_all_but_validation_cases` and `dimr_dflowfm_oldext_cases` share the same test
case data in MinIO, but `dimr_dflowfm_oldext_cases` has much older timestamps (from 2024, mostly). So these test cases are oooold. We somehow need to make them pass with the latest binaries.

If any change is required to the test case input or the references of a test case in `oldext_cases`, we can't simply update the data in MinIO. People who have to update the `all_but_validation_cases` will get conflicts, and this will get out of hand quickly. So I decided on a "copy-on-write" policy. If I cannot avoid updating the test case data or references, I will make a copy of the test case in MinIO, and add `_oldext` somewhere to the test case name and path. That way, the test case data is kept separately, and there will be no conflicts with the `all_but_validation_cases` config.

I've done the following things to the oldext config to make the test cases pass:

- Removed test cases that were already using the new external forcings format, or where there was no external forcings file at all (about a 100 of them were removed)
- Copied about 10 crashing/faililng test cases to a new location in MinIO
- Severely loosened up the tolerances for the test cases that did not crash, but had `NOK` comparison results.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


FlorisBuwaldaDeltares (approved):
checked out the branch and ran the testbench with the xml. locally the cases look good, thanks for doing this!

FlorisBuwaldaDeltares (approved):

PR #922, opened 2026-05-28
UNST-9644: wave forces are interpolated incorrectly on SWAN MPI domain edges

jreyns:
# What was done 

- Fix mpi domain halo for forces
- when neighbor cells are marked as exceptions, do one-sided inter/extrapolation to retrieve force value

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
TODO: e04_f11_c01
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9644


src/third_party_open/swan/src/swanout1.for
```diff
@@ -408,8 +408,22 @@
...
+                CALL SWEXCHG( COMPDA(1,JHS ), KGRPNT )
+                ALLOCATE(AC2LOC(MCGRD))
+                DO ID = 1, MDC
+                   DO IS = 1, MSC
+                      AC2LOC(:) = AC2(ID,IS,:)
+                      CALL SWEXCHG( AC2LOC, KGRPNT )
+                      AC2(ID,IS,:) = AC2LOC(:)
+                   END DO
+                END DO
+                DEALLOCATE(AC2LOC)
```
hrajagers: No question about the improvement that you observed given this change. Just surprised that it wouldn't have been caught before by any other SWAN user. I see an explicit synchronization of `AC2` and `COMPDA` in `SWCOMP` (in `swancom1.f*`). Why aren't those enough ... are we doing something wrong there?

Furthermore, there are two versions of `SWEXCHG` ... we use the default wavefront approach `!WFR`, but in SWAN there is also a block Jacobi `!JAC` option. For the latter implementation `SWEXCHG`  needs a third argument (inserted as second argument). This may not be relevant for the Delft3D version, but if this bugfix is transferred to https://github.com/Deltares/SWAN, we should probably take this into account.

jreyns: Good point. I started with testing whether the exchange would improve the results and it did partially. The SWOEXF did the rest. I did not revert the exchange after that, assuming it was still needed. I reverted that now, and the result is still okay; so we have no problem between the different SWEXCHG approaches.

marcio-albernaz:
@hrajagers ,
Johan added the root cause and workaround more clearer in the code. I am copying it below as it may help your review. I am running my tests again with the artifacts from this PR. I will get back with results. 

In parallel runs, output points on a computational-grid node may
!       lie on a subdomain cut.  The owner rank writes the point, but the
!       central stencil can include a non-owned/missing neighbour.  Use a
!       one-sided stencil only when the point itself and the opposite
!       neighbour are valid; otherwise keep the original exception path.

marcio-albernaz:
I have tests now with D-Waves standalone. I can reproduce the problem with the latest weekly as well (2.31.10).
Below you see wave forces from the weekly (left) and from the branch (right). The test had 4 MPI partitions, so you can see 3 stripes on the left. Plot is based on the .dat file and uses patch to avoid interpolation issues.

image

src/third_party_open/swan/src/swanout1.for
```diff
@@ -5592,6 +5591,39 @@
...
+            ENDIF
+            IF (.NOT.OK7 .AND. OK8) THEN
+              JYLO = JY
+              IND7 = IND9
+              RRDJ = 1.
+            ELSE IF (.NOT.OK8 .AND. OK7) THEN
+              JYUP = JY
+              IND8 = IND9
+              RRDJ = 1.
+            ENDIF
```
hrajagers, line 5624: Here we do something if `.NOT. OK5`, `.NOT. OK6`, `.NOT. OK7`, `.NOT. OK8`.

src/third_party_open/swan/src/swanout1.for
```diff
@@ -5592,6 +5591,39 @@
...
+          ENDIF
+        ENDIF
         IF (ONY) THEN                                                     40.00
           IF (DEP2(IND5).LE.DEPMIN .OR. .NOT. HS(IND5).NE.0.) GOTO 700
           IF (DEP2(IND6).LE.DEPMIN .OR. .NOT. HS(IND6).NE.0.) GOTO 700
```
hrajagers, line 5629: Here we jump to `700` past most of the work in this loop if `.NOT. OK5`, `.NOT. OK6`, etc. (`ONY` and `ONX` are both true based on previous check, so we're doing the if-clauses.

So, where do we actually do the one-sided discretization based on the code changes above? I've seen the results, but I'm still confused as to how this is supposed to work based on the code changes committed.

hrajagers (commented):
Probably okay, but I don't understand why/how this is working. Maybe one comment by you helps me to understand how this works and then I'll be glad to approve.

src/third_party_open/swan/src/swanout1.for
jreyns: The logic is as follows, correct me if I made a thinking error:

                 IND8 = (JX,   JYUP)
                         north

IND5 = (JXLO,JY)            IND9 = (JX,JY)      IND6 = (JXUP,JY)
             west                       center                             east

                 IND7 = (JX,   JYLO)
                         south
                         
The gradients are calculated using central differences, eg 0.5*(IND6-IND5). If a point on the side is not available (eg IND5), we substitute with the centre value by changing the index (IND5=IND9), and replace the weight RRDI by 1.0 instead of 0.5, as we divide over one cell instead of two.

This enters the calculation as follows:
DDI = RRDI * (DEP2(IND6)-DEP2(IND5))
DXI = RRDI * (XCGRID(JXUP,JY)-XCGRID(JXLO,JY))
DYI = RRDI * (YCGRID(JXUP,JY)-YCGRID(JXLO,JY))
ACWI = RRDI * (AC2(ID,IS,IND6) - AC2(ID,IS,IND5))

effectively making it one sided.

src/third_party_open/swan/src/swanout1.for
jreyns: if the indexes have been replaced above, this checks on the validity of the center value IND9. If it is valid, nothing happens, else you have an exception

src/third_party_open/swan/src/swanout1.for
hrajagers: Ah, yes, of course. You update the indices `IND5`, `IND6`, `IND7`, `IND8` (as well as `JX/YUP/LO` and `RRDI/J`) and hence the check below is **not** the same as the check here. Clear. Now I follow.

hrajagers (approved):
Still surprising to see that this needs a fix on SWAN side after so many years, but I agree with the implemented logic.

PR #929, opened 2026-06-01
UNST-9968 Create framework for pushing conan packages to nexus.

harmenwierenga:
- Created user-facing build.py and run_conan.py that handle basic workflows for internal and external developers.
- Created teamcity configurations for building the packages and pushing them to nexus when not available there.
- Provide a method for configuring conan across developer machines

# Issue link UNST-9968


run_conan.py
```diff
@@ -0,0 +1,379 @@
...
+    MISSING = "missing"  # Build packages from source if a pre-built binary is not available from remotes.
+    ALL = "all"  # Build all packages from source using local recipes only, do not use any pre-built binaries from remotes.
+
+
+def clean_conan_cache() -> None:
+    subprocess.run(["conan", "remove", "*", "--confirm"], check=True)
+    subprocess.run(["conan", "cache", "clean"], check=True)
+
+
+def update_lockfile(profile: str) -> None:
```
harmenwierenga, line 114: update-lockfile still needs changing, it does not work at the moment.

conan/config/profiles/delft3d_alma8_intel_2024
harmenwierenga, line 1: I doubt about the names of the profiles. My considerations:

1. We already let many developers set up their machines. Changing the name now would require them to run the initialize again.
2. Using different profile names also means that the current error in build.py can be used again to trigger the running of run_conan.py initialize, since it triggers based on whether a profile with a certain name is present. Adding versioning to the names of the profiles may be very useful for that in general
3. If we decide that we want versioning in the names, what would be useful? I first thought delft3d_linux_intel_2024_2 and delft3d_windows_vs_2022. This reflects the compilers (understandable names), but it does not make it easier to change the profiles unless there is a compiler update. The question is if we need profile updates often without a compiler update
4. We could also add a version that is only meaningful in the context of the profiles: delft3d_linux_1_0, update the minor if we change the profile for whatever reason, and update the major when we update compilers. Major version profiles can live side-by-side, minor ones replace the old ones.

What do you think?

build.py
```diff
@@ -0,0 +1,266 @@
...
+        choices=["Debug", "Release", "RelWithDebInfo"],
+        help="CMake build type (default: Debug).",
+    )
+    parser.add_argument(
+        "--vs",
+        choices=list(VS_GENERATORS.keys()),
+        help="Visual Studio version to use. Auto-detected if not specified.",
+    )
+    parser.add_argument(
+        "--keep-build",
```
rene-deltares, line 193: I would expect that 'build-build' is default behavior, and that there would be an option to clean the build directories instead. I think throwing the build directory every time is a bit wasteful

build.py
```diff
@@ -0,0 +1,266 @@
...
+        args.config,
+        vs_year=vs_year,
+        build_type=args.build_type,
+        build_dir=build_dir,
+        install_dir=install_dir,
+    )
+
+    # Build and install
+    if args.build:
+        run_cmake_build(args.config, build_type=args.build_type, build_dir=build_dir)
```
rene-deltares, line 260: Isn't this just equivalent to `cmake --build --target install`? Do we need separate `run_cmake_build` and `run_cmake_install` functions?

build.py
harmenwierenga: As discussed in person, this is a historical convention coming from build.bat, which did not even have the option to keep the build folder until a change two years ago. Most people use the IDE to incrementally build and expect the build script to do a clean build

build.py
harmenwierenga: I personally like the distinction, I find it weird that our builds come with an install at the moment. Running the software in a build environment should not require an install I think. The unit tests already run after only --build. So this is more to prepare for a future where we move to install only being necessary for creating the artifacts for other projects (and users) to use.

run_conan.py
harmenwierenga: I changed it.

build.py
KoenBussemakerDeltares: I agree with Rene that it's ugly, but let's not change the default behavior. I expect more than enough pushback on the changes as they are today :P

build.py
KoenBussemakerDeltares: +1 for separating build and install

.gitignore
```diff
@@ -66,3 +66,6 @@ test/deltares_testbench/data/cases/*
 # precice
 examples/dflowfm/**/precice_debug_output.txt
 examples/dflowfm/**/precice-run/**
+
+# conan
+conan/**/test_package/**
```
rene-deltares, line 71: Does using conan create these `test_package` directories?

build.py
```diff
@@ -0,0 +1,266 @@
...
+    if platform.system() == "Windows":
+        return f"build_{config}"
+    return f"build_{config}_{build_type.lower()}"
+
+
+def install_dir_name(config: str, build_type: str) -> str:
+    """Install directory name: on Linux includes build type, on Windows does not."""
+    if platform.system() == "Windows":
+        return f"install_{config}"
+    return f"install_{config}_{build_type.lower()}"
```
rene-deltares: This is different then what I'm used to. For some reason, on Linux we have used `build_{config}_{build_type}/install` as the install folder. Do we want to deviate from that now? That means I have to change some devcontainer stuff, otherwise the testbench can't find the binaries and the cmakepresets example doesn't work anymore. Amongst maybe other stuff

.gitignore
harmenwierenga: Yes, certain commands run the package tests and add those in the recipes folder

build.py
harmenwierenga: I changed it to match the devcontainer

conan/config/profiles/delft3d_alma8_intel_2024
rene-deltares: We're deciding on whether to create multiple configs with different names and reusing one config and updating the content, right?

If we keep one config and update its content in main, then people doing a build.py will get different package hashes then the ones in Nexus, no? So they'll get something like: "Package not found" errors. If we instead add a new config in main, with a new name, then I guess you would immediately see: "Profile not found", which directly points to the correct solution: The new config needs to be installed. To me that seems much better.

If we go with multiple config files, I would prefer the descriptive names over the major/minor scheme. I think mostly the compiler updates will be the most likely source of config updates. And if there's a minor update, we can also add those to the name, you know, like `delft3d_linux_intel_2024_2_with_c++23` or something like that.

build.py
rene-deltares: Thanks, indeed, I get the compatibilty argument. I will be using `--keep-build` in my vscode task I think

build.py
```diff
@@ -0,0 +1,271 @@
...
+    parser.add_argument(
+        "--build",
+        action="store_true",
+        help="Run build and install steps after CMake configure.",
+    )
+    parser.add_argument(
+        "--build-type",
+        default="Debug",
+        choices=["Debug", "Release", "RelWithDebInfo"],
+        help="CMake build type (default: Debug).",
```
rene-deltares, line 185: We could add...

```python
class BuildType(StrEnum):
    DEBUG = "Debug"
    RELEASE = "Release"
    RELWITHDEBINFO = "RelWithDebInfo"
```

...and add `type=BuildType` to this `add_argument`. But, maybe not absolutely necessary. The benefit would be that it prevents typing error by causing typing errors when making changes in this script. You could say the same thing with VS_GENERATORS and CONFIGURATIONS I guess.

build.sh
rene-deltares, line 1: Good riddance!

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+import platform
+import shutil
+import subprocess
+import sys
+import tempfile
+from collections.abc import Generator
+from contextlib import contextmanager
+from pathlib import Path
+
+CONFIG_DIR = Path("conan/config")
```
rene-deltares: Perhaps adding a `ROOT = Path(__file__).resolve().parent` in this script, and defining these three files like this...
```python
CONFIG_DIR = ROOT / "conan/config"
RECIPES_DIR = ROOT / "conan/recipes"
LOCKFILE = ROOT / "conan.lock"
```
... would be nice to prevent errors when people run `python ../../run_conan.py`, or something.

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+        help=(
+            "CMake build type for the consumer. "
+            "On Linux, determines which CMakeDeps files are generated. "
+            "Ignored on Windows (all configurations are always generated)."
+        ),
+    )
+    parser_install.add_argument(
+        "--output-folder",
+        default="build/conan",
+        help="Output folder for Conan install files.",
```
rene-deltares, line 445: If stel voor om hier een `type=Path` bij te zetten. In `build.py` gebruiken we ook `pathlib`

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+        choices=["Release", "Debug", "RelWithDebInfo"],
+        help=(
+            "CMake build type for the consumer. "
+            "On Linux, determines which CMakeDeps files are generated. "
+            "Ignored on Windows (all configurations are always generated)."
+        ),
+    )
+    parser_install.add_argument(
+        "--output-folder",
+        default="build/conan",
```
rene-deltares: Deze default waarde is eigenlijk nooit goed. Kunnen we deze niet required maken?

build.py
rene-deltares: Bij nader inzien, doe eigenlijk maar niet. Dan zouden we het ook in `run_conan.py` moeten doen, en dan krijg je code duplicatie, en dan krijg je twee 'losse' scripts die eigenlijk dingen moeten delen, en dan moet je er eigenlijk al een python package van maken om het netjes op te lossen.
Niet dat ik het een slecht idee (van mezelf) vind, maar... kan ook later

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+    ALL = "all"  # Build all packages from source using local recipes only, do not use any pre-built binaries from remotes.
+
+
+def clean_conan_cache() -> None:
+    subprocess.run(["conan", "remove", "*", "--confirm"], check=True)
+    subprocess.run(["conan", "cache", "clean"], check=True)
+
+
+@contextmanager
+def _isolated_conan_home() -> Generator[Path, None, None]:
```
rene-deltares, line 95: Leuk trucje, hè?!

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+def conan_install(
+    profile: str,
+    output_folder: str,
+    build_type: str,
+    *,
+    consumer_build_type: str | None = None,
+    ci: bool = False,
+    lockfile: Path | None = None,
```
rene-deltares: `lockfile` mag nooit `None` zijn vind ik. Dat is een duidelijk signaal dat er iets mis is, en dan heb ik liever dat er zo vroeg mogelijk een error wordt gegooid in plaats van dat er toch 'iets' gebeurt. In geen enkele aanroep van `conan_install` is deze `None` volgens mij, dus ik zou er gewoon keihard een `Path` van maken.

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+def _iter_packages(data: dict) -> Generator[tuple[str, str, str], None, None]:
+    return (
+        (ref, rrev, pkg_id)
+        for ref, ref_data in data.items()
+        for rrev, rrev_data in ref_data.get("revisions", {}).items()
+        for pkg_id in rrev_data.get("packages", {})
+    )
+
+
+def upload_new_packages(remote: str, *, ci: bool = False) -> None:
```
rene-deltares: Dit vind ik eigenlijk de meest problematische code in deze file. Ik snap wat het doet, maar er is een sterke afhankelijkheid van het formaat json wat die commando's terug geven. We gaan ervan uit dat het dicts teruggeeft, halen daar keys uit, en als ze niet bestaan gebruiken zeggen we... oh, dan geven we een lege `dict` terug (bijvoorbeeld `ref_data.get("revisions", {})`). Ik vind dat als dat "revisions" niet bestaat, dat er dan een error gegooid moet worden. Anders lijkt deze code gewoon te werken, maar doet het in feite het verkeerde

run_conan.py
```diff
@@ -0,0 +1,424 @@
...
+            subprocess.run(cmd, check=True)
+            uploaded += 1
+
+    print(f"\nDone. Uploaded: {uploaded}, skipped: {skipped}")
+
+
+def _get_profile() -> str:
+    os_name = platform.system()
+    if os_name == "Windows":
+        return "delft3d_windows"
```
rene-deltares: Misschien willen we in plaats van een string literal hier een constante van maken en die bovenaan de file zetten. Als we dan het 'default' profiel willen aanpassen staat die verandering bovenaan.

conan/config/profiles/delft3d_alma8_intel_2024
harmenwierenga: I have renamed the profiles to reflect the OS (distro on linux) and the main compiler version (intel 2024 on linux, msvc 194 on windows)

rene-deltares (approved):
We can work with this!

PR #931, opened 2026-06-01
UNST-9957 don't write averaged dmiss

FlorisBuwaldaDeltares:
… statistical output

# What was done 

- for averaging output, if at any point during the output interval a missing value is encountered, the entire output interval is skipped and dmiss is written
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_statistical_output.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_statistical_output.f90
@@ -141,18 +141,36 @@ contains
    !> Update the stat_output of an item, depending on the operation_type.
    elemental subroutine update_statistical_output(item, dts)
       use precision, only: dp
+      use m_missing, only: dmiss
+      use precision_basics, only: equal
       type(t_output_variable_item), intent(inout) :: item !< statistical output item to update
       real(kind=dp), intent(in) :: dts !< current timestep
+      logical, allocatable :: mask(:)
+      real(kind=dp), pointer :: source(:)
+
+      if (item%operation_type == SO_CURRENT) then
+         return
+      end if
+
+      ! Check for dmiss values, they may not be included in any averaging
+      mask = equal(item%source_input, dmiss)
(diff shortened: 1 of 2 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_statistical_output.f90
```diff
@@ -141,18 +141,36 @@ end subroutine update_source_input
...
+         return
+      end if
+
+      ! Check for dmiss values, they may not be included in any averaging
+      mask = equal(item%source_input, dmiss)
+      if (any(mask)) then ! we cannot modify input, create a copy with dmiss values set to 0.
+         allocate(source(size(item%source_input)))
+         source = item%source_input
+         where (mask)
+            source = 0.0_dp
```
hrajagers: Stepping back for a moment: what are we doing here? Does this mean that in a simulation with (fairly) constant water temperature, the "average temperature" output for cells that are dry part of the time, will be (significantly) lower than the same quantity for all other cells. Similar for "average water level", etc.? We're taking zero as the value during dry/masked periods. An alternative would be to average over wet periods only (that makes the averaging period spatially varying) or exclude all cells that are `dmiss` during part of the period?

I agree that `dmiss` shouldn't be used in the average, but `0` might not always give the expected alternative.

FlorisBuwaldaDeltares: Hi @hrajagers, the question is what the user expects and wants. The average temperature is strange if dry means 0, but average water level and discharge etc are natural.

Current proposal by Julien is to exclude writing average values if in any part of the timesteps in the his-interval a value dmiss was encountered in that point. averaging period spatially varying is a headache, so i'd prefer to exclude that but i'm looking into what would be the least-ugly option

thomas-pijls (approved):
Great work!

PR #940, opened 2026-06-03
fix: add usedforsecurity=False to all md5 uses

aslowwriter:
# What was done 
 
The test benches use md5 hashes in a couple of places without marking them as not being used for security which is not allowed in FIPS environments so when those were running on FIPS agents it would break. This PR adds the `usedforsecruity=False` to all remaining uses of MD5 hashes in the test benchess to fix that. The tests already ran successfully on normal agents, and I'm running them now on the FIPS agent as well to be shure. 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


aslowwriter:
should run in all all branch. closing this one and reopening with the correct branch name

PR #942, opened 2026-06-04
Function to convert NF2FF source sinks to FF ready source sinks

AabhasSrivastava:
# What was done 

- added some documentation
- added convertNFSinksToFF(..) function with unit test, based on my current understanding, to be verified
- the function is not currently called by the pipeline
- Tests need to be updated/expanded
- functionality can possibly be separated in originally envisioned 3 blocks (entrainment, discharge, intake to Source/Sink) 
- decide what to do if sum of source weights is 0
- Question - do we need to link intake and discharge volume, or FM handles it, as long as the quantity is resent in created endpoint structure?
- UNST-9807 should flatten endpoints structure for preCICE write

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [x]	Documentation updated \
- [ ]	Not applicable 

# Issue link
UNST-9808


src/tools_gpl/pre_c_sumo/include/endpoints.hpp
```diff
@@ -75,6 +82,25 @@ namespace pre_c_sumo
         std::optional<Constituents> constituents; ///< Optional constituent concentrations carried by this source.
```
adrimourits, line 82: TODO: Optional momentum magnitude and direction

src/tools_gpl/pre_c_sumo/include/NF2FF_reader.hpp
```diff
@@ -5,6 +5,7 @@
 #include "parsing_types.hpp"         // For parsing_utils::Point2D
```
adrimourits, line 6: Big chance that your changes will conflict with [UNST-9806](https://issuetracker.deltares.nl/browse/UNST-9806)

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -202,6 +266,106 @@ namespace pre_c_sumo
 
     void convertNFSinksToFF() { std::println("Processing sinks..."); }
 
+    // This function fills up the endpoints data structure, using entrainment, discharge and intakes
+    // Combined it in function for first try, can maybe be seprated into different functions like in FM.
+    // The endpoint structure needs to be converted
```
adrimourits, line 271: Sounds over-complicated to me. If additional conversion is needed, what is the benefit of the endpoint structure?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
erwin-mulder, line 233: Needs at least an `std::vector<pre_c_sumo::NF2FFReader> nf2ff_readers` parameter.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -171,9 +235,9 @@ namespace pre_c_sumo
         for (const auto& diffuser : csumo_settings.diffusers())
         {
             std::println("Converting NF data to sources/sinks for diffuser {} ...", diffuser.nf2ff_file.value());
-            convertNFSinksToFF();
-            convertNFIntakesToFF();
-            convertNFSourcesToFF();
+            convertNFSinksToFF(); //needs to reference to new function with arguments
+            convertNFIntakesToFF(); // new function already takes care of this, can be removed.
+            convertNFSourcesToFF();// new function already takes care of this, can be removed.
```
erwin-mulder, line 240: Why not remove them?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -202,6 +266,106 @@ namespace pre_c_sumo
 
     void convertNFSinksToFF() { std::println("Processing sinks..."); }
 
+    // This function fills up the endpoints data structure, using entrainment, discharge and intakes
+    // Combined it in function for first try, can maybe be seprated into different functions like in FM.
+    // The endpoint structure needs to be converted
+    // to flat preCICE array in sendSourcesSinksToFF function
+    double convertNFSinksToFF(const NF2FFReader& nf2ff_reader, SourcesSinks& sources_sinks, const double first_record_id,
```
erwin-mulder, line 273: Consider renaming this function if Intake and Source processing has been included already.

erwin-mulder (changes requested):

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -1,19 +1,83 @@
...
+#include "endpoints.hpp"
 #include "pre_c_sumo_lib.hpp"
 #include "FF2NF_writer.hpp"
 #include "parsing_types.hpp"
 
 namespace pre_c_sumo
 {
+    namespace
+    {
+        constexpr double zero_connected_id = 0.0;
```
adrimourits, line 22: There is a lot of messing with id's (double to int). Is it possible to keep it int everywhere, and only convert it to double when it is written into the preCICE arrays?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -1,19 +1,83 @@
...
+
+        int asEndpointId(const double id)
+        {
+            return static_cast<int>(std::lround(id));
+        }
+
+        void appendEndpointRecord(SourcesSinks& sources_sinks, const Endpoint& endpoint)
```
adrimourits, line 29: Confusing names: `SourcesSinks`, `Endpoint`, `SourceOrSinkData`, `Source.` Can they be improved (or at least clarified)?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -1,19 +1,83 @@
...
+            }
+            const auto additional_count = std::min(values.size(), std::size_t{2} + constituent_count);
+            for (std::size_t i = 2; i < additional_count; ++i)
+            {
+                constituents.additional_constituents[i - 2] = values[i];
+            }
+            addConstituents(source, constituents);
+        }
+
+        std::pair<double, double> sourceDepthBounds(const SourceOrSinkData& source, const std::size_t source_count)
```
adrimourits, line 71: I prefer `z` or `z-coordinate` over `Depth`. Might be personal.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -202,6 +266,106 @@ namespace pre_c_sumo
...
+    // to flat preCICE array in sendSourcesSinksToFF function
+    double convertNFSinksToFF(const NF2FFReader& nf2ff_reader, SourcesSinks& sources_sinks, const double first_record_id,
+                              const std::optional<parsing_utils::Point2D>& intake_point)
+    {
+        auto sources = nf2ff_reader.sources();
+        auto sinks = nf2ff_reader.sinks();
+
+        if (sources.empty())
+        {
+            return first_record_id;
```
adrimourits, line 281: What is the idea of returning an id?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -202,6 +266,106 @@ namespace pre_c_sumo
...
+            const auto& previous_sink = sinks[sink_index - 1];
+            const auto& sink = sinks[sink_index];
+            const double delta_s = sink.entrainment - previous_sink.entrainment;
+
+            for (std::size_t source_index = 0; source_index < sources.size(); ++source_index)
+            {
+                const auto& source = sources[source_index];
+                const double discharge = delta_s * source_flow_rate * source_weights[source_index];
+
+                const double sink_id = next_id++;
```
adrimourits, line 320: Why are `sink_id` and `source_id` based on `first_record_id`, which is an index related to the NF2FF-file read? Might cause conflicts.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -202,6 +266,106 @@ namespace pre_c_sumo
...
+                const double source_id = next_id++;
+                const auto [source_z_min, source_z_max] = sourceDepthBounds(source, sources.size());
+
+                const auto sink_endpoint = makeEndpoint(asEndpointId(sink_id), asEndpointId(source_id), sink.x_coordinate,
+                                                        sink.y_coordinate, -sink.z_coordinate - sink.half_plume_height,
```
adrimourits, line 325: Why are `sink_z_min` and `sink_z_max` computed on-the-fly and `source_z_min` and `source_z_max` via subroutine `sourceDepthBounds`?

src/tools_gpl/pre_c_sumo/include/pre_c_sumo_lib.hpp
```diff
@@ -70,15 +70,22 @@ namespace pre_c_sumo
...
     };
 
+    /**
+     * @brief preCICE-ready source/sink payload prepared by pre-C-SUMO.
+     *
+     * The container keeps the raw geometric and discharge data separate from
+     * the FM cell mapping. The FM side is responsible for converting x/y
+     * coordinates to cell ids and for applying nearfield::dischargeToSrc.
+     */
     struct SourcesSinks
```
adrimourits, line 80: Discussion needed (out of scope for this sprint?):
Sources have more fields than sinks (momentum, constituents). What are we going to feed to preCICE:

1. Separated arrays for sources and sinks
2. All sources and sinks mixed together, zeroes (or NaN/-huge/...) when n/a

adrimourits (commented):
Note the dependency on / overlap with [UNST-9806](https://issuetracker.deltares.nl/browse/UNST-9806).
I propose that code related to momentum/constituents has low priority.
I wonder what the added value is of the Endpoint struct. May be a discussion is needed.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: We sould revisit if that data structure is actually needed or not. Right now, I wanted to use that structure. i think we need to figure out what would be the bet way to handle this data when converting to flat precice arrays.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: yeah, it needs to be changed to int, first, and we need to see how we handle the connectivity part.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: This is another point I still want to tune. THanks for noting it down here since I forgot to mention it in the description. ties with the data structure endpoint and whatever structure we need for data to be sent to precice. I would prefer minmum data structures. If things can be reused between file reading structure and this conversion, would prefer that.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: WHat is the netCDF convention? Maybe we can use that?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: for multiple diffusers, it keeps track of having ids traceable, or making sure that there is no conflict. Again a work in progress.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
AabhasSrivastava: NO, it is not indexed from the file. We would start with 1, and then start keeping track of the ids as they get filled up per diffuser NF2FF file.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
adrimourits: Depth is something different than z-coordinate. I think here it is z-coordinate. The NetCDF convention that comes closest seems to be `ocean_sigma_z_coordinate` but using it here does not seem to be appropriate.

AabhasSrivastava:
implemented in PR#1021

PR #946, opened 2026-06-05
chore: format all python files with ruff

aslowwriter:
# What was done 

- no code changes were made in this pr, except for using ruff to format all python files in the repo. These are safe changes, so no behaviour should have changed. This will make sure other PRs don't have to fix these formatting issues when they touch a file
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


conan/recipes/cmake/3.x.x/test_v1_package/conanfile.py
rene-deltares, line 1: I would revert all changes to the python files in the `conan/recipes` directory. Any tiny change in these `conanfile.py` files will change the 'revision identifier' of the conan packages. And this will cause 'conan install' to look for a package in Nexus that doesn't exist there.
While it is technically possible to commit these changes, rebuild the packages, upload the packages to nexus and then merge them to main, I would refrain. Some of these recipes we don't make ourselves, but we copy from Conan's own github repo. Making our own formatting changes to these files will make it more difficult to merge changes from Conan's github repo back to our own.

src/third_party_open/fortrangis/include/mkintcproto.py
rene-deltares, line 1: I would revert all changes to python files in the `third_party_open` directory. This directory includes wholesale copies of some third-party-code. We will remove these files anyway rather soon actually. We're getting rid of `fortrangis` and `googletest` in favor of conan packages.

ci/teamcity/Delft3D/windows/scripts/determineProduct.py
```diff
@@ -12,7 +12,9 @@
     else:
         effective_branch = branch
 
-    if effective_branch.startswith("revert-") or effective_branch.startswith("dependabot/"):
```
rene-deltares, line 15: Hey, what --line-length are you using? I think the default is 88 for some reason. But in the python code that we do format consistently (test/deltares_testbench and ci/python) we use 120. In my experience 88 is rather short, so let's keep it consistent to 120 for the other code as well. I like to use named arguments in function calls. help= strings in argparse and documentation strings also go over the limit rather easily. Also, qualified import statements.

If I skim through the changes in this PR, I think setting it to 120 will also get rid of many small changes. But I may be wrong.

tools/clang_format/clang_format_all.py
```diff
@@ -11,16 +11,19 @@
 # TODO: Remove filtering logic when third party code is no longer in <repo_root>/src
 IGNORE_DIRS = {"third_party", "third_party_open", "thirdParty"}
 
+
 def should_format(path: Path) -> bool:
```
rene-deltares, line 15: I appreciate the irony of having an unformatted python script to format all the C++ code

rene-deltares (changes requested):
I did request some changes. I hope that soon we're going to find a good solution to these few loose python scripts in our repo, so we can do these kind of things automatically like we do in testbench and ci-python

conan/recipes/cmake/3.x.x/test_v1_package/conanfile.py
aslowwriter: gotcha, good point, reverted them :)

ci/teamcity/Delft3D/windows/scripts/determineProduct.py
aslowwriter: I originally did it just using the default but I've recreated the pr using `--line-length 120` so it should be good now.

conanfile.py
rene-deltares, line 1: Sorry, deze had ik over het hoofd gezien! Ik ben bang dat deze iets met de conan packages zal veranderen. Dus zou je deze ook kunnen reverten. De rest ziet er goed uit!

rene-deltares (changes requested):
Sorry, one more change :( looks good otherwise

rene-deltares (approved):

PR #947, opened 2026-06-05
All/task/unst 9902 initial conan setup

harmenwierenga:
# What was done 

- Replace NetCDF, NetCDF-Fortran, GDAL, PROJ, Expat, and related libraries with Conan packages
- On windows, install dlls in the bin folder
- Update documentation with new build steps
- Update teamcity steps to make use of new build scripts

# Issue link UNST-9902


KoenBussemakerDeltares (approved):
This is a significant approvement in terms of maintainability and eventually software security. It is also a large and fundamental change. We have been very careful and have tested a lot, but with a change this impactful there are bound to be issues. We made it clear to the rest of the department that we are available for help and to fix any issues that arise. IMO there is no reason to wait any longer, let's do this!

rene-deltares (approved):
This allows us to finally do package management properly

PR #949, opened 2026-06-08
UNST-9974: Merge "prepare_lateral_mask" and "construct_target_mask"

thomas-pijls:
# What was done 

- Created a new file, `flowgeom_mask.f90` containing the `construct_mask` subroutine.
- The `construct_mask` subroutine combines the behaviour of the old `prepare_lateral_mask` and `construct_target_mask` subroutines.
- Moved the `parse_location_type` function to `flowgeom_mask.f90`.
- Deleted the now empty `m_lateral_helper_functions.f90` file.
- Removed the now no longer in use `construct_target_mask` subroutine.
- Extended the behaviour of the old `prepare_lateral_mask` subroutine to also work for `UNC_LOC_CN` and `UNC_LOC_U`

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+   use precision_basics, only: dp
+   use fm_location_types, only: UNC_LOC_S, UNC_LOC_S3D, UNC_LOC_U
+   use m_laterals, only: ILATTP_1D, ILATTP_2D, ILATTP_ALL, ILATTP_INVALID
+
+   implicit none(type, external)
+
+   private
+
+   public :: construct_mask
+   public :: parse_location_type
```
mihai-pricope: Can we have parse_location_type private and have two versions (or one with optionals) of construct_mask. One with ilattype as integer and one as string?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_construct_mask.f90
```diff
@@ -0,0 +1,23 @@
...
+   use assertions_gtest
+
+   implicit none(type, external)
+
+contains
+
+   !$f90tw TESTCODE(TEST, test_init_spatial_field, test_parse_location_type, test_parse_location_type,
+   !> parse_location_type must map all recognized strings, empty to INVALID, and unknown to ALL.
+   subroutine test_parse_location_type() bind(C)
+      use m_construct_mask, only: parse_location_type
```
mihai-pricope: This is kinda of a superflous test. I would as well remove it as it does not test much. If you want to write a test, maybe write a construct_mask test. See also above comment about making parse_location_type module private

mihai-pricope (changes requested):
I think the public api can be made simpler, by having only a construct_mask method

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+
+contains
+
+   !> Construct a mask array; either based on the lateral type (1D, 2D, 1D2D) or based on a target mask file (polygon file).
+   subroutine construct_mask(mask, target_location_type, ilattype, target_num_points, target_mask_file, invert_mask, ierr)
```
arthurvd: More intuitive name / docstring:
```suggestion
   !> Construct a spatial mask array for flow geometry; either based on the location type (1D, 2D, 1D2D) or based on a target mask file (polygon file).
   !! Can be called for the various topological target location types (e.g., cells, flow links, etc.)
   subroutine construct_spatial_mask(mask, target_location_type, spatial_location_type, target_num_points, target_mask_file, invert_mask, ierr)
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+      use m_flowgeom, only: lnx1d, ln, lnx, ndx, ndxi, ndx2d, lnxi, prof1d, xz, yz, kcs
+      use timespace_parameters, only: LOCTP_POLYGON_FILE
+      use timespace, only: selectelset_internal_nodes, selectelset_internal_links
+      use dfm_error, only: DFM_NOERR, DFM_WRONGINPUT, DFM_NOTIMPLEMENTED
+
+      ! Parameters
+      integer, dimension(:), allocatable, intent(inout) :: mask !< Mask array for the target element set.
+      integer, intent(in) :: target_location_type !< The location type parameter (one from fm_location_types::UNC_LOC_*) for this quantity's target element set.
+      integer, intent(in) :: ilattype !< Type of the lateral.
```
arthurvd: Now that this subroutine is generic, let's generalize the terminology (it is no longer lateral-specific).
```suggestion
      integer, intent(in) :: spatial_location_type !< Allowed spatial location type (one of ILATTP_ALL/_1D/_2D).
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+      integer, intent(in), optional :: target_num_points !< Number of points in target element set. Will be used to allocate the mask array.
```
arthurvd:
```suggestion
      integer, intent(in), optional :: target_num_points !< Number of points in target element set. Will be used to allocate the mask array. When absent, it wil be derived form the target_location_type.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+!  "D-Flow Flexible Mesh" and "Deltares" are registered trademarks of
+!  Stichting Deltares, and remain the property of Stichting Deltares.
+!  All rights reserved.
+!
+!-------------------------------------------------------------------------------
+
+module m_construct_mask
+   use precision_basics, only: dp
+   use fm_location_types, only: UNC_LOC_S, UNC_LOC_S3D, UNC_LOC_U
+   use m_laterals, only: ILATTP_1D, ILATTP_2D, ILATTP_ALL, ILATTP_INVALID
```
arthurvd: Consider renaming to: `m_flowgeom_mask` and moving ILATTP_1D, etc. to `fm_location_types` without LAT in the name.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
```diff
@@ -0,0 +1,219 @@
...
+            do L = lnx1D + 1, lnxi
+               n1 = ln(1, L)
+               mask(n1) = 1
+               n2 = ln(2, L)
+               mask(n2) = 1
+            end do
+
+         end select
+
+      else if (present(target_num_points) .and. present(target_mask_file) .and. present(invert_mask) .and. present(ierr)) then
```
arthurvd: This else case is indeed equal to the original behavior (mask only based on polygon), but the idea was that the 1d/2d selection ("ilattype") now also becomes available in combination with a polygon and for other target_location_type than S/S3D.

See JIRA issue text:
> TODO: unify both routines into a single one that constructs the target mask regardless of present polygon and for all location types

Note that you could use `selectelset_internal_nodes/links` by passing mask as the third "kcs" argument, with already the 1d/2d premasking done in a first step.

arthurvd (changes requested):
See remarks + discussed in person.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
thomas-pijls: Yes, done!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_construct_mask.f90
thomas-pijls: It is indeed superflous, I removed it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
thomas-pijls: Thanks, updated the docstring

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
thomas-pijls: Updated variable name!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/construct_mask.f90
thomas-pijls: Renamed the module and moved & renamed the ILATTP_*

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+   interface construct_mask
+      module procedure construct_mask_string_spatial_location
+      module procedure construct_mask_integer_spatial_location
+   end interface construct_mask
+
+contains
+
+   !> Construct a spatial mask array for flow geometry; either based on the location type (1D, 2D, 1D2D) or based on a target mask file (polygon file).
+   !! Can be called for the various topological target location types (e.g., cells, flow links, etc.)
+   subroutine construct_mask_string_spatial_location(mask, location_type, spatial_location_type, target_mask_file, invert_mask, ierr)
```
arthurvd: This entire subroutine is not truly needed, it is better to call these utility functions with (integer) enums only and leave the string parsing to the call site. Because that is where I/O and readable warning for the user are happening anyway.

So, I'd remove this string version of the subroutine (and the construct_mask interface), and rename the integer subroutine to `construct_mask`. See my other remark at the call site (inside `init_spatial_fields()`).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -912,7 +912,7 @@ module function init_spatial_fields(block_ptr, base_dir, file_name, group_name)
 
          call get_location_target_properties(target_location_type, target_num_points, target_x, target_y, is_static_field, ierr)
 
-         call construct_mask(mask, target_location_type, parse_location_type(input%location_type), target_num_points, target_mask_file, invert_mask, ierr)
+         call construct_mask(mask, target_location_type, input%location_type, target_mask_file, invert_mask, ierr)
```
arthurvd: Here, the original code already handles the string-to-int parsing on the call site (`..., parse_location_type(input%location_type)`). Consider keeping that original, so that you don't need a string version of your construct_mask.

Additionally, the string-to-int for the spatial location type that used to be done here should actually be part of the subroutine `validate_spatial_field_input()`, because that one is already doing the same for  the operand and interpolationMethod. But maybe you consider that scope creep for this particular PR.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init_old.f90
```diff
@@ -293,8 +293,7 @@ module subroutine init_old(iresult)
                ! NOTE: we intentionally re-use the lateral coding here for selection of 1D and/or 2D flow nodes
                select case (trim(qid(18:)))
                case ('1d', '2d')
-                  ilattype = parse_location_type(qid(18:))
-                  call construct_mask(mask, UNC_LOC_S, ilattype)
+                  call construct_mask(mask, UNC_LOC_S, trim(qid(18:)))
```
arthurvd: Also here, I would restore the original code: input parsing on the init side (also is a better place to print warning messages), and only call construct_mask with integer enum values.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+         spatial_location_type = SPATIAL_LOCATION_1D
+      case ('2d')
+         spatial_location_type = SPATIAL_LOCATION_2D
+      case ('1d2d', 'all')
+         spatial_location_type = SPATIAL_LOCATION_ALL
+      case default
+         spatial_location_type = SPATIAL_LOCATION_ALL
+      end select
+
+   end function parse_spatial_location_type
```
arthurvd: This function fits better inside `fm_location_types`, because that is where the enums are too.
(You may want to use location types elsewhere without any masking, so the `parse_spatial_location_type` should not be specific inside the `m_flowgeom_mask` module.)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
```diff
@@ -662,17 +662,17 @@ subroutine readIniFieldProvider(inifilename, node_ptr, groupname, quantity, file
...
+               ilocType = SPATIAL_LOCATION_1D
             case ('2d')
-               ilocType = ILATTP_2D
+               ilocType = SPATIAL_LOCATION_2D
             case ('1d2d')
-               ilocType = ILATTP_ALL
+               ilocType = SPATIAL_LOCATION_ALL
             case default
-               ilocType = ILATTP_ALL
+               ilocType = SPATIAL_LOCATION_ALL
```
arthurvd: This can be replaced by a call to `parse_spatial_location_type()` (which is preferrably placed in module `fm_location_types`).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -557,13 +558,13 @@ function init_lateral_forcings(block_ptr, base_dir, block_number, major) result(
...
       case ('2d')
-         ilattype = ILATTP_2D
+         ilattype = SPATIAL_LOCATION_2D
       case ('1d2d', 'all')
-         ilattype = ILATTP_ALL
+         ilattype = SPATIAL_LOCATION_ALL
       case default
-         ilattype = ILATTP_ALL
+         ilattype = SPATIAL_LOCATION_ALL
       end select
```
arthurvd: Can be replaced by a call to `parse_spatial_location_type()`.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+      ! Parameters
+      integer, dimension(:), intent(inout) :: mask !< Mask array for the target element set.
+      integer, intent(in) :: location_type !< The location type parameter (one from fm_location_types::UNC_LOC_*).
+      integer, intent(in) :: spatial_location_type !< The spatial location parameter (one from fm_location_types::SPATIAL_LOCATION_*).
+
+      ! Local variables
+      integer :: L !< Loop variable for links.
+      integer :: n1 !< The left flow cell number in the link.
+      integer :: n2 !< The right flow cell number in the link.
+
```
arthurvd, line 107: Consider moving the `mask = 0` from the call site to here. Now this subroutine only *ADDS* new cells to the mask=1, it does not set any elements to 0.
Or, leave the code as is, but properly document it in the docstring of the function ("Any existing values 1 in the input mask are left untouched.").

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+         end select
+
+      case (SPATIAL_LOCATION_ALL) ! Everything in 1D and 2D, except 1D pipes
+
+         select case (location_type)
+         case (UNC_LOC_S, UNC_LOC_S3D)
+
+            do L = 1, lnx1D
+               ! When is lateral allowed?
```
arthurvd, line 163:
```suggestion
               ! The code below originates from 1D/2D selection for laterals. When is a lateral allowed?
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+         end select
+
+      case (SPATIAL_LOCATION_INVALID)
+
+         ! No spatial mask applied, return input mask.
+         continue
+      
+      end select
+
+   end subroutine
```
arthurvd:
```suggestion
   end subroutine apply_spatial_location_mask
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+
+   !> Apply a polygon mask to the provided mask array, based on the provided target mask file and location type (flow nodes, flow links, etc.).
```
arthurvd:
```suggestion
   !> Apply a polygon mask to the provided mask array, based on the provided
   !! target mask polygon file and location type (flow nodes, flow links, etc.).
   !! The polygon mask is applied in AND/intersection fashion, that is:
   !! the existing values of the input mask are respected: existing 0's remain 0's
   !! and existing 1's will only remain 1 if they lie inside the polygon mask too.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+   subroutine apply_polygon_mask(mask, location_type, target_mask_file, ierr)
+      use m_flowgeom, only: xz, yz, kcs, ndx, lnx
+      use timespace_parameters, only: LOCTP_POLYGON_FILE
+      use timespace, only: selectelset_internal_nodes, selectelset_internal_links
+      use dfm_error, only: DFM_NOERR, DFM_WRONGINPUT, DFM_NOTIMPLEMENTED
+
+      ! Parameters
+      integer, dimension(:), intent(inout) :: mask !< Mask array for the target element set.
+      integer, intent(in) :: location_type !< The location type parameter (one from fm_location_types::UNC_LOC_*) for this quantity's target element set.
+      character(len=*), intent(in), optional :: target_mask_file !< File name of the target mask file (*.pol). When empty, 100% masking is assumed.
```
arthurvd: Why is this optional? Applying a polygon without a polygon ;-)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
```diff
@@ -0,0 +1,344 @@
...
+      integer, intent(out), optional :: ierr !< Result status (DFM_NOERR if succesful, or different if mask could not be constructed for this quantity's location).
```
arthurvd: `ierr` does not need to be optional here I guess? I'd remove the optional.
But if you decide to keep it, you should add `if (present` checks in several lines below.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -112,21 +112,6 @@ subroutine test_averaging_params_to_transformcoef() bind(C)
    end subroutine test_averaging_params_to_transformcoef
    !$f90tw)
 
-   !$f90tw TESTCODE(TEST, test_init_spatial_field, test_parse_location_type, test_parse_location_type,
```
arthurvd, line 115: It's a pity we lose this unit test. Maybe place it inside a unit test module for the `fm_location_types` module?

arthurvd (changes requested):
Hi @thomas-pijls, you solved my main point from last time: the 1d/2d masking and polygon masking is now properly combined. So, runtime behavior is now overall correct :-)

I've placed several additional remarks regarding module structure, some cleanup and some minor bugs + docs.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
thomas-pijls: Hi, could you discuss this with @mihai-pricope ? Mihai suggested and liked the current approach, using an interface that accepts both integer and string arguments. I find both options acceptable, it's a bit of a style choice

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
thomas-pijls: Good point, moved the function

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
thomas-pijls: Done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
thomas-pijls: Copy paste error ;)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
thomas-pijls: I discussed with Mihai and we both think that the unit test was 'overbodig', it tested very simple behaviour

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_location_types.f90
```diff
@@ -147,4 +156,33 @@ function get_location_specifier_string(location_specifier) result(string)
          call mess(LEVEL_ERROR, 'Programming error, please report: unrecognised location_specifier in fm_location_types::get_location_specifier_string(), location_specifier = ', location_specifier)
       end select
    end function get_location_specifier_string
+
+   !> Parse a locationType= string ('1d', '2d', '1d2d', 'all') to the
+   !! ILATTP_* enum used by prepare_lateral_mask.
```
arthurvd:
```suggestion
   !! SPATIAL_LOCATION_* enum.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
arthurvd: That's OK, we'll discuss it with the three of us, instead of ping-ponging.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
arthurvd: Unit test are supposed to be testing simple behavior. The question whether these were too simple is subjective. Ask @KoenBussemakerDeltares what he thinks about this particular one.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/flowgeom_mask.f90
arthurvd: Discussed in person yesterday in Delft. Separation of I/O and masking logic is now agreed upon. Only a single construct_nask remains with only integer enum input. Code does that now.

arthurvd (approved):
The overall structure is now correct.

mihai-pricope (approved):
Looks good to me

PR #950, opened 2026-06-08
wetting-and-drying implemented for ice modelling in D-Flow FM

goede2606:
# What was done 

Wetting-and-drying has been implemented for ice modelling in D-Flow FM. The icriterion is that the water depth in water levels points, in which also the ice thickness and snow thickness is computed, is larger than 1.1 times the minimum water depth in flow links (=EPSHU value). In general the water depth in cell centres will be close to the EPSHU value for flow links in case of drying. A 10% larger value has been taken in order to stay 'at the safe side' and to avoid flipflopping between a wet and a dry status.

- Noted that this is a quite complex criterion, because the drying-and-flooding-algorithm is D-Flow is somewhat unclear. 
- Moreover, the is no criterion in cell centres like for the flow links (with water depth < EPSHU).
- Noted that variable EPSHS is unsuited for a wetting-and-drying check, because its value is set to 0.2 * EPSHU. This doesn't work in practice because in dry areas the water depth in cell centres will be close to the water depth at flow links, being close to the EPSHU value. 

   

# Evidence of the work done 

- [ ]	Video/figures \
- [X]	Not applicable 

# Tests 
- [ ] Tests updated \
The testcase to be updated to the testbench is in: 
p:\dflowfm\maintenance\JIRA\09000-09999\09670\9670_winter_1997_drying_flooding\winter97_freshw_2D.mdu

# Documentation  
- [ ]	Documentation updated \
In Section 16.10 of the D-Flow FM user manual the limitation has to be removed that wetting and drying isn't possible in combination with ice modelling.

This issue is connected to https://issuetracker.deltares.nl/browse/UNST-9670.


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,38 +445,44 @@ subroutine update_icecover()
...
+                     ice_area_fraction(n) = 1.0_fp
+                  else
+                     ice_thickness(n) = 0.0_fp
+                     snow_thickness(n) = 0.0_fp
+                     ice_area_fraction(n) = 0.0_fp
+                     ice_temperature(n) = celsius_to_kelvin(0.0_fp)
+                  end if
                end if
-            end if
+            endif
```
arthurvd:
```suggestion
            end if  
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,38 +445,44 @@ subroutine update_icecover()
...
-               if (qh_air2ice(n) > qh_ice2wat(n) .and. snow_thickness(n) > 0.0_fp) then
-                  ! melting of snow due to heat exchange with air
-                  snow_thickness_change = dts * (-qh_air2ice(n) + 0.0_fp) / snow_latentheat
-                  if (-snow_thickness_change < snow_thickness(n)) then
-                     ! snow melt less than snow layer thickness
-                     snow_thickness(n) = snow_thickness(n) + snow_thickness_change
-                     ice_thickness_change = 0.0_fp
+             
+            ! determine wetting-drying status   
+            water_depth = s1(n) - bl(n)
```
arthurvd: No need to locally recompute waterdepth, it is already in `m_flow` variable `hs`.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,38 +445,44 @@ subroutine update_icecover()
...
+            if (water_depth > 1.1_dp * epshu) then
```
arthurvd: Please move this 1.1*epshu to a dedicated epsilon parameter in the FM ice module: `m_fm_icecover`.
If you call it `eps_hs_ice` then it is similar to the already existing `epshstem` which is used in switching on/off the heat flux calculation.

I also noticed that the call
```
      call fm_ice_null()
```
inside `subroutine resetModel()` is in the wrong place. It should be inside `subroutine resetFullFlowModel()`, for example after/close to `call default_heatfluxes()`.

Lastly, you can only do the multiplication by epshu *after* an MDU is loaded, that means: inside `subroutine flow_modelinit()`. You have an 'Ice init' there already. I would rename `fm_ice_alloc()` to `fm_ice_init()` and add the statement `eps_hs_ice = 1.1_dp*epshu` there.

arthurvd (changes requested):
Hi @goede2606, see my 3 remarks.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
hrajagers: @arthurvd I see that D-Flow FM `m_flowgeom` includes a variable called `kfs` described as "node code flooding". This variable is for instance used in `write_array_with_dmiss_for_dry_cells_into_netcdf_file` to set the values of dry cells equal to `dmiss`. Could this flag be used here in the ice cover code ... or is it not reliable at this point during a time step?

goede2606: Variabele KFS, which indeed is meant for the wetting-drying status - isn't suited for our check because it is based on the criterion "Volume(n) > 0.0". Thus,  for very thin waterdepths (HS) computational cells are still wet.  This also holds for (much) smaller values than EPSHU. which is used for wetting-drying at flow links.
Our goal is to have a criterion that is close to the criterion for flow links. In dried ares the HS values are often close to EPSHU. For example a few percent smaller. In these cells we would like to switch off ice growth or melt. Therefore, as a first guess we propose to apply "wetting"/ ïce growth" only for "HS >= 1.1 * EPSHU".

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
goede2606: Done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
goede2606: Done.

goede2606:
@arthurvd,
 you also proposed to introduce an "ice threshold" parameter for wetting-drying, like EPSHU and EPSHS. Therefore, in the ice datastructure ICE_WETTING_DRYING_THRESHOLD has been added, lilke other parameters like ICE_ALBEDO and SNOW_ALBEDO.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
hrajagers: Okay for the time being (I don't like that every process that needs to know about drying/flooding uses a different threshold, but avoiding that is out of scope for this pull request).

Furthermore, I agree that ice growth should be stopped when the water depth is (too) small, but I'm not sure whether ice melt should be stopped ... although ice melt on dry points may require a slightly different thermal check. Let's for instance take a reservoir covered with ice. The water level in the reservoir drops, and some cell along the edges become dry. Then spring comes and melts the ice ... except on the dry cells? This is what end users would typically not expect.

I'm okay to check whether this for the time being fixes the problem observed in the application, but I expect a follow-up issue to ask for appropriate melting of the ice on the land.

goede2606: @hrajagers,
I agree with your remark that ice melt should not be stopped in case of drying. After my vacation (29 June - ...) I will implement this. Nevertheless, it will be interesting to check whether with the current implementation the unrealistic ice results will disappear for the Bohai Sea and the Botnic Gulf applications.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -73,13 +73,14 @@ module m_fm_icecover
...
    real(fp), pointer :: ice_frcuni !< module pointer to frict_val inside ice_data
+   real(fp), pointer :: ice_wetting_drying_threshold !< module pointer to ice_wetting_drying_threshold
 
    real(fp), pointer :: snow_albedo !< module pointer to snow_albedo inside ice_data
    real(fp), pointer :: snow_conductivity !< module pointer to snow_conductivity inside ice_data
    real(fp), pointer :: snow_latentheat !< module pointer to snow_latentheat inside ice_data
 
    character(len=*), parameter :: MDU_ICE_CHAPTER = 'ice' !< name of the ice chapter in the mdu file
-
+
```
hrajagers:
```suggestion
 
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,38 +446,52 @@ subroutine update_icecover()
...
+                     ice_area_fraction(n) = 1.0_fp
+                  else
+                     ice_thickness(n) = 0.0_fp
+                     snow_thickness(n) = 0.0_fp
+                     ice_area_fraction(n) = 0.0_fp
+                     ice_temperature(n) = celsius_to_kelvin(0.0_fp)
+                  end if
                end if
-            end if
+            end if
```
hrajagers:
```suggestion
            end if
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -73,13 +73,14 @@ module m_fm_icecover
...
    real(fp), pointer :: ice_frcuni !< module pointer to frict_val inside ice_data
+   real(fp), pointer :: ice_wetting_drying_threshold !< module pointer to ice_wetting_drying_threshold
 
    real(fp), pointer :: snow_albedo !< module pointer to snow_albedo inside ice_data
    real(fp), pointer :: snow_conductivity !< module pointer to snow_conductivity inside ice_data
    real(fp), pointer :: snow_latentheat !< module pointer to snow_latentheat inside ice_data
 
    character(len=*), parameter :: MDU_ICE_CHAPTER = 'ice' !< name of the ice chapter in the mdu file
-
+
```
hrajagers:
```suggestion

```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,36 +445,50 @@ subroutine update_icecover()
...
+            ! Determine wetting-drying status for ice growth or melt for computational cell (n).
+            !
+            ! This is a trial-and-error approach. For example, variabele KFS, which indeed is meant for the wetting-drying status - isn't suited for our check because
+            ! it is based on the criterion "Volume(n) > 0.0". Thus, for very thin waterdepths (HS) computational cells are still wet. This also holds for (much) smaller
+            ! values than EPSHU. which is used for wetting-drying at flow links. The goal is to have a criterion that is close to the criterion for flow links. 
+            ! In dried ares the HS values are often close to EPSHU. For example a few percent smaller. In these cells we would like to switch off ice growth or melt. 
+            ! Therefore, as a first guess we propose to apply ïce growth or melt only for "HS > 1.1 * EPSHU". 
+            ! 
+            ice_wetting_drying_threshold = 1.1_dp * epshu
+
```
arthurvd: This initialization of the parameter should not be done *each* update and even more not *inside* an `ndx` loop. Instead, place it in a subroutine `fm_ice_init()`, see the original conversation on: https://github.com/Deltares/Delft3D/pull/950#discussion_r3386220236

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
arthurvd: Conversation continues after Erik's vacation. My main wish is that any threshold we set is truly understood. Currently, some models (e.g. in Gulf of Bothnia) apparently show unstable behavior, but is is unclear whether closed flow links (following from epshu checks) have a true causal relation to this, or not.

I will fix the init/reset subroutine structure, such that it matches most other fm reset/default structure.

arthurvd (dismissed):
See new comment + pending conversation from last week.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
arthurvd: I fixed the default/reset structure (was null/clr).
Better understanding of ice and drying in a followup issue: UNST-10027.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
arthurvd: Fixed in 5a25892a349729fd851c41f87bebb6e6d937d3d3.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -442,36 +447,40 @@ subroutine update_icecover()
...
-            if (air_temperature(n) < 0.0_fp .or. ice_thickness(n) > 0.0_fp) then
-               if (qh_air2ice(n) > qh_ice2wat(n) .and. snow_thickness(n) > 0.0_fp) then
-                  ! melting of snow due to heat exchange with air
-                  snow_thickness_change = dts * (-qh_air2ice(n) + 0.0_fp) / snow_latentheat
-                  if (-snow_thickness_change < snow_thickness(n)) then
-                     ! snow melt less than snow layer thickness
-                     snow_thickness(n) = snow_thickness(n) + snow_thickness_change
-                     ice_thickness_change = 0.0_fp
+
+            if (hs(n) > ice_wetting_drying_threshold) then
```
hrajagers: Consider replacing this by an equivalent
`if (hs(n) <= ice_wetting_drying_threshold) then`
`   cycle`
`end if`
Then it's immediately clear that nothing happens under these conditions, and it avoid additional indentation.

hrajagers (approved):
Approving the changes as is, but I would personally prefer a `cycle` instead of one all encompassing `if` block, but technically they have the same result.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
arthurvd: Done in 479f2816b633cfa65846587b69030b9de7694646.

hrajagers (approved):
Looks all good now ... we'll revisit the ice growth/melting (on land) under UNST-10027.

PR #952, opened 2026-06-08
Add python container to run linux workloads (Or if you will... run Python CI tools in container)

rene-deltares:
# What was done 

Don't run python executable on the agent, use container.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/ci/dockerfiles/linux/python.Dockerfile
+++ b/ci/dockerfiles/linux/python.Dockerfile
@@ -2,7 +2,7 @@
 
 # note that although the BASE_IMAGE_URL argument allows you to easily change the base image,
 # all the following code assumes that you're running an Alma Linux or compatible environment.
-ARG BASE_IMAGE_URL=containers.deltares.nl/base_linux_containers/8-base:latest
+ARG BASE_IMAGE_URL=containers.deltares.nl/docker-proxy/almalinux/8-base:latest
 
 FROM ${BASE_IMAGE_URL}
 
--- a/ci/dockerfiles/linux/python.Dockerfile
+++ b/ci/dockerfiles/linux/python.Dockerfile
@@ -1,0 +1,25 @@
+# syntax=containers.deltares.nl/docker-proxy/docker/dockerfile:1.4
+
+# note that although the BASE_IMAGE_URL argument allows you to easily change the base image,
+# all the following code assumes that you're running an Alma Linux or compatible environment.
+ARG BASE_IMAGE_URL=containers.deltares.nl/base_linux_containers/8-base:latest
+
+FROM ${BASE_IMAGE_URL}
+
+ARG PYTHON_VERSION=3.12
+
+RUN curl -LsSf https://astral.sh/uv/0.11.19/install.sh | UV_INSTALL_DIR=/usr/bin sh
+
+RUN <<EOF
+set -eo pipefail
+
+# Configure `uv` to install python and tools in the `/opt/uv` directory. So they are shared with all users.
+export UV_PYTHON_INSTALL_DIR=/opt/uv/share/uv/python
+export UV_PYTHON_BIN_DIR=/opt/uv/bin
+
+# Install python and python tools.
+uv python install ${PYTHON_VERSION} --default
+EOF
+
+# Add python and uv tools to PATH for all users.
+ENV PATH=/opt/uv/bin:$PATH
--- a/ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
+++ b/ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
@@ -37,6 +37,10 @@ object TestPythonCiTools : BuildType({
         cleanCheckout = true
     }
 
+    params {
+        param("docker_image", "containers.deltares.nl/delft3d-dev/delft3d-python:alma8-python3.12")
+    }  
+
     triggers {
         vcs { 
             branchFilter = "+:pull/*"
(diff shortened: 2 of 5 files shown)


ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
```diff
@@ -43,65 +44,36 @@ object TestPythonCiTools : BuildType({
...
-                    --cov=.
-                """.trimIndent()
-            }
-            executionMode = BuildStep.ExecutionMode.ALWAYS
+            scriptContent = """
+                #!/usr/bin/env bash
+                set -eo pipefail
+                uv run ruff format --diff
+                uv run ruff check --output-format=junit --output-file=ruff.xml
+                uv run mypy ci_tools --junit-xml=mypy.xml
```
harmenwierenga, line 76: Did it hurt a little not to use ty?

ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
```diff
@@ -43,65 +44,36 @@ object TestPythonCiTools : BuildType({
...
-                    --junitxml=pytest.xml
-                    --cov-report=html
-                    --cov=.
-                """.trimIndent()
-            }
-            executionMode = BuildStep.ExecutionMode.ALWAYS
+            scriptContent = """
+                #!/usr/bin/env bash
+                set -eo pipefail
+                uv run ruff format --diff
```
harmenwierenga, line 74: I think that it's fine to have these steps run in one script step, but it does make failures a little less clear. Previously, the name of the step signaled what failed

ci/dockerfiles/linux/python.Dockerfile
```diff
@@ -0,0 +1,25 @@
+# syntax=containers.deltares.nl/docker-proxy/docker/dockerfile:1.4
+
+# note that although the BASE_IMAGE_URL argument allows you to easily change the base image,
+# all the following code assumes that you're running an Alma Linux or compatible environment.
+ARG BASE_IMAGE_URL=containers.deltares.nl/base_linux_containers/8-base:latest
```
harmenwierenga: I think that topops wants us to use docker proxy

ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
```diff
@@ -43,65 +44,36 @@ object TestPythonCiTools : BuildType({
...
+                uv run ruff check --output-format=junit --output-file=ruff.xml
+                uv run mypy ci_tools --junit-xml=mypy.xml
+                uv run pytest --junitxml=pytest.xml --cov-report=html --cov=.
+            """.trimIndent()
+            dockerImage = "containers.deltares.nl/delft3d-dev/delft3d-python:3.12"
+            dockerImagePlatform = ScriptBuildStep.ImagePlatform.Linux
+            dockerPull = true
+            dockerRunParameters = """
+                --rm
+                --mount type=volume,source=uv-cache-python-ci-tools,destination=/root/.cache/uv
```
harmenwierenga: Can be removed?

ci/dockerfiles/linux/python.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM ${BASE_IMAGE_URL}
+
+ARG PYTHON_VERSION=3.12
+
+RUN curl -LsSf https://astral.sh/uv/0.11.19/install.sh | UV_INSTALL_DIR=/usr/bin sh
```
harmenwierenga, line 11: There almost seems to be a binary choice: either we always take the latest version (can break stuff silently, not good), or we hardcode the version and chances are that it won't be updated for the coming three years. A bit similar for our python version everywhere. Do you know if there is a middle way?

harmenwierenga (approved):

ci/dockerfiles/linux/python.Dockerfile
matthijs-deltares: I think so as well

rene-deltares: Me too!

ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
rene-deltares: I understand, I think I'll add `set -x` so that it's a little more clear what command is being run/failing. I find the separate steps a bit verbose with all the repeated docker stuff. But hold on... I have an idea...

UPDATE: It was not a very good idea.

ci/dockerfiles/linux/python.Dockerfile
rene-deltares: For these install scripts there is no middle way, currently. No way to say: I want 0.11.19 or higher, but lower than 0.12.0 (or 1.0.0). For such a tool I think it's fine. If there's some feature we're missing from a future release, or there is a problem with this version we can always update manually. In my experience, automatic updates break things more often than staying on an old version.

ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
rene-deltares: I haven't been using python much, lately so not really. But I would like to replace this with ty at some point because it is much snappier than mypy

ci/teamcity/Delft3D/ciUtilities/TestPythonCiTools.kt
rene-deltares: Yes, I think so. The `uv sync` from the previous step copies the libs from the uv cache to the `.venv` in the current dir.

harmenwierenga (approved):
Nice!

harmenwierenga (approved):

PR #953, opened 2026-06-09
Update build type descriptions.

robin-deltares:
Only non-'CI utilities'

### DIFF
--- a/ci/teamcity/Delft3D/linux/runtimeContainers.kt
+++ b/ci/teamcity/Delft3D/linux/runtimeContainers.kt
@@ -9,7 +9,11 @@ import Delft3D.linux.containers.*
 
 object LinuxRuntimeContainers : BuildType({
 
-    description = "Build two separate container images: one for running the Delft3D software and the other for executing its tests."
+    description = ""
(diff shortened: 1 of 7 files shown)


ci/teamcity/Delft3D/linux/runtimeContainers.kt
```diff
@@ -9,7 +9,11 @@ import Delft3D.linux.containers.*
 
 object LinuxRuntimeContainers : BuildType({
 
-    description = "Build two separate container images: one for running the Delft3D software and the other for executing its tests."
+    description = ""
```
tersteegh, line 12: why is this line needed

tersteegh (changes requested):

tersteegh (approved):

ci/teamcity/Delft3D/linux/runtimeContainers.kt
tribbin: Because it builds two containers for different purposes. In Windows these are two separate build configurations so this description avoids confusion.

PR #911, opened 2026-05-27
UNST-9795: Migrate source/sink variables and functions to own module and put variables in own type

thomas-pijls:
# What was done 

- Created the new `m_source_sink` module in the newly created `source_sink.f90` file.
- Moved `addsorsin`, `setsorsin`, `reallocsrc` subroutines to `m_source_sink` and deleted their now empty files.
- Moved all source/sink variables in `fm_external_forcings_data.f90` to `m_source_sink`.
- Put all source/sink variables in the new `SourceSinks` derived type. `SourceSinks` is structured as a structure-of-arrays (SoA).
- Renamed the `reallocsrc` subroutine to `resize_source_sinks` and mady it type-bound to `SourceSinks`
- Updated all references to the old source/sink variables to the new type-attributes in the code base.

# Evidence of the work done 

- [ ]	Video/figures \
<add video/figures if applicable> 
- [x]	Clear from the issue description 
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
<add testcase numbers if applicable, Issue number>
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
<add description of changes if applicable, Issue number> 
- [x]	Not applicable 

# Issue link

PR #968, opened 2026-06-12
All/bugfix/unst 9795 revert merge

harmenwierenga:
Revert the merges of #965 and #911 since they break tests


harmenwierenga:
Merged a fix instead in #969, won't revert the merge

PR #970, opened 2026-06-12
UNST-9887: couple atmospheric stability module to dflow-fm

sumihar:
# What was done 

- add the following interfaces in fm_external_forcings.f90: prepare_wind, prepare_air_pressure_temperature_dew_point_temperature, compute_air_water_interaction_most_fluxes

- refactor calculate_wind_stresses(), i.e. prepare_wind_model_data() is taken out of it.

- introduce the call to m_atmospheric_stability via compute_air_water_interaction_most_fluxes() from fm_external_forcings_update > set_external_forcings() and m_flow_initimestep > flow_initimestep()

- introduce setting wind stress to m_atmospheric_stability wind stress output in m_setwindstress > setwindstress()

- introduce setting sensible_heat_flux and latent_heat_flux to m_atmospheric_stability output in m_heatu > heatu()

- refactor heatun.f90 to make it clearer what fluxes components are computed and introduce a condition for using m_atmospheric_stability output 

- create m_get_surface_temperature for obtaining water surface temperature

- create m_flowgeom_interpolate containing routines for link-to-node and node-to-link interpolations

- introduce the following mdu keywords and use them in the respective places in the codes:
[meteo]
AirSeaInteractionModel              = 1                                                                   # Air-water interaction model (0: none, 1: Bulk formulae for heat and momentum fluxes based on Monin-Obukhov Similarity Theory)
StabilityFunction                   = 1                                                                   # Atmospheric stability function (0: none, 1: ECMWF atmospheric stability function)
FreeConvection                      = 0                                                                   # Switch for free convection (0: off, 1: on)
QsatFactor                          = 1.0                                                                 # Salinity reduction factor for saturation humidity calculation, 1.0 means no reduction (0.0, 1.0]

A first test was done using dcsm5 and ECMWF-HRES meteo input, with AirSeaInteractionModel = 1 and StabilityFunction = 0. Here, the results are compared to those of a simulation without this feature (with ICdtyp=4/Charnock and Relativewind=0). Expected results: the wind stress calculated in the two simulations should be similar. This should lead to similar water levels. The figure below shows that this is the case. 

20260612_wldiff_relwind0

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable; we will ask user to do functionality tests; the mdu keywords are not definitive yet. Testcase will be created when the functionality has been tested and mdu keywords are definitive.

# Documentation  
- [X]	Documentation will be updated \
https://issuetracker.deltares.nl/browse/UNST-9893
- [ ]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9887


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
```diff
@@ -0,0 +1,132 @@
...
+      real(kind=dp), intent(in) :: xcomp_node(:)  !< x-component of vector at flow nodes
+      real(kind=dp), intent(in) :: ycomp_node(:)  !< y-component of vector at flow nodes
+      real(kind=dp), intent(out) :: xcomp_link(:) !< x-component of vector at flow links
+      real(kind=dp), intent(out) :: ycomp_link(:) !< y-component of vector at flow links
+      integer, intent(in) :: number_of_links  !< number of flow links
+
+      integer :: i_link, k1, k2
+
+      xcomp_link = 0.0_dp
+      ycomp_link = 0.0_dp
```
FlorisBuwaldaDeltares: this doesn't do anything, since they will be overwritten by the subsequent loop

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
```diff
@@ -0,0 +1,132 @@
...
+
+      do i_link = 1, number_of_links
+         k1 = ln(1, i_link)
+         k2 = ln(2, i_link)
+
+         xcomp_link(i_link) = acl(i_link) * xcomp_node(k1) + (1.0_dp - acl(i_link)) * xcomp_node(k2)
+         ycomp_link(i_link) = acl(i_link) * ycomp_node(k1) + (1.0_dp - acl(i_link)) * ycomp_node(k2)
+      end do
```
FlorisBuwaldaDeltares: do concurrent is good here, as it tells the compiler it can safely optimize as there are no dependencies between loops. Additionally, we could reuse the node maps from m_prefetch. The AI wrote this as a suggestion:

`   !> Interpolate vector components from flow nodes (cell centres) to flow links.
   subroutine node_to_link_vector(xcomp_node, ycomp_node, xcomp_link, ycomp_link, number_of_links)
      use m_prefetch, only: node_map_1, node_map_2
      use m_flowgeom, only: acl

      real(kind=dp), intent(in), dimension(:), contiguous :: xcomp_node  !< x-component of vector at flow nodes
      real(kind=dp), intent(in), dimension(:), contiguous :: ycomp_node  !< y-component of vector at flow nodes
      real(kind=dp), intent(out), dimension(:), contiguous :: xcomp_link !< x-component of vector at flow links
      real(kind=dp), intent(out), dimension(:), contiguous :: ycomp_link !< y-component of vector at flow links
      integer, intent(in) :: number_of_links  !< number of flow links

      integer :: L
      real(kind=dp) :: w1

      do concurrent (L = 1:number_of_links)
         w1 = acl(L)
         xcomp_link(L) = w1 * xcomp_node(node_map_1(L)) + (1.0_dp - w1) * xcomp_node(node_map_2(L))
         ycomp_link(L) = w1 * ycomp_node(node_map_1(L)) + (1.0_dp - w1) * ycomp_node(node_map_2(L))
      end do
   end subroutine node_to_link_vector`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
```diff
@@ -0,0 +1,132 @@
...
+         do LL = 1, nd(i_node)%lnx
+            LLL = abs(nd(i_node)%ln(LL))
+            k = 1
+            if (nd(i_node)%ln(LL) > 0) then
+               k = 2
+            end if
+            xcomp_node(i_node) = xcomp_node(i_node) + xcomp_link(LLL) * wcL(k, LLL)
+            ycomp_node(i_node) = ycomp_node(i_node) + ycomp_link(LLL) * wcL(k, LLL)
+         end do
+      end do
```
FlorisBuwaldaDeltares: it is very inefficient to loop over the nodes, it creates lots of tiny little stack arrays. Better just to loop over the links! see this snippet in setumod:

Image

If we loop over all the links, we trust that every node has received has received all contributions.

This is what the AI suggests:

`   !> Interpolate scalar value from flow links to flow nodes (cell centres).
   !! Uses link-based scatter: contiguous reads from link arrays, scatter-add to nodes.
   subroutine link_to_node_scalar(scalar_link, scalar_node, number_of_links, number_of_nodes)
      use m_prefetch, only: node_map_1, node_map_2, wcL_1, wcL_2

      real(kind=dp), intent(in), dimension(:), contiguous :: scalar_link  !< scalar values at flow links
      real(kind=dp), intent(out), dimension(:), contiguous :: scalar_node !< scalar values at flow nodes
      integer, intent(in) :: number_of_links  !< number of flow links
      integer, intent(in) :: number_of_nodes  !< number of flow nodes

      integer :: L, k1, k2

      scalar_node = 0.0_dp

      do L = 1, number_of_links
         k1 = node_map_1(L)
         k2 = node_map_2(L)
         scalar_node(k1) = scalar_node(k1) + scalar_link(L) * wcL_1(L)
         scalar_node(k2) = scalar_node(k2) + scalar_link(L) * wcL_2(L)
      end do
   end subroutine link_to_node_scalar`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -327,10 +340,10 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
          if (buoyancy_parameter > 0.0_dp) then
             free_convection_velocity = buoyancy_parameter * KINEMATIC_VISCOSITY_AIR / PRANDTL_NUMBER_SQUARED
             free_convection_velocity = free_convection_coefficient * free_convection_velocity**(1.0_dp / 3.0_dp)
-            if (.not. sensible_heat_flux_available) then
+            if (.not. (sensible_heat_flux_available .or. air_water_interaction_model_is_used)) then
                free_convective_sensible_heat_flux = min(0.0_dp, -air_density_in_cell * SPECIFIC_HEAT_AIR * free_convection_velocity * (surface_temperature - air_temperature_in_cell))
             end if
-            if (.not. latent_heat_flux_available) then
+            if (.not. (latent_heat_flux_available .or. air_water_interaction_model_is_used)) then
```
FlorisBuwaldaDeltares: all these extra .or.'s are a bit ugly. Another option is that if air_water_interaction_model_is_used, latent_heat_flux_available is set to true. What do you think, is that cleaner or only confusing?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
```diff
@@ -304,6 +303,53 @@ module subroutine set_external_forcings(time_in_seconds, initialization, iresult
...
+      real(kind=dp), dimension(:), allocatable :: surface_temperature_kelvin, air_temperature_kelvin, dew_point_temperature_kelvin
+      type(t_options) :: atm_stability_options
+
+      allocate(windx(ndx))
+      allocate(windy(ndx))
+      allocate(charnock(ndx))
+      allocate(surface_temperature(ndx))
+      allocate(surface_temperature_kelvin(ndx))
+      allocate(air_temperature_kelvin(ndx))
+      allocate(dew_point_temperature_kelvin(ndx))
```
FlorisBuwaldaDeltares: this is a lot of allocations that have to happen every time (ndx is pretty big). an option is to give these arrays the save attribute. alternatively, celcius to kelvin is such a trivial operation that it would be almost be preferrable to change compute_scales_and_fluxes to accept celcius temperatures and to the celcius => kelvin operation itself without having to allocate all these extra arrays..

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/get_surface_temperature.f90
```diff
@@ -0,0 +1,71 @@
...
+
+      real(kind=dp), intent(out) :: surface_temperature(ndx)
+
+      integer :: n, kb, kt
+
+      do n = 1, ndx
+         call getkbotktop(n, kb, kt)
+         surface_temperature(n) = tem1(kt)
+      end do
+   end subroutine get_surface_temperature
```
FlorisBuwaldaDeltares: am i missing something or is this routine never called anywhere?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -88,7 +88,12 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
       real(kind=dp) :: surface_albedo !< local surface albedo (may differ from albedo when ice/snow is present)
       real(kind=dp) :: salinity !< water salinity (ppt)
       integer :: cell_index_3D, k_bot, k_top, k2, L, LL, j, j2, ncols
+      logical :: air_water_interaction_model_is_used = .false.
 
+      if (air_water_interaction_model == AIR_WATER_INTERACTION_MODEL_MOST) then
+         air_water_interaction_model_is_used = .true.
+      end if
```
FlorisBuwaldaDeltares: if we know the model is used if air_water_interaction_model == AIR_WATER_INTERACTION_MODEL_MOST, do we really need an "air_water_interaction_model_is_used" variable?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
```diff
@@ -267,6 +267,28 @@ integer function flow_flowinit() result(error)
...
+      call initialize_salinity_and_temperature_with_nudge_variables()
+
+      if (jasal > OFF) then
+         call fill_constituents_with_salinity()
+      end if
+      if (itemp > OFF) then
+         call fill_constituents_with_temperature()
+      end if
+
+      call initialise_density_at_cell_centres()
```
FlorisBuwaldaDeltares, line 290: can you comment on why these initializations had to be moved?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
```diff
@@ -133,29 +133,27 @@ module subroutine set_external_forcings(time_in_seconds, initialization, iresult
       if (ja_airdensity > 0) then
          call get_timespace_value_by_item_and_consider_success_value(item_air_density, time_in_seconds)
       end if
+      if (ja_computed_airdensity == 1 .or. air_water_interaction_model == AIR_WATER_INTERACTION_MODEL_MOST) then
+         ! prepare air pressure, air temperature and dew point temperature values
```
FlorisBuwaldaDeltares: i don't think this is a very useful comment

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -169,8 +174,13 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
             surface_temperature = water_temperature_in_cell
          end if
 
-         relative_humidity_in_cell = min(1.0_dp, max(0.0_dp, 0.01_dp * relative_humidity(n)))
-         cloudiness_in_cell = min(1.0_dp, max(0.0_dp, 0.01_dp * cloudiness(n)))
+         if (allocated(relative_humidity)) then
```
FlorisBuwaldaDeltares, line 172: why did you add all those if (allocated) checks? why was this previously not necessary? if we're inside the heatun routine shouldn't rel humidity always be allocated?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1632,6 +1632,11 @@ subroutine readMDUFile(filename, istat)
          end if
       end if
 
+      call prop_get(md_ptr, 'meteo', 'AirSeaInteractionModel', air_water_interaction_model)
+      call prop_get(md_ptr, 'meteo', 'StabilityFunction', atmospheric_stability_function)
+      call prop_get(md_ptr, 'meteo', 'FreeConvection', free_convection)
+      call prop_get(md_ptr, 'meteo', 'QsatFactor', salinity_reduction_factor_saturation_humidity)
```
FlorisBuwaldaDeltares, line 1638: is 'QsatFactor' the best name? Q is a pretty generic letter for some kind of flux, no hint that it is related to salinity or humidity?

FlorisBuwaldaDeltares (commented):
few comments, but otherwise great work!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
sumihar: That is right. The same thing holds for scalar_link in node_to_link_scalar(). I have removed them.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
sumihar: Good suggestion about using 'concurrent', suggestion has been applied.

However, about reusing node_map_1 and node_map_2 or m_prefetch: they are currently not public variables, hence not accessible from this module. I wonder if it is more preferable to modify m_prefetch than just using the array _ln_ here as it is now. I suppose using _ln_ directly is cleaner as it avoids an extra layer of module dependency.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
sumihar: This is a nice trick indeed: looping over links instead of over nodes. It can potentially be more efficient than the current implementation for sequential execution. 

However, it prevents the potential of using 'do concurrent'. I mean, In the current implementation, it is possible to insert 'do concurrent' in the outer loop. With the loop-over-links, it is not possible to use 'do concurrent', because a node can be processed at different iterations.

What is wise? I guess, as most real-life dflow-fm models run on multiple cores, the current implementation with 'do concurrent' is potentially more efficient? Let's discuss. For the time being, I inserted 'do concurrent' here.

Note that arrays wcL_1 and wcL_2 don't exist now. I guess, the AI meant wcL(1,L) and wcL(2,L).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/get_surface_temperature.f90
sumihar: Oeps!

In the beginning, I thought we had to use _tem1_ for water temperature. When working on this, I came to realize that _tem1_ is only updated in the beginning and at the end of a simulation. It is temperature in variable _constituents_ that is updated during the simulation. 

To be sure, both routines were kept for a while. Now the one based on _tem1_ has been removed.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
sumihar: Yes, that is true. 

This extra flag was introduced to avoid having to write a longer version of it, i.e. ".or. air_water_interaction_model == AIR_WATER_INTERACTION_MODEL_MOST" in four different places in the code.

More about this later, in reaction to your next review point below.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
sumihar: That is indeed what I had expected too. But then I realized that all those variables are set only when the respective forcings are specified as model input. 

Note that this block of codes had been intended for TEMPERATURE_MODEL_COMPOSITE. This temperature model requires users to specify all those input forcings. 

Now we are using it also for the case of air_water_interaction_model_is_used. In this case, it is not necessary for user to specify those input forcings.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
sumihar: I agree, I find them ugly too; hence the introduction of air_water_interaction_model_is_used flag. Clearly it doesn't help much : )

I am afraid that setting e.g. latent_heat_flux_available if air_water_interaction_model_is_used can be confusing. I mean, in that case, the flag latent_heat_flux_available won't be exclusively used to indicate that user provides heat fluxes as input forcing. 

For the time being, I proposed to remove the flag air_water_interaction_model_is_used and apply the fluxes directly to compute total_heat_flux, taking longwave_radiation_flux into consideration:

         if (air_water_interaction_model == AIR_WATER_INTERACTION_MODEL_MOST) then
            total_heat_flux = latent_heat_flux(n) + sensible_heat_flux(n) + longwave_radiation_flux 
         else
            total_heat_flux = forced_latent_heat_flux + forced_sensible_heat_flux + longwave_radiation_flux + free_convective_sensible_heat_flux + free_convective_latent_heat_flux
         end if

This may be clearer than the previous implementation, because in this case the latent_heat_flux and sensible_heat_flux both may contain 'forced' and 'free_convection', computed in atmospheric_stability module.

For the longer run, perhaps good to refactor heatun to contain at least 5 subroutines, each computes one of the heat components: forced_latent_heat_flux, forced_sensible_heat_flux, longwave_radiation_flux, free_convective_sensible_heat_flux, free_convective_latent_heat_flux. Then we can introduce 'if then' blocks for the various temperature models: each contains specific combinations of those heat fluxes.

What do you think about this?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
sumihar: For using the atmospheric_stability module, initial surface temperature is needed. Hence, temperature constituent should be available before calling the routine set_external_forcings(). 

I decided not only to move temperature, but also salinity, for the sake of clarity: they both form a block required for computing initial density.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
sumihar: Agree. Comment is removed.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
sumihar: Yes, thanks, it is important to use the save attribute here; it has been added now.

Changing compute_scales_and_fluxes() to accept Celcius is more invasive. Besides, all the internal formula's in atmospheric_stability module use the assumption of absolute temperature; it would be clear if all temperature variables everywhere in the module are expressed in Kelvin.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
sumihar: I suppose _Q_ here refers to humidity and _sat_ to saturation. There is no hint indeed related to salt in the keyword.

Apparently, it is a common term used in the meteorological community. This keyword was suggested by user (Firmijn). 

It's not definitive yet. We will revisit all these options in https://issuetracker.deltares.nl/browse/UNST-9891.

sumihar (commented):
Hi @FlorisBuwaldaDeltares,

Thanks a lot for the review. I have implemented all the obvious points that clearly improve the codes. There are a few things I would like to discuss with you. Perhaps it would be easier to talk directly?

Many thanks,
Julius

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
FlorisBuwaldaDeltares: this is up to you, do concurrent already instructs the compiler to try to optimize because of independency so I don't think using node_map_1 / 2 will improve much. If you think it's cleaner like this that is ok

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/flowgeom_interpolate.f90
FlorisBuwaldaDeltares: Indeed, the loop over nodes can be done concurrently, however, accessing the nd%ln array is unfortunately very slow in my experience. My gut feeling says that a loop over links is better, but if we want to be sure we will have to benchmark it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
FlorisBuwaldaDeltares: I quite like that idea! but maybe it's a bit too much for the current issue as it is already big. I leave that up to you

FlorisBuwaldaDeltares (approved):
looks good

PR #976, opened 2026-06-16
Run the TestBench.py checks in the new python container

rene-deltares:
…on.kt

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


ci/teamcity/Delft3D/ciUtilities/TestBenchValidation.kt
```diff
@@ -70,23 +66,52 @@ object TestBenchValidation : BuildType({
...
-                workingDir = "test/deltares_testbench"
-                commandArgs = "bake validate"
-            }
+        script {
+            name = "Install dependencies"
+            workingDir = "test/deltares_testbench"
+            scriptContent = """
+                #!/usr/bin/env bash
+                uv venv
+                source .venv/bin/activate
```
harmenwierenga, line 75: This source is redundant

ci/teamcity/Delft3D/ciUtilities/TestBenchValidation.kt
```diff
@@ -70,23 +66,52 @@ object TestBenchValidation : BuildType({
...
+                uv pip sync pip/lnx-dev-requirements.txt
```
harmenwierenga, line 76: I understand now that these requirements were generated from the pyproject.toml. Why did you choose uv pip sync instead of uv sync?

ci/teamcity/Delft3D/ciUtilities/TestBenchValidation.kt
```diff
@@ -70,23 +66,52 @@ object TestBenchValidation : BuildType({
...
+            """.trimIndent()
+            dockerImage = "%docker_image%"
+            dockerImagePlatform = ScriptBuildStep.ImagePlatform.Linux
+            dockerPull = true
+            dockerRunParameters = """
+                --mount type=volume,source=uv-cache-test-bench-validation,destination=/root/.cache/uv
+                --env UV_LINK_MODE=copy
+                --rm
```
harmenwierenga, line 84: I think that teamcity already adds the --rm, but it's fine to leave it

tools/deltares_fortran_styler/uv.lock
harmenwierenga, line 1: Why did you add this now, and why only for the fortran styler?

harmenwierenga (commented):
Nice change!

ci/teamcity/Delft3D/ciUtilities/TestBenchValidation.kt
rene-deltares: That is what I thought as well, but alas with `uv pip sync` (not the same as `uv sync`) it is required to source the environment unfortunately. I thought about making a lock-file for TestBench.py project, but that requires yet more changes, so I want to do that... at some point.

ci/teamcity/Delft3D/ciUtilities/TestBenchValidation.kt
rene-deltares: To install dependencies from the current (requirements.txt based) lock-files, you need `uv pip sync`. `uv sync` will ignore the requirements file, install the dependencies from the `uv.lock` file if present and create a new `uv.lock` file if it is not there. `uv pip sync` is the old "pip-tools" based interface (which is supposed to be a drop-in replacement for pip-tools). We started using `uv pip sync` for TestBench.py before `uv` supported the `uv.lock` file. The `uv.lock` file was introduced later.

tools/deltares_fortran_styler/uv.lock
rene-deltares: To me, especially for CI builds, it is better to lock the dependencies, rather than relying on loosey-goosey version numbers (although the ones in fortran-styler's `pyproject.toml` are pretty strict, which is good). Since the fortran-styler did not have any sort of lock-file yet, I just added it.

If you're wondering why I didn't switch from pip-tools lock-file to the `uv.lock` file in TestBench.py: I did consider it, tried it. But I ran into some problems that required me to change the versions of some packages. Also, I would need to keep it in sync with the old requirements.txt lock files... Unless we completely switch over to `uv`, which also requires me to modify documentation and let people know, etc... I want to do all of this, but best to leave that for another issue. There's more stuff I want to do at some point, like publishing TestBench.py as a package to nexus.

harmenwierenga (approved):

PR #977, opened 2026-06-16
UNST-9968 Do not check config in build.py anymore, rely on cmake chec…

harmenwierenga:
…king. Fix one last occurence of old windows 2022 server container

# Issue link UNST-9968


rene-deltares (approved):
Looks good

PR #978, opened 2026-06-16
Fm/task/unst 10003 generic location reader

FlorisBuwaldaDeltares:
# What was done 

- extract generic location file reader from sourcesink location input reader
- reuse generic location reader for bubblescreens and long culverts
- bubblescreens no longer read extfile twice, but store all relevant info into the bubblescreen objects on the first pass.
- extracted md_unstruc_model_data to avoid cyclic dependencies

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_unstruc_model_data.f90
FlorisBuwaldaDeltares, line 1: not much here to review, pure move from unstruc_model to data, to avoid cyclic dependencies from modules that only need md_ settings

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -135,6 +136,10 @@ subroutine convertLongCulvertsAsNetwork(structurefile, jaKeepExisting, culvertpr
       character(len=IdLen) :: temppath, tempname, tempext
       logical :: write_converted_files_
       integer :: longculvertindex
+      real(kind=dp), allocatable :: x_coordinates(:), y_coordinates(:), z_coordinates(:)
+      real(kind=dp), allocatable :: xpl(:), ypl(:), zpl(:)
```
KoenBussemakerDeltares: You are no longer using the reapol module (at least not directly), but these `xpl ypl zpl` do seem to hint towards that, and could even cause conflicts if somebody choses to use the reapol module here. I would prefer more meaningful names that make sense in within the function context. Some like `all_x_coordinates all_y_coordinates all_z_coordinates` would already make things more understandable.

Same applies to `npl` by the way.

FlorisBuwaldaDeltares: you're right, will fix

KoenBussemakerDeltares (approved):
Nice that you were able to reuse all of this logic!

PR #982, opened 2026-06-17
UNST-9995 replace source sink by connected sink source

erwin-mulder:
# What was done 
Added the ConnectedSinksSources class, to encapsulate the collection of sources and sinks data that has to be sent to FM trough preCICE for the upcoming update. This collection is to be filled based on the data read from the NF2FF files produced by C-SUMO.

The encapsulation is a structures of vectors, each of which are sent to FM as quantities on vertices. All these vectors need to have a guaranteed to have the same length. This is achieved by having only a member function that extends all of them at once.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9995


src/tools_gpl/pre_c_sumo/include/connected_sinks_sources.hpp
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @brief Connected Sinks and Sources administration
+     */
+    class ConnectedSinkSources
+    {
+    public:
+        // constructor with coordinates etc?
+
+        // add an entry
+        void add_entry(double sink_x, double sink_y, double sink_z_bottom, double sink_z_top, double source_x,
```
AabhasSrivastava, line 33: I am not sureif it counts as nitpicking, but I would rther have meaningful variable names. That is more with respect to Discharge rate and Momentum.

AabhasSrivastava: Also, the header for function shoud contain doxygen style description, which could also make clear what q and u are.

src/tools_gpl/pre_c_sumo/src/connected_sinks_sources.cpp
```diff
@@ -0,0 +1,60 @@
...
+        sink_x_vector.push_back(sink_x);
+        sink_y_vector.push_back(sink_y);
+        sink_z_bottom_vector.push_back(sink_z_bottom);
+        sink_z_top_vector.push_back(sink_z_top);
+        source_x_vector.push_back(source_x);
+        source_y_vector.push_back(source_y);
+        source_z_bottom_vector.push_back(source_z_bottom);
+        source_z_top_vector.push_back(source_z_top);
+        q_vector.push_back(q);
+        u_magnitude_vector.push_back(u_magnitude);
```
AabhasSrivastava: Aren't momentum values optional? What is the error handling if the source/sink does not have momentum with the push back?

AabhasSrivastava: Can we add unit test here to test this? WIth/without momentum terms?

erwin-mulder: There are no optional things in this structure. If (for some reason) a quantity is not relevant, it is simply zero.
In fact that is what this class is all about: You are **forced** to provide some number for each item. We can make additional wrapper functions if needed that just take less input values that calls the main one with the 'missing' values zeroed out if that makes more sense at the call site, but that remains to be seen during further integration.

erwin-mulder: I added a a basic unittest for adding and clearing.

src/tools_gpl/pre_c_sumo/include/connected_sinks_sources.hpp
erwin-mulder: Added comments.

AabhasSrivastava (approved):
Looks good

PR #984, opened 2026-06-17
All/task/devopsdsc 965 consolidate smoke test xml

aslowwriter:
# What was done 

We have some smoke tests that run a small set of very fast models on H7 with apptainer to make sure things like the environment are setup properly. THese tests have not been running regularly and so they might break of their own accord making triage harder when it is done during something like a release.  To solve this we have included a `dimr_container_smoke_test.xml` which defines the set of models we wish to run for these purposes. This xml is included by a dimr config that is run alongside the regular test bench so that we can ensure it remains green. This way if the smoke test fails we know it is due to an environment error. 

There are quite some old tests that have been removed. This set is kept intentionally small, but we can relatively easily add new tests to it in the future if we want. 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

[DEVOPSDSC-965](https://issuetracker.deltares.nl/browse/DEVOPSDSC-965)


tribbin (approved):
As discussed. Good work!

robin-deltares (approved):
This time an approval from my Deltares account.

ahmad-el-sayed (approved):

PR #986, opened 2026-06-18
DEVOPSDSC-985:  Abort TC build cancellation GitHub workflow when the trigger build config state is unknown

ahmad-el-sayed:
# What was done 

Workflow fails with error if the state the Trigger build config is unknown.

# Issue link
DEVOPSDSC-985


ci/github/cancel_teamcity_builds.sh
harmenwierenga, line 183: The water quality builds are triggered separately, and not as part of 'Trigger.' Are those also cancelled by this action or not?

harmenwierenga (approved):
This seems good enough indeed, it is very helpful that it will actually show up as red now

ci/github/cancel_teamcity_builds.sh
ahmad-el-sayed: If they are in a different project, and depend on any build configs cancelled in the current project, they will likely fail due to cancelled dependency(ies). I have not tested that. 

The property `affectedProject` is used in the locator of the URLs.  This should find configurations connected via snapshot/artifact dependencies inside and even outside the project.

ahmad-el-sayed: If this does not happen, please report the details and we can gladly look into it..

PR #989, opened 2026-06-19
UNST-9314 support friction type and value defaults for long culverts

FlorisBuwaldaDeltares:
… that verifies

# What was done 

- see title
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
@@ -90,7 +90,7 @@ contains
    subroutine convertLongCulvertsAsNetwork(structurefile, jaKeepExisting, culvertprefix, structures_output, crsdef_output, ierr, crsdeffile, write_converted_files)
       use dfm_error, only: dfm_noerr, dfm_wronginput
       use m_missing, only: dmiss
-      use m_Roughness, only: frictiontypestringtointeger
+      use m_Roughness, only: frictiontypestringtointeger, frictionTypeIntegerToString
       use m_readstructures, only: allowedflowdirtoint, get_value_or_addto_forcinglist
       use messagehandling, only: mess, level_error, msgbuf, err_flush, setmessage
       use properties, only: tree_create, prop_inifile, tree_create_node, tree_put_data, node_value, tree_num_nodes, tree_get_name, prop_get, prop_set, tree_remove_child_by_name, prop_write_inifile, tree_destroy
@@ -105,6 +105,7 @@ contains
       use m_CrossSections, only: fill_hashtable
       use m_alloc, only: realloc
       use m_read_location_info, only: read_polyline_coordinates
+      use m_physcoef, only: IFRCTYPUNI, FRCUNI1D
 
       implicit none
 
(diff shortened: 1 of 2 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1371,128 +1373,128 @@ subroutine find1d2dculvertlinks(network, longculvert, numcoords)
 
       integer :: ierror
 
-         longculvert%flowlinks = 0
-         jafounds = 0 ! Found the starting node or not
-         jafounde = 0 ! Found the ending node or not
-         is = 1 ! the starting node of the polyline
-         ie = numcoords ! the ending node of the polyline
+      longculvert%flowlinks = 0
```
rene-deltares, line 1376: Everything from this point onward is just removing an indent, right? The diff makes this a bit hard to see

FlorisBuwaldaDeltares: yeah I'd like the auto-fprettify :/ but you can turn of whitespace changes in the diff viewer!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
```diff
@@ -1019,6 +1019,73 @@ subroutine test_2pt_friction_converted() bind(C)
...
+      call create_mdu_file(mdu_file, NET_FILE, STR_FILE)
+      call convertlongculverts(mdu_file, STR_FILE, NET_FILE)
+      call init_two_culvert_scenario(mdu_file, iresult)
+      call f90_assert_eq(iresult, DFM_NOERR, to_c_string("model init must succeed"))
+      call f90_assert_eq(longculverts(1)%numlinks, 1, to_c_string("2-point culvert should have 1 link"))
+      call f90_assert_eq(iresult, DFM_NOERR, to_c_string("model init must succeed"))
+      call f90_assert_eq(nlongculverts, 2, to_c_string("two long culverts should be registered"))
+
+      do i = 1, 4
+         call flow_spatietimestep()
```
rene-deltares, line 1077: Not that it matters much in runtime, because we're writing and reading files anyway. But is this unit test not doing too much? We want to know if the default friction type and value are loaded when they're not in the structure.ini, right? But I guess checking that the discharge also works

rene-deltares (approved):
Looks good, but just a few questions

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
rene-deltares: Ah, good tip!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
FlorisBuwaldaDeltares: I like this test more, if you just check whether they're registered, unfortunately due to the mess that is the flow1d parsing and applying you really never know what you're going to get..

jjnoort (approved):

PR #990, opened 2026-06-22
DEVOPSDSC-961: Improve verschilanalyse task scheduling

ahmad-el-sayed:
# What was done 

`srun` is invoked to download the model input and verschillentool json configs from MinIO.  A `16vcpu_spot` partition is requested. SLURM tries to obtain resources immediately. If the requested partition is not available `srun` hangs and waits until a partition is available. Until then no resources are allocated. TeacmCity build configs are configured to time out after 60 mins. If  `srun` hangs longer, the `Submit` build config fails and the Verschilanalyse is aborted.

`sbatch` is used instead and stricter dependencies are set in the steps of the submit script. 
 
# Issue link
DEVOPSDSC-961


robin-deltares (approved):
Looks nice

PR #994, opened 2026-06-23
Remove unit test: TEST(FF2NFWriterTest, DISABLED_WriteToFile)

robin-deltares:
It is blocking the release pipeline.


harmenwierenga (changes requested):
Please do not remove this, it is explicitly ignored, so nothing is wrong. This test only exists to make the creation of the file writer easier, you can run it manually locally to see what happens. Please make a change to accept ignored tests.

PR #996, opened 2026-06-23
sedmor branch: minor flow improvements

hrajagers:
# What was done 

`zlin`: move `dp` inclusion to module level.
`getvanrijnwci`: define all constants in `dp` instead of keeping a few `fp` ones (which are currently equivalent to `dp` within FM)
`setbobs`: update various variables only when needed. 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/setbobs.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/setbobs.f90
@@ -177,21 +177,20 @@ contains
                cycle ! skip update of bobs for structures
             end if
 
-            n1 = ln(1, L)
-            n2 = ln(2, L) ! flow ref
-            k1 = lncn(1, L)
-            k2 = lncn(2, L) ! net  ref
-            zn1 = zk(k1)
-            if (zn1 == dmiss) then
-               zn1 = zkuni
-            end if
-            zn2 = zk(k2)
-            if (zn2 == dmiss) then
-               zn2 = zkuni
-            end if
-
             if (kcu(L) == 1) then ! 1D link
 
+               n1 = ln(1, L)
+               n2 = ln(2, L) ! flow ref
+               k1 = lncn(1, L)
+               k2 = lncn(2, L) ! net  ref
+               zn1 = zk(k1)
+               if (zn1 == dmiss) then
+                  zn1 = zkuni
+               end if
+               zn2 = zk(k2)
+               if (zn2 == dmiss) then
+                  zn2 = zkuni
+               end if
                if (ibedlevtyp == BEDLEV_TYPE_WATERLEVEL .or. ibedlevtyp == BEDLEV_TYPE_WATERLEVEL6) then ! tegeldieptes celcentra ! TODO: [TRUNKMERGE] WO/BJ: do we need stm_included in this if (consistent?)
                   if (stm_included) then
                      bl1 = bl(n1)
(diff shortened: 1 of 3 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/setbobs.f90
```diff
@@ -177,21 +177,20 @@ subroutine setbobs() ! and set blu, weigthed depth at u point
...
-            k2 = lncn(2, L) ! net  ref
-            zn1 = zk(k1)
-            if (zn1 == dmiss) then
-               zn1 = zkuni
-            end if
-            zn2 = zk(k2)
-            if (zn2 == dmiss) then
-               zn2 = zkuni
-            end if
-
```
FlorisBuwaldaDeltares, line 192: why do you not want to set bobs for 1d2d links here?

FlorisBuwaldaDeltares (commented):
one question

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/setbobs.f90
FlorisBuwaldaDeltares: I guess if the testbench is green, this was dead code.. but still I don't get it, if we're looping over numl1d, it either contains both type 1 links and other links, or it is only type 1 links but then we don't need an if statement to check it

hrajagers: Thanks! The code change was only to move the computations of these quantities inside the if `kcu(L) == 1` clause since they are not used outside that block. As such moving these lines doesn't affect the functionality of the code (hence the green test bench). Having said that you raise a very valid point: what should the code do exactly? That was outside my initial scope of just merging these code improvements, but it's good to reflect on it. You ask about 1d2d links, I'm starting to wonder how many links in the range `L = 1, lnx1D` would have `kcu(L) /= 1`, i.e. what are exactly the exceptions being made. I will check this and reply by tomorrow morning the latest.

FlorisBuwaldaDeltares: ok! in the meantime i'll approve, i'll leave it up to you if you want with merging until figuring out whether this code can be improved or not

FlorisBuwaldaDeltares (approved):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/setbobs.f90
hrajagers: OK, I checked. `setbobs` contains two loops over the `L = 1, lnx1D` range. This first one only processes the `kcu(L) == 1` case which are the 1D links. The second loop immediately below it addresses the cases `kcu(L) == 3` (lateral), `5` (street inlet), `7` (roof gutter) and the `kcu(L) == BEDLEV_TYPE_MIN` case, which corresponds to `4` (longitudinal). So, in the end all 1D2D cases seem to be handled, just separately. From morphology perspective we still have to check the code for the 1D2D links; those have been out of scope for us so far.

Overall conclusion, all code is okay upon first glance ... except for the comparison `kcu(L) == BEDLEV_TYPE_MIN` which is nonsense. Although it's tempting to replace `BEDLEV_TYPE_MIN` now by 4, I will not revert to this magic number. I have created UNST-10053 to replace all magic numbers for `kcu` and I will merge this change as is.

PR #1001, opened 2026-06-24
UNST-10051 Prevent HDF5 init check from causing errors

KoenBussemakerDeltares:
# What was done 

The recent conan merge bumped HDF5 (dependency of NetCDF) to version 1.14.6, where previously we used 1.14.2. Between those versions some runtime checks were added to HDF5 which cause NetCDF initialization issues, and can cause model runs to fail. This happens only in Debug.

The fix is to disable floating point exceptions for the intel compiler. Unfortunately this means that it is disabled throughout our code, which is really unfortunate. More info: https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-3/[fpe](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-3/fpe.html).html

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


KoenBussemakerDeltares:
This was mainly a test to see if this would fix the issue. It does but the fix is far from ideal. Will be fixed by #1009 instead.

PR #1002, opened 2026-06-24
All/bugfix/unst 10035 python install race condition

harmenwierenga:
# What was done
Install python through UV to remove the race condition in the python install and subsequent pip install from the test-environment docker container on windows. Perform the package install in the teamcity test.kt configuration, and use a fixed tagged container with just a python installation. We create a volume mount for the uv cache, so that each agent only has to download these packages once.

It did not work to place the uv cache in the image, because the UV_INDEX_URL is not standard but is not forwarded to the docker build step, so there is a mismatch in the caches of the image and the run in the container.

# Issue link UNST-10035


ci/teamcity/Delft3D/windows/test.kt
```diff
@@ -125,15 +125,25 @@ object WindowsTest : BuildType({
...
 
             dockerImage = "containers.deltares.nl/delft3d-dev/test/delft3d-test-environment-windows:%container.tag%"
             dockerImagePlatform = ScriptBuildStep.ImagePlatform.Windows
             dockerPull = true
-            dockerRunParameters = "--memory %teamcity.agent.hardware.memorySizeMb%m --cpus %teamcity.agent.hardware.cpuCount%"
+            dockerRunParameters = """
+                --memory %teamcity.agent.hardware.memorySizeMb%m
+                --cpus %teamcity.agent.hardware.cpuCount%
+                --env UV_LINK_MODE=copy
+                --volume test-environment-uv-cache:C:\uv\cache
```
rene-deltares, line 147: Nice, I hope to see a build where we can see that the caching is effective here, and we see that the uv install is indeed using the cache.

ci/teamcity/Delft3D/windows/testEnvironment.kt
```diff
@@ -33,23 +33,6 @@ object WindowsTestEnvironment : BuildType({
     }
 
     steps {
-        powerShell {
```
rene-deltares, line 36: P-drive begone!

ci/teamcity/Delft3D/windows/test.kt
```diff
@@ -125,15 +125,25 @@ object WindowsTest : BuildType({
                         set argsList=%%argsList%% --copy-failed-cases
                     )
 
-                    python TestBench.py %%argsList%%
+                    rem Fresh venv for this branch; wheels come from the mounted uv cache volume
+                    uv venv
+                    if errorlevel 1 exit /b 1
+                    uv pip sync pip/win-requirements.txt
```
rene-deltares, line 134: I wonder if you should move the `uv pip sync` after the `call .venv/Scripts/activate`

rene-deltares: Let's check the logs... I guess if the dependencies are *not* installed in the venv then doing `python TestBench.py` will fail because modules can't be found. And then we can fix it by moving activate before the uv pip sync

rene-deltares: If it works, then forget I set anything

rene-deltares (commented):
Minor thing about the order of 'activate' and 'uv pip sync'. Let's await the pipeline so we can see what's going on

ci/teamcity/Delft3D/windows/test.kt
harmenwierenga: I have changed that yesterday, so it's done

ci/teamcity/Delft3D/windows/test.kt
harmenwierenga: image
[This build](https://dpcbuild.deltares.nl/buildConfiguration/Delft3D_WindowsTest_virtual_Batch_1886145855_1/7445463?showLog=7445463_163_139&logView=flowAware) never has 'Downloading...' in the uv log, so it resolves everything from the conan cache 😄

rene-deltares (approved):
I love it

PR #1005, opened 2026-06-25
UNST-10056 Create script that injects the host user into apptainer, a…

harmenwierenga:
…ttempt to prevent missing user errors

# Issue link UNST-10056


harmenwierenga:
This is not a solution. It turned out that there are three agents where the user had disappeared (whoami failed)

PR #1008, opened 2026-06-25
UNST-9887: improvement 1 obtained from testing

sumihar:
# What was done 

- Replace mdu keyword StabilityFunction by StabilityFunction**s** 
- Introduce info for when the iteration in atmospheric_stability module does not converge
- Improve heatun(): assign values to variables forced_latent_heat_flux, forced_sensible_heat_flux, free_convective_sensible_heat_flux, and free_convective_latent_heat_flux as they are needed in output files too

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9887


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
```diff
@@ -240,7 +241,11 @@ pure function compute_scaling_parameters(wind_velocity_x, wind_velocity_y, air_t
             exit
          end if
       end do
-
+      
+      if (iteration == MAXIMUM_ITERATION) then
+         call mess(LEVEL_INFO, 'AirSeaInteractionModel MOST: iteration does not converge.')
```
FlorisBuwaldaDeltares: iteration does not converge is not really the right message here, as we have an algorithm that converges across multiple iterations. I would instead say: no convergence after maximum number of iterations, or just 'no convergence'

FlorisBuwaldaDeltares: also, you write this as level INFO but at what point is the convergence bad enough that we abort the simulation? it's also not helpful for users to hear this, as what are they going to do next?  I think we should either abort on non-convergence, or give the user some info on what they can do to make sure this doesn't happen.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -344,11 +344,14 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
...
-         else
-            total_heat_flux = forced_latent_heat_flux + forced_sensible_heat_flux + longwave_radiation_flux + free_convective_sensible_heat_flux + free_convective_latent_heat_flux
+            forced_latent_heat_flux = latent_heat_flux(n)
+            forced_sensible_heat_flux = sensible_heat_flux(n)
+            free_convective_sensible_heat_flux = 0.0_dp
+            free_convective_latent_heat_flux = 0.0_dp
          end if
          
+         total_heat_flux = forced_latent_heat_flux + forced_sensible_heat_flux + longwave_radiation_flux + free_convective_sensible_heat_flux + free_convective_latent_heat_flux
+
```
FlorisBuwaldaDeltares, line 354: much cleaner!

FlorisBuwaldaDeltares (approved):
tiny comment, nothing major

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/atmospheric_stability.f90
sumihar: The INFO was intended to give insight that may be useful to analyze what happened, now we are still in development phase, we are still exploring ways to understand what going on is. For example to analyse why a simulation crashed (the first test simulation by Firmijn crashed).

But in its current form, I think it is not really useful indeed. It may be more useful if info about simulation time is added. But it will require a lot of modifications. 

I have discussed this with Firmijn. We agreed to focus first on using output files for analyzing possible issues. Moreover, in another JIRA issue, extra variables will be added to output file. They may give better insights than an INFO message.

Hence, the INFO has been removed here.

sumihar:
> tiny comment, nothing major

Hi @FlorisBuwaldaDeltares,
thanks for the review. I have made a small modification.

FlorisBuwaldaDeltares (approved):

PR #1010, opened 2026-06-26
UNST-9695 Replace removed msvc extension std::locale::empty reference…

harmenwierenga:
…s in FBC

# Issue link UNST-9695


FlorisBuwaldaDeltares (approved):
nice

PR #1012, opened 2026-06-29
UNST-9999 Add error logging on input errors, and unit tests to check if they work

FlorisBuwaldaDeltares:
# What was done 

- vibe coded some checks for various user input errors during sealock initialize
- vibe coded some unit tests to check whether the errors trigger and are correct

# Evidence of the work done 

image

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dsle/packages/src/sealock.c
+++ b/src/engines_gpl/dsle/packages/src/sealock.c
@@ -63,6 +63,9 @@ int sealock_init(sealock_state_t *lock, time_t start_time, unsigned int max_num_
     // Initialize parameters consistent with current and given settings.
     status = dsle_initialize_state(&lock->parameters, &lock->phase_state,
                                    lock->phase_state.salinity_lock, lock->phase_state.head_lock);
+    if (status != SEALOCK_OK) {
+      log_error("Lock '%s': failed to initialize state: %s!\n", lock->id, dsle_error_msg(status));
+    }
   }
 
   if (status == SEALOCK_OK) {
@@ -91,15 +94,24 @@ int sealock_load_timeseries(sealock_state_t *lock, char *filepath) {
         load_phase_wise_timeseries(&lock->timeseries_data, filepath) ? SEALOCK_ERROR : SEALOCK_OK;
     break;
   default:
+    log_error("Lock '%s': unknown computation_mode %d while loading '%s'!\n", lock->id,
+              lock->computation_mode, filepath);
     status = SEALOCK_ERROR;
     break;
+  }
+  if (status != SEALOCK_OK) {
+    log_error("Lock '%s': failed to read operational parameters file '%s'!\n", lock->id, filepath);
+    return status;
   }
 
   if (status == SEALOCK_OK) {
     lock->current_row = NO_CURRENT_ROW;
     num_rows = get_csv_num_rows(&lock->timeseries_data);
-    if (!num_rows)
+    if (!num_rows) {
+      log_error("Lock '%s': operational parameters file '%s' contains no data rows!\n", lock->id,
+                filepath);
       return SEALOCK_ERROR;
+    }
     time_column = malloc(num_rows * sizeof(double));
     if (time_column != NULL) {
       status = get_csv_column_data(&lock->timeseries_data, "time", time_column, num_rows);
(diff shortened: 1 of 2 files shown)


src/engines_gpl/dsle/packages/src/sealock.c
KoenBussemakerDeltares, line 79: I find this `if (status == SEALOCK_OK)` control flow very silly and error prone. Something like this would be cleaner:

```
status = <do thing>
if (status != OK) return status;

status = <do another thing>
if (status != OK) return status;
```

You didn't write this so it's fine if you consider this out of scope, but it would be a nice improvement IMO.

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -91,15 +94,24 @@ int sealock_load_timeseries(sealock_state_t *lock, char *filepath) {
...
+  if (status != SEALOCK_OK) {
+    log_error("Lock '%s': failed to read operational parameters file '%s'.\n", lock->id, filepath);
+    return status;
+  }
 
   if (status == SEALOCK_OK) {
     lock->current_row = NO_CURRENT_ROW;
     num_rows = get_csv_num_rows(&lock->timeseries_data);
-    if (!num_rows)
+    if (!num_rows) {
```
KoenBussemakerDeltares: The boolean operator is confusing for integers. Use `if (num_rows != 0) ` in this case

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -109,15 +121,29 @@ int sealock_load_timeseries(sealock_state_t *lock, char *filepath) {
           if (times_strictly_increasing(lock->times, num_rows)) {
             lock->times_len = num_rows;
           } else {
+            // Report the first non-increasing record.
+            for (size_t i = 1; i < num_rows; i++) {
```
KoenBussemakerDeltares: Use an int instead of size_t, especially since you are using i - 1 on line 126. If you ever change something and the size_t underflows then you get a giant number, which can lead to very annoying bugs. I speak from experience.

src/engines_gpl/dsle/packages/src/sealock.c
```diff
@@ -322,7 +351,12 @@ static int sealock_update_phase_wise_parameters(sealock_state_t *lock, time_t ti
   // Update when we enter a new phase, or don't know the time_step yet (at start).
   if (sealock_update_current_row(lock, time)) {
     status = get_csv_row_data(&lock->timeseries_data, lock->current_row, &row_data);
-    if (status == SEALOCK_OK) {
+    if (status != SEALOCK_OK) {
+      log_error("Lock '%s': failed to read row %zu from '%s'.\n", lock->id, lock->current_row + 1,
+                lock->operational_parameters_file);
+      return status;
+    }
+    {
```
KoenBussemakerDeltares: Looks like you are creating an additional scope here, probably accidentally

src/engines_gpl/dsle/packages/src/sealock.c
FlorisBuwaldaDeltares: let's do this another time, this whole status thing is across the entire codebase so it's not an insignificant effort

hrajagers: Although I agree with the remark, I can also imagine that using a very generic preprocessor macro name such as `OK` or `ERROR` may cause clashes. Maybe the prefix `SEALOCK_` was used to avoid that.

src/engines_gpl/dsle/packages/tests/test_sealock.c
```diff
@@ -15,6 +18,41 @@ void setUp(void) {}
 
 void tearDown(void) {}
 
+// Write the given contents to a fresh CSV file alongside the running test
+// (rather than relying on a pre-generated fixture being found on the path).
+// Returns the path; caller must remove(path) when done.
+static const char *write_temp_csv(const char *filename, const char *contents) {
```
KoenBussemakerDeltares: There's nothing temp about this csv file unless you delete it afterwards. And that is handled at the call site (or not, which would be bad). So it shouldn't really be part of this function name.

KoenBussemakerDeltares (approved):

PR #1013, opened 2026-06-29
UNST-9277: Support multiple external forcing files

thomas-pijls:
# What was done 

- Cleaned up the `init_new()` subroutine in `fm_external_forcings_init.f90` by putting many blocks in subroutines
- Modified the `initialize_bubblescreens()` subroutine to handle multiple extforce files. For this to work, a new `allocate_bubblescreens_array()` subroutine was created to initialize the bubblesceen arrays to the correct size before the filling (`initialize_bubblesceens()`) takes place.
- All new extforce files and the inifieldfile are put in one array when the mdu is readout.
- `init_new()` now loops over the extforce+inifieldfile list
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
       use m_deprecation, only: check_file_tree_for_deprecated_keywords
       use fm_deprecated_keywords, only: deprecated_ext_keywords
       use dfm_error, only: DFM_NOERR, DFM_WRONGINPUT
       use m_alloc, only: realloc
       use unstruc_messages, only: threshold_abort
-      use m_source_sink, only: source_sinks
 
-      character(len=*), intent(in) :: external_force_file_name !< file name for new external forcing boundary blocks
+      ! Parameters
       integer, intent(inout) :: iresult !< integer error code. Intent(inout) to preserve earlier errors.
```
FlorisBuwaldaDeltares: why did you add this comment? iresult is not a parameter

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+
+      call init_registered_items()
+
+      ! First cycle, validate all external forcing files and add their contents to the bnd_ptrs list.
+      do i_ext = 1, size(extfile_new_list)
+
+         call validate_and_open_external_forcing_file(trim(extfile_new_list(i_ext)), bnd_ptrs(i_ext)%node_ptr, major(i_ext), iresult)
+         if (iresult /= DFM_NOERR) then
+            cycle ! Skip this external forcing file if it could not be validated or opened.
+         end if
```
FlorisBuwaldaDeltares, line 114: it is strange to cycle here if the forcing file is not validated, but we continue? undoubtedly things will go wrong in the later loops when the routine will still attempt to loop over this invalid file? I think you should just abort with an error here or take the file out of the list

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+      num_bubblescreens = 0
+
+      allocate (bnd_ptrs(size(extfile_new_list)))
+      allocate (major(size(extfile_new_list)))
+      allocate (file_names(size(extfile_new_list)))
+      allocate (base_dirs(size(extfile_new_list)))
+
+      call init_registered_items()
+
+      ! First cycle, validate all external forcing files and add their contents to the bnd_ptrs list.
```
FlorisBuwaldaDeltares: I find cycle a bit of an odd choice of words here, what is cyclical about this? a cycle implies that it keeps going indefinitely.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+         bnd_ptr => null()
          return
       end if
 
-      res = .true.
+      ! Inifieldfile is deprecated, issue a warning if the file name ends with '.ini'.
+      if (file_name(len_trim(file_name) - 3:) == '.ini') then
+         write (msgbuf, '(a)') 'The inifieldfile is deprecated. Consider moving the content of '//trim(file_name)//' to the external forcings file.'
+         call warn_flush()
+      end if
```
FlorisBuwaldaDeltares: this warning was removed on main, as the deprecation warning should be emitted when reading the .mdu file

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -96,106 +284,104 @@ module subroutine init_new(external_force_file_name, iresult)
...
          write (msgbuf, '(a,i0,".",i2.2,a,i0,".",i2.2,a)') 'Unsupported format of new external forcing file detected in ''' &
             //file_name//''': v', major, minor, '. Current format: v', ExtfileNewMajorVersion, ExtfileNewMinorVersion, &
             '. Ignoring this file.'
+
          call err_flush()
+
          iresult = DFM_WRONGINPUT
          return
+
       end if
```
FlorisBuwaldaDeltares, line 283: why add all these extra line breaks?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -96,106 +284,104 @@ module subroutine init_new(external_force_file_name, iresult)
...
-      ! Allocate lateral provider array now, just once, because otherwise realloc's in the loop would destroy target arrays in ecInstance.
-      maxlatsg = tree_count_nodes_byname(bnd_ptr, 'lateral')
-      if (maxlatsg > 0) then
-         call realloc(balat, maxlatsg, keepExisting=.false., fill=0.0_dp)
-         call realloc(qplat, [max(1, kmx), maxlatsg], keepExisting=.false., fill=0.0_dp)
-         call realloc(lat_ids, maxlatsg, keepExisting=.false.)
-         call realloc(n1latsg, maxlatsg, keepExisting=.false., fill=0)
-         call realloc(n2latsg, maxlatsg, keepExisting=.false., fill=0)
-      end if
+   end subroutine build_itpenzr_and_itpenur
```
FlorisBuwaldaDeltares: i think if you're going to the trouble of putting this in an extra routine, you can also rename itpenzr and itpenur  as i have no clue what that is supposed to mean

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -96,106 +284,104 @@ module subroutine init_new(external_force_file_name, iresult)
...
 
-      call split_filename(file_name, base_dir, fnam) ! Remember base dir of input file, to resolve all refenced files below w.r.t. that base dir.
+   !> Builds temporary reverse lookup tables that map boundary block # in file -> boundary condition nr in openbndsect (separate for u and z).
+   subroutine build_itpenzr_and_itpenur(itpenzr, itpenur, num_items_in_file)
+      use fm_external_forcings_data, only: nbndz, itpenz, nbndu, itpenu
 
-      num_items_in_file = tree_num_nodes(bnd_ptr)
+      ! Parameters
+      integer, dimension(:), allocatable, intent(out) :: itpenzr
+      integer, dimension(:), allocatable, intent(out) :: itpenur
```
FlorisBuwaldaDeltares: missing apidocs

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -96,106 +284,104 @@ module subroutine init_new(external_force_file_name, iresult)
...
-         case ('spatial', 'meteo', 'parameter', 'initial')
-            res = res .and. init_spatial_fields(block_ptr, base_dir, file_name, group_name)
+      ! Allocate source-sink related arrays now, just once, because otherwise realloc's in the loop would destroy target arrays in ecInstance.
+      if (num_source_sinks > 0) then
+         call source_sinks%initialize(num_source_sinks)
+      end if
 
-         case ('sourcesink')
-            res = res .and. init_sourcesink_forcings(block_ptr, base_dir, file_name, group_name)
+   end subroutine allocate_laterals_and_source_sinks
```
FlorisBuwaldaDeltares: this routine is only called once, why did you decide to put it in a separate routine?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -208,26 +394,20 @@ module subroutine init_new(external_force_file_name, iresult)
...
 
-      call tree_destroy(bnd_ptr)
-
-      if (res) then
-         iresult = DFM_NOERR
-      else
-         iresult = DFM_WRONGINPUT
-      end if
-   end subroutine init_new
+   end subroutine finalize_lateral_forcings
```
FlorisBuwaldaDeltares: if you're going to make this a routine, it should use proper input and output arguments and not use globals

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -1435,13 +1615,12 @@ subroutine initialize_bubblescreens(bnd_ptr, base_dir, file_name, num_bubblescre
       type(tree_data), pointer, intent(in) :: bnd_ptr !< tree of extForceBnd-file's [boundary] blocks
       character(len=*), intent(in) :: base_dir !< Base directory of the ext file
       character(len=*), intent(in) :: file_name !< Name of the ext file, only used in error messages, actual data is read from block_ptr
+      integer, intent(inout) :: i_bubblescreen !< Loop index for bubblescreens within the .ext file.
```
FlorisBuwaldaDeltares: why is this an ) argument? because if it is supposed to continue from the index of the previous ext file, it means this comment is now wrong

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+         deallocate (bubblescreen_air_discharge)
+      end if
+
+      ! Allocate the bubblescreens and bubblescreen_air_discharge arrays based on the number of bubblescreens
+      allocate (bubblescreens(num_bubblescreens))
+      allocate (bubblescreen_air_discharge(num_bubblescreens))
+
+   end subroutine allocate_bubblescreens_array
+
+   !> Validates the external forcing file and opens it, returning a pointer to the tree of boundary blocks.
```
FlorisBuwaldaDeltares: I think it is important to mention this routine absolutely does not validate the external forcing file, there can be total crap inside and this routine will happily return success. it only validates the version number

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+         if (.not. associated(bnd_ptr)) then
+            cycle ! Skip this external forcing file if it could not be validated or opened.
+         end if
+
+         num_laterals = num_laterals + tree_count_nodes_byname(bnd_ptr, 'lateral')
+         num_source_sinks = num_source_sinks + tree_count_nodes_byname(bnd_ptr, 'sourcesink')
+
+         call initialize_bubblescreens(bnd_ptr, base_dirs(i_ext), file_names(i_ext), i_bubblescreen, bubblescreen_source_sinks)
+         num_source_sinks = num_source_sinks + bubblescreen_source_sinks
+      end do
```
FlorisBuwaldaDeltares: why are bubblescreens and laterals and sourcesinks not all counted in a single loop? also the comment says nothing about why bubblescreens are initialized in this loop. Also, the name "initialize_bubblescreens" implies it is only called once, but you call it for every file. It should then be renamed to "initialize_bubblescreens_in_extfile" or something

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,243 @@
...
+
+      res = res .and. add_bubblescreen_source_sinks()
+
+      ! Fourth cycle, read all external forcing files and initialize the boundary, lateral, spatial, and source-sink forcings.
+      do i_ext = 1, size(extfile_new_list)
+         bnd_ptr => bnd_ptrs(i_ext)%node_ptr
+
+         if (.not. associated(bnd_ptr)) then
+            cycle ! Skip this external forcing file if it could not be validated or opened.
+         end if
```
FlorisBuwaldaDeltares: i think this is unnecessary if we're going to stop with an error if an extfile could not be validated or opened

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1700,6 +1700,28 @@ subroutine readMDUFile(filename, istat)
...
+         if (len_trim(md_inifieldfile) > 0) then
+            call realloc(extfile_new_list, size(extfile_new_list) + 1, fill=' ', keepExisting=.true.)
+            extfile_new_list(size(extfile_new_list)) = md_inifieldfile
+         end if
+      else
+         if (len_trim(md_inifieldfile) > 0) then
+            allocate(extfile_new_list(1))
+            extfile_new_list(1) = md_inifieldfile
+         else
+            ! If no new external forcing files or initial field file is specified, allocate an empty list to avoid issues later on
```
FlorisBuwaldaDeltares: what issues are we avoiding?

FlorisBuwaldaDeltares (changes requested):
functionally everything seems fine, but some parts of the new code need some cleanup

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: It is right, it's the argument of the subroutine?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -1636,13 +1580,30 @@ subroutine initialize_bubblescreens(bnd_ptr, base_dir, file_name, i_bubblescreen
...
+         if (str_tolower(group_name) == 'bubblescreen') then
+            num_bubblescreens = num_bubblescreens + 1
+         end if
+      end do
+
+      ! Reallocate the bubblescreens array to hold all bubblescreen blocks found in the external forcings file
+      if (allocated(bubblescreens)) then
+         num_bubblescreens = num_bubblescreens + size(bubblescreens)
+      end if
+      call reallocate_bubblescreens_array(num_bubblescreens)
```
FlorisBuwaldaDeltares: to make this less confusing, you could consider num_bubblescreens => num_bubblescreens_in_file and a separate variable total_num_bubblescreens

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -1686,7 +1647,34 @@ subroutine initialize_bubblescreens(bnd_ptr, base_dir, file_name, i_bubblescreen
       call realloc(nodeCountBubbleScreen, size(bubblescreens), fill=0)
       nNodesBubbleScreen = 0
 
-   end subroutine initialize_bubblescreens
+   end subroutine initialize_bubblescreens_in_extfile
+
+   !> Reallocates the bubblescreens and bubblescreen_air_discharge arrays to a new size, preserving existing data.
```
FlorisBuwaldaDeltares, line 1657: you could perhaps make a comment saying that this routine cannot be called after set_bubblescreen_source_sinks because doing so would break ec module connections, just so that in 5 years nobody makes the same mistake again

FlorisBuwaldaDeltares (approved):
thanks for the changes! two tiny comments because I see you have a red pipeline anyway, up to you if you want to do something with it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
FlorisBuwaldaDeltares: we left the name but at least there's a prober description now :)

FlorisBuwaldaDeltares (approved):
nice :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1700,6 +1700,28 @@ subroutine readMDUFile(filename, istat)
...
+         end if
+      else
+         if (len_trim(md_inifieldfile) > 0) then
+            allocate(extfile_new_list(1))
+            extfile_new_list(1) = md_inifieldfile
+         else
+            ! If no new external forcing files or initial field file is specified, allocate an empty list so functions can still 'loop' over this list.
+            allocate(extfile_new_list(0))
+         end if
+      end if
```
arthurvd, line 1724: Here's a simpler version without unnecessary nested IF.
It keeps all the cumbersome (de)allocate where they are needed (around strsplit), but the inifieldfile appendnig now is simpler:

```suggestion
      if (len_trim(md_extfile_new) > 0) then        
         call strsplit(md_extfile_new, 1, extfile_new_list, 1)
      end if
      if (.not. allocated(extfile_new_list)) then
         ! If no new external forcing files were specified, allocate an empty list so functions can still 'loop' over this list.
         allocate(extfile_new_list(0))
      end if

      ! IniFieldFile is treated entirely by ExtForceFileNew code (during deprecation phase)
      if (len_trim(md_inifieldfile) > 0) then
         call realloc(extfile_new_list, size(extfile_new_list) + 1, fill=' ', keepExisting=.true.)
         extfile_new_list(size(extfile_new_list)) = md_inifieldfile
      end if
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
```diff
@@ -535,16 +535,21 @@ subroutine findexternalboundarypoints() ! find external boundary points
...
+
+      do i_ext = 1, size(extfile_new_list)
+         if (len(trim(extfile_new_list(i_ext))) > 0) then
+            inquire (file=trim(extfile_new_list(i_ext)), exist=jawel)
+
+            if (jawel) then
+               ext_force_bnd_used = .true.
+            else
+               call qnerror('Boundary external forcing file '''//trim(extfile_new_list(i_ext))//''' not found.', '  ', ' ')
+               write (msgbuf, '(a,a,a)') 'Boundary external forcing file ''', trim(extfile_new_list(i_ext)), ''' not found.'
```
arthurvd: I'd revert this change: "External forcing file" without "Boundary".

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
-      character(len=*), intent(in) :: external_force_file_name !< file name for new external forcing boundary blocks
+      ! Arguments
       integer, intent(inout) :: iresult !< integer error code. Intent(inout) to preserve earlier errors.
 
+      ! Local variables
       integer :: initial_threshold_abort
       logical :: res
-      logical :: is_successful
       type(tree_data), pointer :: bnd_ptr !< tree of extForceBnd-file's [boundary] blocks
+      type(tree_data_ptr), dimension(:), allocatable :: bnd_ptrs !< array of pointers to extForceBnd-file's [boundary] blocks
```
arthurvd:
```suggestion
      type(tree_data_ptr), dimension(:), allocatable :: bnd_ptrs !< array of pointers to extForceBnd-file's [boundary] blocks, one tree for each *.ext file.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+      integer, dimension(:), allocatable :: itpenzr !< Reverse lookup table that maps boundary block number in external forcings file to boundary condition number in openbndsect for z. {size=num_items_in_file}
+      integer, dimension(:), allocatable :: itpenur !< Reverse lookup table that maps boundary block number in external forcings file to boundary condition number in openbndsect for u. {size=num_items_in_file}
+
+      integer :: num_laterals !< Total number of laterals in all external forcing files
+      integer :: num_source_sinks !< Total number of source-sinks in all external forcing files
+      integer :: bubblescreen_source_sinks !< Number of source-sinks in bubblescreen
+
+      character(len=256), dimension(:), allocatable :: file_names !< List of file names
+      character(len=256), dimension(:), allocatable :: base_dirs !< List of base directories
+      integer, dimension(:), allocatable :: major !< Major version numbers of the external forcing files
```
arthurvd, line 86: I like all the docs 👍 It is good practice to use regular `!` for local code/var comments and only use Doxygen-style comments `!>`, `!<` and `!!` for module/function documentation and their dummy arguments.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+      integer :: ib, ibqh
+      integer, dimension(:), allocatable :: itpenzr !< Reverse lookup table that maps boundary block number in external forcings file to boundary condition number in openbndsect for z. {size=num_items_in_file}
+      integer, dimension(:), allocatable :: itpenur !< Reverse lookup table that maps boundary block number in external forcings file to boundary condition number in openbndsect for u. {size=num_items_in_file}
+
+      integer :: num_laterals !< Total number of laterals in all external forcing files
+      integer :: num_source_sinks !< Total number of source-sinks in all external forcing files
+      integer :: bubblescreen_source_sinks !< Number of source-sinks in bubblescreen
+
+      character(len=256), dimension(:), allocatable :: file_names !< List of file names
+      character(len=256), dimension(:), allocatable :: base_dirs !< List of base directories
```
arthurvd: Consider using `max_prop_length` from module `properties` instead of 256. (I know many existing code has hardcoding too, but: boy scout rule).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+      integer, dimension(:), allocatable :: major !< Major version numbers of the external forcing files
 
+      ! Initialization
       iresult = DFM_NOERR
+      num_laterals = 0
+      num_source_sinks = 0
+
+      allocate (bnd_ptrs(size(extfile_new_list)))
```
arthurvd, line 93: Maybe now is a good moment to rename all "bnd_ptr" variables to "ext_ptr": it is not about boundaries specifically anymore. Also see (old) docstrings on lines 65/66: it is *not* about [Boundary] blocks, but the entire *.ext file with all its blocks.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+
+            return
+         end if
+
+         file_names(i_ext) = trim(extfile_new_list(i_ext))
+         call split_filename(file_names(i_ext), base_dirs(i_ext), fnam)
+
+      end do
+
+      ! Third loop, count laterals and sourcesink blocks, including bubblescreen source-sinks. Then allocate the lateral and source-sink arrays.
```
arthurvd: I don't see any second loop ;-) third/fourth -> second/third?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+      i_bubblescreen = 0
+      do i_ext = 1, size(extfile_new_list)
+         bnd_ptr => bnd_ptrs(i_ext)%node_ptr
+
+         num_laterals = num_laterals + tree_count_nodes_byname(bnd_ptr, 'lateral')
+         num_source_sinks = num_source_sinks + tree_count_nodes_byname(bnd_ptr, 'sourcesink')
+
+         call initialize_bubblescreens_in_extfile(bnd_ptr, base_dirs(i_ext), file_names(i_ext), i_bubblescreen, bubblescreen_source_sinks)
```
arthurvd, line 129: Note to ourselves (you don't need to change this in this PR): it's a pity that bubblescreens are an exception and are initialized in/right after this loop already, whereas all the others (boundary/lateral/sourcesinks, etc.)  are initialized in the last loop below.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+
+         if (allocated(itpenzr)) then
+            deallocate (itpenzr)
+         end if
+
+         if (allocated(itpenur)) then
+            deallocate (itpenur)
+         end if
+
+      end do
```
arthurvd: Sometimes it improves readibility when you add comment about which loop this is ending (coding guidelines 5.2):

```suggestion
      end do ! i_ext
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -38,55 +38,217 @@
...
+
+      if (res) then
+         iresult = DFM_NOERR
+      else
+         iresult = DFM_WRONGINPUT
+      end if
+
+   end subroutine init_new
+
+   !> Checks the version number of the external forcing file and opens it, returning a pointer to the tree of boundary blocks.
```
arthurvd:
```suggestion
   !> Checks the version number of the external forcing file and opens it, returning a pointer to the tree of external forcings file blocks.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -96,106 +258,77 @@ module subroutine init_new(external_force_file_name, iresult)
...
          return
+
       end if
 
-      call init_registered_items()
+   end subroutine check_version_number_and_open_external_forcing_file
 
-      call split_filename(file_name, base_dir, fnam) ! Remember base dir of input file, to resolve all refenced files below w.r.t. that base dir.
+   !> Builds temporary reverse lookup tables that map boundary block number in external forcings file to boundary condition number in openbndsect (separate for u and z).
+   subroutine build_itpenzr_and_itpenur(itpenzr, itpenur, num_items_in_file)
```
arthurvd: There is something subtly wrong now with the `itpenzr`/`itpenur` building *and* usages. You're using per-file counters `ib` and `num_items_in_file`, but the starting point is `itpenz`/`itpenu`, which contain boundary segment counters that continue numbering across all extforcefile's.

Some explanation (done for waterlevel/z-type, but the same applies to velocity/u-type boundaries):
* Each open [Boundary] block is added in the open boundary segment administration, see `addopenbndsection()`.
* A single [Boundary] block may open up multiple net links, and for each of those netlinks, we remember the open boundary segment that they are part of. This is the array `itpenz(1:nbndz)`.
  This is done inside `processexternalboundarypoints()`. Existing code, nothing wrong there.
  Note: this array contains open boundary segment numbers which are counted across *all* loaded extforcefile's, so **no counter per ext file**.
* In the creation of the EC-module timespacerelation for each boundary, inside `init_boundary_forcings()`, you sometimes need to know for the current boundary block, what the boundary netlink position is in the global `zbndz` array, see the code `target_index = itpenzr(ib)`.
* The current construction of `itpenzr` is almost correct, but it should not check on `ib <= num_items_in_file`, probably `ib <= nopenbndsect` is correct.
* The `ib` counter in `init_new()` gets reset to 0 for each file, and gets passed as input to `init_boundary_forcings()`, where it is used to lookup inside the `itpenzr()`, which is expecting "global" segment numbers, across **all** ext files. Probably moving `ib = 0` to outside the file loop should help, but do have a closer look yourself.
* Actually, I feel that `ib` can sometimes be different than the open boundary segment number, but let's leave this single remark for outside of this PR.

arthurvd (changes requested):
Hi @thomas-pijls, here's my review. One important fundamental issue, all other comments are minor remarks.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
thomas-pijls: Nice simplification, I applied it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
thomas-pijls: Done ✅

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Done ✅

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Personally I also like to use Doxygen-style docs for local variables. When working in the function/subroutine it helps me a lot in understanding what the local variables mean while working there

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Done ✅

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: You're right, there once was a second loop, however I was able te remove it. Forgot to update the naming, done now ✅

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Good remark. Sven is also a big fan of this, I will keep it in mind in the future for long and/or nested loops

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Done ✅

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
thomas-pijls: Done, made and tested your suggested changes

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
arthurvd: Just for me to understand, what is not working well if you would document your local variables, but with `! comment` instead of `!< comment`? Is your IDE doing something special with them?

arthurvd (approved):
Thanks for the second round, all looks good now.
I'll approve now, but can you only still check my question on:
https://github.com/Deltares/Delft3D/pull/1013#discussion_r3627834004

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
arthurvd: Either overlooked or postponed. Let's leave this for another volunteer in a future PR.

PR #1015, opened 2026-06-29
UNST-9274: Bugfix and add tests for "ininudge"-functionality

JulienGroenenboom:
# What was done 
- bugfix: **First**, set_external_forcings() to do nudging at TStart (so temperature and salinity data are available from the external file). **Next**, call initialize_salinity_and_temperature_with_nudge_variables
- Add three testcases
 
# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated: e02_f045_c001, e02_f045_c002, e02_f045_c003
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9274

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
@@ -278,6 +278,12 @@ contains
          call restore_au_q1_3D_for_1st_history_record()
       end if
 
+      call set_external_forcings(tstart_user, INITIALIZATION_PHASE, error)
+      if (is_error_at_any_processor(error)) then
+         call qnerror('Error occurred when setting external forcings.', ' ', ' ')
+         return
+      end if
+
       call initialize_salinity_and_temperature_with_nudge_variables()
 
       if (jasal > OFF) then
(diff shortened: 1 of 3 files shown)


FlorisBuwaldaDeltares (approved):
as mentioned, we could see if maybe initialize_salinity_and_temperature_with_nudge_variables can be part of the INITIALIZATION_PHASE of set_external_forcings. But for me the code is also fine as-is.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
```diff
@@ -278,6 +278,12 @@ integer function flow_flowinit() result(error)
          call restore_au_q1_3D_for_1st_history_record()
       end if
 
+      call set_external_forcings(tstart_user, INITIALIZATION_PHASE, error)
+      if (is_error_at_any_processor(error)) then
+         call qnerror('Error occurred when setting external forcings.', ' ', ' ')
+         return
+      end if
+
```
sumihar, line 286: At the moment, constituents(itemp,:) should be available before calling set_external_forcings(), where surface temperature is required for calling the atmospheric stability module. 

As a workaround, we can extend get_surface_temperature() with INITIALIZATION_PHASE as input; in that case, surface temperature should be read from variable tem1 instead of from constituents(itemp,:).

For now perhaps good to revert this file to before PR #970? The follow order would be clearer that way. Now a call to set_external_forcings() is inserted in between processes of temperature and salinity, which may not be so clear why.

In general, we need to discuss whether the initialization follow order is correct for temperature. The current workflow is (1) set tem1 by values temini, (2) if rst is available, set tem1 by values from rst file and then set constituents(itemp,:) by tem1, (3) replace/update tem1 by 3D expansion, boundary and nudging, (4) set constituents(itemp,:) by tem1. Perhaps if rst is available, then step (3) and (4) are not needed.

PR #1017, opened 2026-06-29
Fm/task/unst 10096 nvfortran poc

FlorisBuwaldaDeltares:
# What was done 

- create NVHPC toolkit-based third party libs container
- disabled PETSC (out of scope for POC)
- create devcontainer based on this image
- iteratively fix all nvhpc compilation errors
- iteratively fix all unit tests due to hidden bugs nvhpc exposed

# Evidence of the work done 
all unit tests green!
image

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-10096


conan/config/profiles/delft3d_ubuntu22_nvhpc
```diff
@@ -0,0 +1,14 @@
+[settings]
+arch=x86_64
+compiler=gcc
+compiler.cppstd=20
+compiler.libcxx=libstdc++11
+compiler.version=11
+os=Linux
+os.distro=Ubuntu22
```
harmenwierenga, line 8: Why ubuntu?

run_conan.py
```diff
@@ -16,7 +16,7 @@
 from typing import Any, Iterator
 
 WINDOWS_PROFILE = "delft3d_windows_msvc_194"
-LINUX_PROFILE = "delft3d_alma8_intel_2024"
+LINUX_PROFILE = os.environ.get("DELFT3D_CONAN_PROFILE", "delft3d_alma8_intel_2024")
```
harmenwierenga, line 19: We need to find a nice method for changing the profiles still, and they need to be forwarded by build.py as well after it has detected the OS and compiler. I think that an environment variable is not very suitable for passing it on, but fine for this PoC

harmenwierenga (changes requested):
Nice that it compiles as a PoC! The fact that some third party code had to be updated or excluded does show that it is not a light decision to add support for nvfortran. I think that an architect-level discussion should happen on the next steps. This PoC can inform that decision. I have the following questions that I think would be informative:
1) Does nvfortran provide its own MPI implementation? We can probably use the GCC or intel compiled third-party dependencies as long as they don't use MPI, but PETSc and PreCICE do require the same MPI lib as the rest of our software, or we risk runtime issues.
2) What is the state of the nvfortran compiler in your view after working on this? Fortran tooling can be pretty bad, and seeing the changes it makes me a bit worried about the maturity and support of this compiler.
3) What is preventing PETSc compilation with the nvidia compilers?
4) Which C and C++ compilers are you using? The NVIDIA HPC C++ compiler support for C++23 seems a bit less strong, and apparently C++26 support is lacking [https://cppreference.com/cpp/compiler_support/23]

conan/config/profiles/delft3d_ubuntu22_nvhpc
FlorisBuwaldaDeltares: That is the only OS distribution Nvidia currently distributes its nvhpc container for
https://catalog.ngc.nvidia.com/orgs/nvidia/-/containers/nvhpc/-?_lr=1

PR #1018, opened 2026-06-30
UNST-10022: reuse pump capacity from rst

mihai-pricope:
# What was done 

current_discharge seems only used for output, but discharge is what is used for computation. 
So initialized this also from RST file

# Evidence of the work done 

- [ ]	Video/figures \
- [x ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x ]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-10022

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
@@ -18215,6 +18215,7 @@ contains
                   istru = network%sts%pumpIndices(i)
                   pstru => network%sts%struct(istru)
                   pstru%pump%current_capacity = tmpvar(i)
+                  pstru%pump%capacity = tmpvar(i)
                end do
             end if
 
--- a/test/deltares_testbench/configs/dimr/dimr_dflowfm_full_scale_models_win64.xml
+++ b/test/deltares_testbench/configs/dimr/dimr_dflowfm_full_scale_models_win64.xml
@@ -259,7 +259,7 @@
             </checks>
         </testCase>		
         <testCase name="e02_f150_c31_pudong-1d" ref="dimr_default">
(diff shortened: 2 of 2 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -18215,6 +18215,7 @@ subroutine read_structures_from_rst(ncid, filename, it_read)
                   istru = network%sts%pumpIndices(i)
                   pstru => network%sts%struct(istru)
                   pstru%pump%current_capacity = tmpvar(i)
+                  pstru%pump%capacity = tmpvar(i)
```
FlorisBuwaldaDeltares, line 18218: what is the difference between current_capacity and capacity? in which situations are they not the same?? i see why this fixes it but they fact they both exist is confusing to me

FlorisBuwaldaDeltares (approved):
one question, otherwise looks good

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
mihai-pricope: current_capacity seems only used for output, but capacity is what is used for computation. So at output times capacity is copied to current_capacity. 

I would say current_capacity should not be initialized from rst file .. but is harmless and reduces the test splash if it's actually somehow required.

FlorisBuwaldaDeltares: ok, i'd like to remove current_capacity altogether then and just output capacity.. but that's something for another PR i think.

PR #1023, opened 2026-06-30
UNST-9931: Support constituent names in new extfile without delta postfix

rene-deltares:
# What was done 

- Support constituent names in new extfile without delta postfix.
- Added unit test for `get_constituent_name`
- `.bc` files also needed to support the quantity names without the `delta` postfix. Both need to be set consistently: Either both the new ext-file and the bc-file use the 'delta' postfix, or neither of them have it.
- Fixed test case e02_f017_c010 case, which contained invalid quantity names in the '.bc' files. Added a check in the testbench config xml on the his file for a salinity variable.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_utils.f90
```diff
@@ -177,11 +179,19 @@ subroutine get_constituent_name(original_quantity, constituent_name, base_quanti
...
+            return  ! Discharge is not a constituent. Do nothing.
+         end if
+
+         if (strcmpi(original_quantity(index_suffix_start:quantity_length), 'Delta')) then
+            base_quantity = 'sourcesink_constituentDelta'
+            constituent_name = original_quantity(index_prefix_end + 1:index_suffix_start - 1)
+         else
+            base_quantity = 'sourcesink_constituent'
+            constituent_name = original_quantity(index_prefix_end + 1:)
+         end if
```
JulienGroenenboom, line 193: Don't let his thread block your PR (as I'm away for the coming days, so you/others can resolve it), but I was just wondering if it wouldn't be possible to use the same base_quantity = ....  (either with or without suffix Delta) here. In that case you don't have to extend the list of case(... , 'sourcesink_constituent','sourcesink_constituentDelta', ... ) at two other places.

arthurvd: I was thinking the same, too. I guess it can work indeed, as long as the EC relation uses the original name (such that the .bc file keeps the same quantity name in its header as the .ext file).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/helpers/util_helpers.f90
arthurvd, line 1: Could you move this function to `src\utils_lgpl\deltares_common\packages\deltares_common\src\iso_c_utils.f90`
The functions over there are strongly related. (and deltares_common is reachable from your unit tests.

arthurvd (changes requested):
Hi @rene-deltares, see my two comments.

hrajagers:
@rene-deltares Please note you're now changing something inside `deltares_common` while working in an `fm` branch. Although the addition of this new function will probably not fail any other code (unless a function of identical name is defined elsewhere), it would be good to switch to an `all` branch.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_utils.f90
rene-deltares: Yes, that's true. I have options, because I can make it so that `base_quantity` is either always `sourcesink_constituentDelta` or always `sourcesink_constituent`. Then I can remove the other option in those `case(...` lists. The test cases still pass with either.
I think I like `sourcesink_constituent` better, but I think I'll be conservative and keep the old `sourcesink_constituentdelta` to minimize unnecessary changes.

In either case I can still use either `salinityDelta` or `salinity` in the new extfile, and `sourcesink_salinitydelta` or `sourcesink_salinity` in the bc-file. The only constraint is that the name in the extfile and the bc file should be consistent. So e.g. no `salinity` in the ext-file and `sourcesink_salinityDelta` in the bc file.

rene-deltares:
> @rene-deltares Please note you're now changing something inside `deltares_common` while working in an `fm` branch. Although the addition of this new function will probably not fail any other code (unless a function of identical name is defined elsewhere), it would be good to switch to an `all` branch.

Closing this PR, in favor of this one: https://github.com/Deltares/Delft3D/pull/1039

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_utils.f90
arthurvd: I tested this in your new branch, and indeed it works for both with and without the Delta postfix, as long as ext and bc file consistently use the same postfix.

PR #1026, opened 2026-07-01
All/task/unst 10086 oneapi 2026 linux

harmenwierenga:
Add OneAPI 2026 to the options in the Linux buildtools container

# Issue link UNST-10086

### DIFF
--- a/ci/dockerfiles/linux/buildtools.Dockerfile
+++ b/ci/dockerfiles/linux/buildtools.Dockerfile
@@ -6,7 +6,7 @@ ARG BASE_IMAGE_URL=containers.deltares.nl/base_linux_containers/8-base:latest
 
 FROM ${BASE_IMAGE_URL} AS buildtools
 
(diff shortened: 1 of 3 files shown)


hrajagers (approved):
The changes look good at a conceptual level. This isn't used yet in any regular build pipeline, so nothing should fail (the current failures on Windows must be due to other instabilities).

Have you already tried building the system using these settings somewhere?

harmenwierenga:
> The changes look good at a conceptual level. This isn't used yet in any regular build pipeline, so nothing should fail (the current failures on Windows must be due to other instabilities).
> 
> Have you already tried building the system using these settings somewhere?

Thanks Bert. Yes, I have tried to build our software locally. The findings so far: 1) PETSc needs to be updated to the latest version to build with intel 2026 (working on it) 2) There are some C issues in our code that need to be fixed (in progress) 3) The triangle library does not build with OneAPI 2026. We already wanted to get rid of triangle, so this is an extra reason

PR #1030, opened 2026-07-02
prescribed missing value in the .grd value is now taken into account …

marcio-albernaz:
The missing value prescribed in the .grd file is taken into account.

# What was done 

The xymiss variable was loading the missing value, for example -999, but that was not used. Instead an old hard coded value equal to zero was used. Now the code uses xymiss to correctly handle missing values

 

# Evidence of the work done 

- [x]	Video/figures \
image

- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
<e04_wave/f04-general/c08_missing_gridvalue>
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/SWAN-314


src/engines_gpl/wave/packages/data/src/read_grids.f90
```diff
@@ -394,6 +395,7 @@ subroutine read_grd(filnam    ,xb     ,yb   ,codb ,covered, mmax  ,nmax ,sferic
 !
     ! Default value for missing value: zero
     !
+    tol    = 1.0e-6_hp
     xymiss = 0.0_hp
```
hrajagers: If `tol` is always assigned the same value, then it should be defined as a parameter (which should be capitalized):
If `tol` is always assigned the same value, then it should be defined as a parameter (which should be capitalized):
`real(hp), parameter :: TOL = 1.0e-6_hp`

hrajagers: Initially I also made a remark about `xymiss`, but I forgot to first look at the collapsed code.

marcio-albernaz: Thanks, Bert. I was in doubt about how to prescribe TOL. This was actually the initial advice from copilot, but I havent seen any simular example in this code. I now implemented as parameter.

marcio-albernaz:
Testing with the collected binaries shows that the grid containing -999.0 values has been handled correctly. The -999.0 values are also correctly passed to the SWAN input file (INPUT).

image

hrajagers:
@marcio-albernaz Can you try to add a (simple and quick) test case to the test bench?

marcio-albernaz:
@hrajagers test case added

hrajagers:
@marcio-albernaz Thank you for adding the test. I see that the `doc/chapters/case_text/tex` has been copied from case `e04_f04_c04`. Can you update the text such that it describes the purpose and approach of the new test instead of the other test. This will be reflected by an update of the time stamp in the xml-file in this repository.

marcio-albernaz:
Hi @hrajagers, you caught me :)
The documentation in the tex-file is also updated and the testbench/TC is green.

hrajagers (approved):
Thank you @marcio-albernaz !

PR #1031, opened 2026-07-02
UNST-9889_support_relative_wind_for_atmospheric_stability

sumihar:
# What was done 
- create m_relative_wind.f90 containing procedures for computing relative corrected wind velocity components and wind speed 
- implement the procedure in routine compute_air_water_interaction_most_fluxes() in set_external_forcings_update.
- use the procedures in setwindstress() and heatun() 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable; tests have been done locally, with relativewind = 0 and 1, the results are compared to an existing setup generated earlier before relativewind is supported (practically relativewind=0); when relativewind = 0, results are identical with the existing one; when relativewind = 1, results are different; see the two figures below. Further tests will be done by user.

When RelativeWind = 0:
20260702_wldiff_relwind_0 0

When RelativeWind = 1:
20260702_wldiff_relwind_1 0

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9889

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
@@ -1,0 +1,114 @@
+!----- AGPL --------------------------------------------------------------------
+!
+!  Copyright (C)  Stichting Deltares, 2017-2026.
+!
+!  This file is part of Delft3D (D-Flow Flexible Mesh component).
+!
+!  Delft3D is free software: you can redistribute it and/or modify
+!  it under the terms of the GNU Affero General Public License as
+!  published by the Free Software Foundation version 3.
+!
+!  Delft3D  is distributed in the hope that it will be useful,
+!  but WITHOUT ANY WARRANTY; without even the implied warranty of
+!  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+!  GNU Affero General Public License for more details.
+!
+!  You should have received a copy of the GNU Affero General Public License
+!  along with Delft3D.  If not, see <http://www.gnu.org/licenses/>.
+!
+!  contact: delft3d.support@deltares.nl
+!  Stichting Deltares
+!  P.O. Box 177
+!  2600 MH Delft, The Netherlands
+!
+!  All indications and logos of, and references to, "Delft3D",
+!  "D-Flow Flexible Mesh" and "Deltares" are registered trademarks of Stichting
+!  Deltares, and remain the property of Stichting Deltares. All rights reserved.
+!
+!-------------------------------------------------------------------------------
+
+!> Utilities for correcting wind velocity with the relative motion of the surface flow.
+module m_relative_wind
+
+   use precision, only: dp
+
+   implicit none
+
+   private
+
+   interface compute_surface_relative_wind
+      module procedure compute_surface_relative_wind_scalar
+      module procedure compute_surface_relative_wind_links
+   end interface compute_surface_relative_wind
+
+   public :: compute_surface_relative_wind
+   public :: compute_surface_relative_wind_on_link
+
+contains
+
+   !> Correct a single wind vector with the relative surface-flow contribution.
+   !!
+   !! corrected_wind = wind - relativewind * surface_velocity
+   !!
+   !! If `corrected_wind_speed` is provided, this routine also returns
+   !! `sqrt(corrected_wind_x**2 + corrected_wind_y**2)`.
+   subroutine compute_surface_relative_wind_scalar(wind_x, wind_y, relativewind, surface_u, surface_v, corrected_wind_x, corrected_wind_y, corrected_wind_speed)
+      real(kind=dp), intent(in) :: wind_x, wind_y !< Input wind components in global x/y direction [m/s].
+      real(kind=dp), intent(in) :: relativewind !< Weight factor for relative-wind correction [-].
+      real(kind=dp), intent(in) :: surface_u, surface_v !< Surface-flow components in global x/y direction [m/s].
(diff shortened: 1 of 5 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
```diff
@@ -0,0 +1,114 @@
...
+
+contains
+
+   !> Correct a single wind vector with the relative surface-flow contribution.
+   !!
+   !! corrected_wind = wind - relativewind * surface_velocity
+   !!
+   !! If `corrected_wind_speed` is provided, this routine also returns
+   !! `sqrt(corrected_wind_x**2 + corrected_wind_y**2)`.
+   subroutine compute_surface_relative_wind_scalar(wind_x, wind_y, relativewind, surface_u, surface_v, corrected_wind_x, corrected_wind_y, corrected_wind_speed)
```
FlorisBuwaldaDeltares: this could be an elemental subroutine to help the compiler optimize! same for the other two routines

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
```diff
@@ -0,0 +1,114 @@
...
+      real(kind=dp), intent(in) :: wind_x(:), wind_y(:) !< Input wind components on links in global x/y direction [m/s].
+      real(kind=dp), intent(in) :: relativewind !< Weight factor for relative-wind correction [-].
+      real(kind=dp), intent(out) :: corrected_wind_x(:), corrected_wind_y(:) !< Corrected wind components on links in global x/y direction [m/s].
+
+      integer :: L
+
+      do L = 1, lnx
+         call compute_surface_relative_wind_on_link(L, wind_x(L), wind_y(L), relativewind, corrected_wind_x(L), corrected_wind_y(L))
+      end do
+   end subroutine compute_surface_relative_wind_links
```
FlorisBuwaldaDeltares: if compute surface relative wind on links is elemental, you don't need this subroutine, you can call it with array arguments. even if its not elemental, i think it's still better to just write this loop out at the call site? a do loop disguised as a subroutine just makes things confusing in my opinion.

FlorisBuwaldaDeltares: if i'm checking your PR correctly, this routine also doesn't seem to be called at all?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
```diff
@@ -0,0 +1,114 @@
...
+!> Utilities for correcting wind velocity with the relative motion of the surface flow.
+module m_relative_wind
+
+   use precision, only: dp
+
+   implicit none
+
+   private
+
+   interface compute_surface_relative_wind
```
FlorisBuwaldaDeltares: personally I would prefer "compute_wind_relative_to_surface", as if you start reading the name you'd think you are computing a surface ('compute_surface') and then later you have to mentally adjust to actually wind being computed. but if you prefer this name that is OK too.

FlorisBuwaldaDeltares (approved):
nice work, a few suggestions.
I'm wondering why you say you don't add a testcase, isn't this new functionality and you already have a case demonstrating the effect of this new calculation?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
sumihar: Suggestion has been applied.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
sumihar: Suggestion has been applied.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_relative_wind.f90
sumihar: Thank you! By using elemental routine, the codes become simpler and this routine is no longer needed indeed.

Note 1: this routine was used earlier via aliasing. 

Note 2: my first try was indeed to write the loop at the call site. But then I felt that the codes at the call site didn't become simpler than how they were. By moving the loop to the subroutine they became simpler. But you have a point there, that it may be confusing too that way.

sumihar (commented):
Hi Floris,

Many thanks for your input; very useful! I have applied them; also your suggestion in PR 1022 about 'get_surface_temperature(**initialization=**.false.).

Greetings,
Julius

sumihar:
> I'm wondering why you say you don't add a testcase, isn't this new functionality and you already have a case demonstrating the effect of this new calculation?

Hi @FlorisBuwaldaDeltares,
About test case. I did create a test case, to use locally on my PC. The test case has not been put yet in the testbench; hence, I said I didn't add a test case. 

This feature is still rather experimental. I want to avoid having a test case in the test bench with unacceptable results. User (Firmijn) is still busy with testing. Only if the results are okay, then it will be ready to add the test cases to the testbench. 

Many thanks,
Julius

FlorisBuwaldaDeltares (approved):
very nice, thanks for the improvements!

PR #1036, opened 2026-07-06
UNST-9690 Add cygwin to windows docker container. Remove some unneede…

harmenwierenga:
…d components and update uv and conan

- Added cygwin
- Removed includeRecommended from the Visual Studio install, explicitly install those parts that are necessary (reduce size docker container, total from 58.8GB to 50.6GB despite adding cygwin)
- Update uv and conan to latest versions
- Ensure that PROGRA~2 alias is available for C:\Program Files (x86), because cygwin cannot handle spaces in paths
- Remove separate git installer, use git from visual studio

# Issue link UNST-9690

### DIFF
--- a/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
+++ b/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
@@ -87,8 +87,8 @@ RUN Start-Process -FilePath 'C:\cygwin-setup-x86_64.exe' -Wait -NoNewWindow -Arg
     Remove-Item -Force C:\cygwin-setup-x86_64.exe; \
     Remove-Item -Recurse -Force C:\cygwin-packages
 
-# Cygwin's link.exe conflicts with the Intel ifort/ifx compiler. Rename it so the
-# Intel/MSVC linker is picked up instead (per PETSc Windows install instructions).
+# Cygwin's link.exe conflicts with the Intel ifort/ifx compiler. Rename it so the Intel/MSVC linker is
+# picked up instead (see https://petsc.org/main/install/windows/#native-microsoft-intel-windows-compilers).
 RUN if (Test-Path 'C:\cygwin64\bin\link.exe') { Move-Item 'C:\cygwin64\bin\link.exe' 'C:\cygwin64\bin\link-cygwin.exe' }
 
 # Add Cygwin to the PATH (appended so it does not shadow the explicitly installed tools).
--- a/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
+++ b/ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
@@ -10,14 +10,11 @@ ADD vs_Community.exe C:/
 
 RUN (start /w vs_community.exe --passive --wait --noUpdateInstaller --norestart --nocache \
     --installPath "C:\\Program Files\\Microsoft Visual Studio\\17\\Community" \
-    --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended \
+    --add Microsoft.VisualStudio.Workload.NativeDesktop \
+    --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 \
     --add Microsoft.VisualStudio.Component.Windows11SDK.26100 \
     --add Microsoft.VisualStudio.Component.VC.ATLMFC \
     --add Microsoft.VisualStudio.Component.VC.CLI.Support \
-    --remove Microsoft.VisualStudio.Component.Windows10SDK.10240 \
-    --remove Microsoft.VisualStudio.Component.Windows10SDK.10586 \
-    --remove Microsoft.VisualStudio.Component.Windows10SDK.14393 \
-    --remove Microsoft.VisualStudio.Component.Windows81SDK \
     || IF "%ERRORLEVEL%"=="3010" EXIT 0)
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'Continue'; $verbosePreference='Continue';"]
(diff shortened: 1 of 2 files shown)


harmenwierenga:
I am creating a teamcity build of this container [here](https://dpcbuild.deltares.nl/buildConfiguration/Delft3D_WindowsBuildEnvironmentI24/7490509). I will run some tests to check that our software still builds with it

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
```diff
@@ -62,10 +64,38 @@ ENV UV_PYTHON_INSTALL_DIR="C:\Program Files\uv\python" \
...
+    '--no-startmenu', \
+    '--no-desktop', \
+    '--no-admin', \
+    '--upgrade-also', \
+    '--root', 'C:\cygwin64', \
+    '--local-package-dir', 'C:\cygwin-packages', \
+    '--site', 'https://mirrors.kernel.org/sourceware/cygwin/', \
+    '--site', 'https://cygwin.mirror.constant.com/', \
+    '--only-site', \
+    '--packages', 'python3,make,git,cmake'; \
```
rene-deltares, line 86: Wait, is this, in addition to cygwin, installying python3, git and cmake again? don't we already have python3, git and cmake in this image?

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
```diff
@@ -62,10 +64,38 @@ ENV UV_PYTHON_INSTALL_DIR="C:\Program Files\uv\python" \
...
+# Add Cygwin to the PATH (appended so it does not shadow the explicitly installed tools).
+RUN setx /M PATH ($Env:PATH + ';C:\cygwin64\bin')
 
 # Copy the setup script into the container
 COPY set-env-vs2022.cmd C:\\set-env-vs2022.cmd
 
 # Set the default command to run the setup script
-CMD ["cmd", "/S", "/C", "C:\\set-env-vs2022.cmd"]
\ No newline at end of file
+CMD ["cmd", "/S", "/K", "C:\\set-env-vs2022.cmd"]
```
rene-deltares, line 101:
```
/C      Carries out the command specified by string and then terminates
/K      Carries out the command specified by string but remains
/S      Modifies the treatment of string after /C or /K (see below)
```

Not sure I understand, but I'm ok with this. Was there a problem that this change solves?

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
```diff
@@ -62,10 +64,38 @@ ENV UV_PYTHON_INSTALL_DIR="C:\Program Files\uv\python" \
 # Install python and conan
 RUN setx /M PATH \"C:\Program Files\uv\bin;C:\Program Files\uv\install;$env:PATH\"
 RUN uv python install 3.12 --default
-RUN uv tool install 'conan ~= 2.28.1'
+RUN uv tool install 'conan ~= 2.30.0'
+
+ADD setup-x86_64.exe C:\\cygwin-setup-x86_64.exe
```
rene-deltares, line 69: I assume you had to add this installer to the `P:/` drive toolchain folder. Too bad there isn't some way to install it with a package manager or something. I've always wondered if we can use winget in these dockerfiles, for instance

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
```diff
@@ -62,10 +64,38 @@ ENV UV_PYTHON_INSTALL_DIR="C:\Program Files\uv\python" \
...
+    '--local-package-dir', 'C:\cygwin-packages', \
+    '--site', 'https://mirrors.kernel.org/sourceware/cygwin/', \
+    '--site', 'https://cygwin.mirror.constant.com/', \
+    '--only-site', \
+    '--packages', 'python3,make,git,cmake'; \
+    Remove-Item -Force C:\cygwin-setup-x86_64.exe; \
+    Remove-Item -Recurse -Force C:\cygwin-packages
+
+# Cygwin's link.exe conflicts with the Intel ifort/ifx compiler. Rename it so the
+# Intel/MSVC linker is picked up instead (per PETSc Windows install instructions).
```
rene-deltares: For these kind of issues where there is a `magic` workaround, I always like to have a link to the documentation/install instructions, so that people who read this know where it comes from.

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
harmenwierenga: This line is not used on teamcity, because it runs the container with a docker exec that runs a custom cmd script. So this line is just for interactively logging in to the container, but /C instantly returns and exits the container if you try to run it interactively. /K is what we need if you want to load an environment and then continue

ci/dockerfiles/windows/set-env-vs2022.cmd
```diff
@@ -1,3 +1,12 @@
 @echo off
+rem Recreate the 8.3 short names for the space-containing "Program Files" folders.
+rem PETSc's Windows configure (run from Cygwin) uses `cygpath -ms` to obtain a
+rem space-free path to the Intel oneAPI MPI/MKL libraries. Windows containers do
+rem not persist 8.3 metadata changes made to base-image folders at build time, so
+rem the short names must be (re)created at container startup instead.
+fsutil 8dot3name set C: 0 >nul 2>&1
+fsutil file setshortname "C:\Program Files" PROGRA~1 >nul 2>&1
+fsutil file setshortname "C:\Program Files (x86)" PROGRA~2 >nul 2>&1
```
rene-deltares, line 9: Ugh... Is this also in PETSc instructions? Is there also a link from this one. I have no idea what an "8.3 short name" even is.

rene-deltares (commented):
Not sure if it's "request changes" or that I just need some motivation for some of the changes. But otherwise very nice!

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
harmenwierenga: I had contact with Harm and Willem and called ICT about it. They looked into storing them in nexus, but they found out that accessing nexus during a windows docker build bakes the credentials into the layers, thereby leaking them. Their solution was to download them from nexus to the agent, then run a powershell script that runs a separate http server on the teamcity agent just to feed those files into the docker build step on windows.

You could also create that server and make use of the P-drive, because there is the benefit of not baking the intel installer into the layers (saves 2GB out of 60GB). I judged that the complexity does not outweigh the benefit right now.

Nexus vs the P-drive does make it read-only, but it makes you require a password. Since those installers are also used by colleagues who install the software on their developer laptops and by colleagues following the 'how to compile Delft3D' courses, I don't think that it is a great idea to put them behind a password and then require ICT to add each new colleague to an AD group just to access these installers (currently I have been waiting for 5 days to add 3 colleagues to an AD group to access the conan repo). 

In conclusion, I think that it is too hairy to make this change now

ci/dockerfiles/windows/set-env-vs2022.cmd
harmenwierenga: On windows 11 these are turned on by default, try "cd c:\PROGRA~1". Apparently it is not on windows server. They are used to prevent spaces in paths in cygwin.

harmenwierenga: Named after the length restrictions apparently, "8.3 filenames are limited to at most eight characters (after any directory specifier), followed optionally by a [filename extension](https://en.wikipedia.org/wiki/Filename_extension) consisting of a [period](https://en.wikipedia.org/wiki/Full_stop) . and at most three further characters." [wikipedia](https://en.wikipedia.org/wiki/8.3_filename)

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
harmenwierenga: Done

harmenwierenga:
> I am creating a teamcity build of this container [here](https://dpcbuild.deltares.nl/buildConfiguration/Delft3D_WindowsBuildEnvironmentI24/7490509). I will run some tests to check that our software still builds with it

The build is green!

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
harmenwierenga: This is inside cygwin, they are linux equivalents. Some tools require git or python on the cygwin command line. PETSc requires python3 and make, and I included cmake and git to prepare for the cloning and building of other components that we will add to conan later. They are added at the end of the path, so they never shadow native tools

harmenwierenga:
image

The [conanpackages build](https://dpcbuild.deltares.nl/buildConfiguration/Delft3D_WindowsConanPackages/7492154?logView=flowAware) is also green in this container

rene-deltares (approved):
Thanks!

ci/dockerfiles/windows/Dockerfile-dhydro-vs2022-i24
rene-deltares: I meant more like this: https://learn.microsoft.com/en-us/windows/package-manager/winget/
just `winget install cygwin`, or something like that. Similar to how we use `dnf`, not copying installers or using nexus. But I don't know, does Nexus have some kind of 'http' mode, where you can just download stuff (installers) that we host? Then you could do with a RUN statement that does webrequest to download the installer, installs, and then removes it immediately.

But indeed, my comment was also not intended as something we need to fix immediately

PR #1040, opened 2026-07-06
Add PETSc to conan

harmenwierenga:
# What was done 

- Update PETSc from 3.24.5 to 3.25.3 (necessary for oneAPI 2026 compilers, still use old compilers now)
- Create conan recipe for PETSc that works on linux (PETSc has no dependencies itself)
- Update recipe to work on windows. Use Cygwin, and create a separate conan package for that. Follow the official steps to build PETSc with msvc + ifx + intel MPI
- Remove PETSc from linux third-party-libs dockerfile
- Change cmake to use PETSc from Conan
- Add conan setting (globally) to choose a Fortran compiler, so that the Fortran compiler becomes part of the package revision of packages that use the Fortran compiler. Use the setting in the NetCDF-Fortran and PETSc recipes
- Remove the PETSc windows sources
 
# Issue link UNST-9690


conan/config/profiles/delft3d_alma8_intel_2024_v2
rene-deltares, line 1: I expect the `run_conan.py` to fail until I reinstall this config. I mean, it will also fail if you make a new profile in a separate file, but then the error message will be something like: "profile not found", right?

Either way, this warrants a message in Teams so that everyone knows what to expect

conan/config/profiles/delft3d_windows_msvc_194_v2
rene-deltares, line 1: Same comment as the linux profile changes.

conan/config/settings_user.yml
```diff
@@ -4,3 +4,5 @@ compiler:
 os:
     Linux:
         distro: ["Alma8", "Alma9", "Alma10"]
+
+fortran_compiler: [null, ifx, gfortran, nvfortran, crayftn, nagfor]
```
rene-deltares, line 8: Cool that we add this now. I didn't know there was a 'Cray' fortran compiler

doc/compiling_Windows.md
```diff
@@ -113,10 +117,14 @@ From the repository root:
 ```
 python run_conan.py initialize external
 ```
-This installs the same compiler profile and settings.
-You will build all third-party dependencies locally from the recipes in [conan/recipes](../conan/recipes).
-Once built, the packages are cached in `~/.conan2` and reused on
-subsequent builds. You only have to rebuild them when the recipes change.
+This installs the same compiler profile and settings. Please check the conan profile in your local Conan
```
rene-deltares, line 120: Nice

rene-deltares (approved):
Looks good!

conan/config/profiles/delft3d_alma8_intel_2024_v2
harmenwierenga: I will have a look at adding a version number to the profile to generate a clearer error message

rene-deltares (approved):
Thanks!

PR #1041, opened 2026-07-06
UNST-10103: Try to fix flaky horton tests

rene-deltares:
# What was done 

There seems to be a bug with the reusing of the global data in 'm_samples' when a model uses this combination:
- Reading the bed level from a geotiff file
- Has observation stations

The two test cases in UNST-10103 have this combination. Details in the JIRA issue: https://issuetracker.deltares.nl/browse/UNST-10103

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


arthurvd (approved):
This fix makes sense: geotiffs are structured sample sets (rows and columns), and therefore set `mxsam` and `mysam` counters. Because they were not reset to 0 upon deletion of those samples, later UNRELATED code that was triggering bilinear interpolation for rows and columns. Moreover, that interpolation call never should have used mxsam and mysam, because it was passing entirely different x/y arrays.

So, two independent bugs, fixed in one PR. Thanks.

PR #1042, opened 2026-07-07
UNST-10113 long culvert 2d2d output fixes

FlorisBuwaldaDeltares:
…ormat even without network. keepexisting = .false. as contact ids are reconstructed anyway and we don't want to write stale ones.

# What was done 

- permute long culvert link lookup after partitioning
- don't keep stale contact ids
- always write ugrid format netfiles for long culvert models 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/partition_write_domains.f90
```diff
@@ -126,11 +127,13 @@ subroutine partition_write_domains(netfilename, icgsolver, jacells, japolygon, j
          if (ierror /= DFM_NOERR) then
             goto 1234
          end if
-
+         
+         permute_longculvertlinks = .true. !> ugly workaround, lperm should only be used when partitioning but not in a sequential run.
```
rene-deltares: Is it possible that `permute_longculvertlinks` is used somewhere after this code? I'm guessing no, because you make no attempt here to save the old value and restore it back later.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_longculverts_data.f90
```diff
@@ -89,9 +91,18 @@ end function is_2D2D
    elemental subroutine is_2D2D_longculvertlink(L, i)
       integer, intent(in) :: L !< Flowlink number
       integer, intent(out) :: i !< Index of the longculvert in longculverts derived type array
+      integer :: Lorg
+
+      Lorg = L
+      if (allocated(Lperm) .and. permute_longculvertlinks) then
```
rene-deltares: I forget why the links had to be permuted in the first place, so this may be a dumb question. But wouldn't the administration be less complex if after permuting the links we also updated `longculverts(i)%netlinks` with the new link numbers (after permuting)? So we don't need to keep checking here (and who knows which other places) if we did a permutation or not and have special handing for it?

rene-deltares (commented):
I made a few remarks to clear up my own understanding

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/partition_write_domains.f90
FlorisBuwaldaDeltares: yes, permute_longculvertlinks is only there for partitioning at this point, so we do not care about the initial value. I did set it back to false after so that long culvert unit tests don't break if you run a partitioning unit test before them

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_longculverts_data.f90
FlorisBuwaldaDeltares: yes, I suppose that could maybe work. I tried to permute the contacts but that failed. I can give it a try

FlorisBuwaldaDeltares: I have implemented your suggestion

rene-deltares (approved):
Thanks!

PR #1045, opened 2026-07-07
UNST-9690 Fix visual studio 2022 install path in windows buildtools c…

harmenwierenga:
…ontainer

# Issue link UNST-9690


rene-deltares (approved):
Looks good to me

PR #1052, opened 2026-07-08
UNST-10106 Corrected a bunch of invalid values in testbench mdu files

KoenBussemakerDeltares:
# What was done 

In the dflowfm_io development branch now runs validation on each mdu file that the testbench runs. Unlike dflowfm all keywords are always checked. This has uncovered that many mdu files contain invalid values which need to be corrected.

This change fixes:
- Duplicate keyword entries (kept the first entry, dflowfm only reads the first one)
- Invalid decimal values, e.g. "0.2." (additional dot at the end)
- Invalid enum values for Vertadvtypsal 
- Invalid RestartDateTime entries, e.g. "yyyymmdd_hhmmss"

# Issue link


rene-deltares (approved):
No references updated, and nothing changed besides cleaning up MDU setting mistakes. Thanks!

PR #1058, opened 2026-07-10
UNST-10137: Fixed run scripts so example 12 for dimr

erwin-mulder:
# What was done 
Altered bin and lib dirs in run scripts so example 12 will start properly with or without preCICE.

- In .bat scripts:
   - we prevent environment variables to be irreversibly broken by wrapping the bat content in setlocal/endlocal
   - we add additional .. in case we run from a level deeper in the case of preCICE
- In .sh scripts:
   - We use readlink -f to get the absolute paths of the installation for use in PATH and LD_LIBRARY_PATH, so that the current working directory does not affect them. 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10137


examples/dflowfm/12_dflowfm_pre_C_SUMO/run.bat
```diff
@@ -1,10 +1,11 @@
 @ echo off
 
+setlocal
 set usePreCICE=1
 set startFM=1
 set startPreCSUMO=1
 
-set bindir=..\..\..\..\install_fm-suite\bin
+set bindir=..\..\..\install_fm-suite\bin
 set libdir=%bindir%\..\lib
```
hrajagers: Since the introduction of conan, the `lib` folder has been removed on Windows. Maybe get rid of references to this old folder while you're working on this batch file.

hrajagers: Thanks for updating the batch files. However, I see that you've also removed the `lib` folder references on Linux ... afaik it's still used there. @harmenwierenga

examples/dflowfm/12_dflowfm_pre_C_SUMO/run.sh
```diff
@@ -4,10 +4,9 @@ usePreCICE=1
 startFM=1
 startPreCSUMO=1
 
-bindir=../../../../install_fm-suite/bin
-libdir=$bindir/../lib
+bindir=$(readlink -f ../../../install_fm-suite/bin)
```
AabhasSrivastava: There is a mismatch between conan script for installation (contains build-type in folder name).
I do not know what should be the best convention to keep here - ideally, it should be without the build tyupe in the path. Best to discuss it with conan team.

examples/dflowfm/12_dflowfm_pre_C_SUMO/run_parallel.bat
```diff
@@ -25,30 +25,29 @@ del /f /q precice_debug_output.txt
 del /f /q precice-profiling\*.*
 rmdir /s /q precice-run
```
AabhasSrivastava, line 27: These directories would only exist for preCICE run. If running with DIMR, it would give us error.

examples/dflowfm/12_dflowfm_pre_C_SUMO/run_parallel.bat
```diff
@@ -25,30 +25,29 @@ del /f /q precice_debug_output.txt
...
-        call mpiexec -n %NPROC% %bindir%\dflowfm-cli.exe FlowFM.mdu --precice
+        set PATH=..\%bindir%
+        call ..\%bindir%\dflowfm-cli.exe --partition:ndomains=%NPROC%:icgsolver=6 FlowFM.mdu
+        call mpiexec -n %NPROC% ..\%bindir%\dflowfm-cli.exe FlowFM.mdu --precice
         cd ..
     ) else (
         echo Please start D-Flow FM
     )
 ) else (
     call %bindir%\run_dimr.bat
```
AabhasSrivastava: for this part with dimr, partitioning still needs to be added

examples/dflowfm/12_dflowfm_pre_C_SUMO/run.bat
erwin-mulder: I discussed with Harmen and have reverted this change.

examples/dflowfm/12_dflowfm_pre_C_SUMO/run.sh
erwin-mulder: I created an additional variable to set the install directory name seperately. Some example scripts allow you to sort of give it on the command line. There is no real standard here.

examples/dflowfm/12_dflowfm_pre_C_SUMO/run_parallel.bat
erwin-mulder: The also don't exist for the first time you ever run preCICE. So sometimes these directories just don't exist.

examples/dflowfm/12_dflowfm_pre_C_SUMO/run_parallel.bat
erwin-mulder: No. Removed it entirely, because parallel is not supported in DIMR at all.

AabhasSrivastava (approved):
Looks good.

AabhasSrivastava (approved):

PR #1060, opened 2026-07-10
UNST-8022_Adding morfac and morft in his writer

ZetaTam-dl:
# What was done 

- Two additional variables, morfac (average morphological factor) and morft (morphological time), are added into the his writer.
- These two variables originally only available in the map file. Similar routine from the map writer is copied to the his writer.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [X] Tests updated \
<e02/f022/c109_timevarying_morfac_morft> newly added.
- []	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
[UNST-8022](https://issuetracker.deltares.nl/browse/UNST-8022)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
```diff
@@ -765,6 +772,22 @@ subroutine unc_write_his(tim) ! wrihis
...
+      dmorfs = dmorft * 86400.0_dp ! seconds
+      mortime = stmpar%morpar%morft * 86400.0_dp ! seconds*morfac since tstart_user
+      if (stmpar%morpar%hydrt > stmpar%morpar%hydrt0) then
+         moravg = dmorft / (stmpar%morpar%hydrt - stmpar%morpar%hydrt0)
+      else
+         moravg = 0.0_dp
+      end if
+      !
+      if (stmpar%morpar%moroutput%morfac) then
+         ierr = nf90_put_var(ihisfile, id_morfac, moravg, [it_his])
```
hrajagers: Replace `ierr = ` by `call check_netcdf_error(`...`)` just like line 790 below. The same could also be done for the `unc_def_var_nonspatial` call aboves, but that's not done yet.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
```diff
@@ -765,6 +772,22 @@ subroutine unc_write_his(tim) ! wrihis
       if (timon) then
          call timstop(handle_extra(64))
       end if
+      
+      !morfac and morft
+      dmorft = stmpar%morpar%morft - stmpar%morpar%morft0 ! days since morstart
+      dmorfs = dmorft * 86400.0_dp ! seconds
```
hrajagers: `dmorfs` isn't used, so it doesn't have to be defined and assigned. In `unstruc_netcdf` the value `dmorfs` is used for some cumulative transport quantities. The history file doesn't have those quantities.

hrajagers (changes requested):
Two small changes requested. You can also update the branch with a merge of main. The main should be a bit more stable than when you made the branch.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
ZetaTam-dl: revised

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
ZetaTam-dl: revised

ZetaTam-dl:
if statements are added to enable morft and morfac only when sediment transport is enabled.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
```diff
@@ -765,6 +772,23 @@ subroutine unc_write_his(tim) ! wrihis
       if (timon) then
          call timstop(handle_extra(64))
       end if
+      
+      !morfac and morft
+      if (his_write_settings%sed > 0 .and. jased > 0 .and. stm_included) then
```
hrajagers: The `jased > 0` is not strictly needed since `stm_included` is defined as `len_trim(md_sedfile) /= 0 .and. len_trim(md_morfile) /= 0 .and. jased == 4`, so if `stm_included` then the value of `jased` must be equal to 4 and hence larger than 0.

hrajagers (commented):
The code update looks good now (with a minor comment). It would be good to update the references for one morphological model in the test bench such that morphological time and morphological factor are available on the his.nc file and checked by the regression test bench.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
ZetaTam-dl: revised

ZetaTam-dl:
Hey @hrajagers , can you please review again. A new test case e02/f022/c109 is added to check time-varying morfac and the morpho time in his. Hope I didn't miss anything. Thanks!

hrajagers:
Hi @ZetaTam-dl, Everything looking good ... if it hadn't been for the one thing that you already raised yourself: this change adds the average morfac `moravg` to the history file "consistent with the map file"; why the average value and not the instantaneous one?

Looking back at Delft3D-FLOW, I see that we always write (as expected) the instantaneous `morfac`. We can also write the `moravg` value, but we only do so when we also write the average sediment transport rates (when `moroutput%cumavg` is set). That makes sense since the `moravg` is the morfac value that is representative for the averaging period and should be combined with those average sediment transport rates to give the cumulative transports.

I see that the `moravg` writing to the D-Flow FM map file was also introduced at the same time as the average sediment transport output. So, we should do some cleaning up. We should write the regular, instantaneous `morfac` by default and should write the `moravg` value in addition when the `cumavg` transport rates are switched on.

ZetaTam-dl:
Hi @hrajagers, I agree to provide instantaneous `morfac` and `moravg` in the output (both map and his). I think this is more handy. Since this issue UNST-8022 focuses on adding the `morft` in the his file, do you agree that we can close this issue and create a new issue for the instantaneous `morfac`?

hrajagers:
@ZetaTam-dl That's fine. I've created UNST-10169 for adding `morfac`. I'll approve this change.

hrajagers (approved):
Approved with new issue UNST-10169.

PR #1063, opened 2026-07-10
UNST-9690 Fix set env, point the TMP folder to a folder local to the …

harmenwierenga:
…container where executables can be executed

- Added lines pointing to the TMP folder
- Removed some untested debug statements
- Removed the visual studio environment sourcing, which is no longer necessary now that the install path environment variable is properly set in the dockerfile so that Intel already loads it

# Issue link UNST-9690


FlorisBuwaldaDeltares (approved):
since the visual studio build also heavily uses the TEMP/TMP folders, and since they are heavily scanned by the virusscanner on the host device, I think this will incidentally also speed up the windows builds. 

One question, is there a virusscanner inside the docker container? if so it would be nice to make sure c:'\build-temp is excluded

harmenwierenga:
> since the visual studio build also heavily uses the TEMP/TMP folders, and since they are heavily scanned by the virusscanner on the host device, I think this will incidentally also speed up the windows builds.
> 
> One question, is there a virusscanner inside the docker container? if so it would be nice to make sure c:'\build-temp is excluded

Thank you! I did check that today actually, and no, windows defender is not supported inside of containers. Since we use hyperv isolation of the windows containers, the layers are also never scanned by ESET's host agent scanner

PR #1064, opened 2026-07-11
ncWriteLatLon should write latitude and longitude for model in known projected coordinate system

hrajagers:
# What was done 

- Include `proj.db` during installation in the `share/proj` folder.
- Update the processing of the `unc_writeopts` flags: make sure that the flags are passed to all routines that needs it. 
- Replaced `iand(unc_writeopts,UG_WRITE_LATLON) == UG_WRITE_LATLON` by `iand(unc_writeopts,UG_WRITE_LATLON) /= 0` checks ... initially because I thought that the original check was wrong when `unc_writeopts` had also other bits set (currently the only other bit used is `UG_WRITE_LYRVAR`) ... but later realizing that both are equivalent for single bit patterns such as `UG_WRITE_LATLON` (1) and `UG_WRITE_LYRVAR` (2) based on general readability considerations for a simple bit check.
- Check whether `crs%proj_string` is known after reading the net-file. If not, switch off the `UG_WRITE_LATLON` bit (i.e. switch off `ncWriteLatLon` if set) to avoid writing latitude and longitude variables only containing missing values.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
e02_f012_c112_ncWriteLatLon added
References updated for 3 WAQ coupling cases: after these changes the PROJ4_string attributes filled.
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link

Closes UNST-10141


hrajagers:
@KoenBussemakerDeltares This is the pull request that I was referring to. Can you check whether the CMake adjustment to include proj.db is good as is?

hrajagers:
@harmenwierenga Since you looked at the Conan sidr of PROJ before, you may also want to have a look.

hrajagers:
@arthurvd Can you look at the code side changes? You are most familiar with the history of coordinate support in net-files.

src/cmake/CMakeLists.txt
```diff
@@ -388,12 +386,6 @@ if(${configuration_type} STREQUAL ${testbench-tc_configuration})
     include(configurations/testbench/waq_configuration.cmake)
 endif()
 
-# QUICKPLOT configuration
-if(${configuration_type} STREQUAL ${testbench-qp_configuration})
-    message(STATUS "Generating solution for QUICKPLOT testbench ...")
-    include(configurations/testbench/waq_configuration.cmake)
```
harmenwierenga, line 406: I think that this qp-configuration (at first I thought that it meant quad precision...) exists to make it possible to merge pull requests on github. GitHub's teamcity integration is not very sophisticated, it requires us to specify some configurations that are required. If other configurations are red, they are shown on github, but you can still merge the PR. So the windows and linux testbench are marked as required, but that means that every branch has to start them, otherwise you can never merge the PR. I think that this quickplot configuration runs the waq tests because they are quick.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
```diff
@@ -210,7 +210,7 @@ subroutine unc_write_his(tim) ! wrihis
 
       ! Only add auto-tranformed lat/lon coordinates if model is Cartesian and user has requested extra latlon output.
 #ifdef HAVE_PROJ
-      add_latlon = jsferic == 0 .and. iand(unc_writeopts, UG_WRITE_LATLON) == UG_WRITE_LATLON
+      add_latlon = jsferic == 0 .and. iand(unc_writeopts, UG_WRITE_LATLON) > 0
```
harmenwierenga: I guess that this is a bit clearer, but they mean the same thing right? Or does UG_WRITE_LATLON have more than one bit set?

And I thought that Fortran does not have unsigned integers, how is this interpreted? Should the check be /= 0?

harmenwierenga (commented):
Nice that the proj.db is added!

src/cmake/CMakeLists.txt
hrajagers: Yes, it was introduced to enable QUICKPLOT pull requests which had a separate pipeline. However, QUICKPLOT code was moved to https://github.com/Deltares/QUICKPLOT some time ago. As main QUICKPLOT developer, I'm sure that this is really not needed anymore.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
hrajagers: Hm ... you're correct. I noticed that the `UG_WRITE_LATLON` bit (1) although set didn't end up writing results. At that point I must have made the wrong conclusion that the result of the `iand` could have more bits set that `UG_WRITE_LATLON` alone ... but I think that some missing `unc_writeopts` arguments was the real problem.

Good catch with respect to the sign. Indeed Fortran integers are signed. Even though we currently don't have flags larger than 2 for this setting, we might have in the future (or for some other setting) so for consistency it should indeed be `/= 0` or `== UG_WRITE_LATLON`. Copilot suggests to use `iand(opts,single_bit_mask) /= 0` if you're checking individual bits, and to use `iand(opts, mask) == mask` if the `mask` may have multiple bits set.

Do you agree that I change it to `/= 0` for these checks on `UG_WRITE_LATLON`?

src/cmake/CMakeLists.txt
```diff
@@ -418,3 +410,9 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/../../bsd-2.txt
               ${CMAKE_CURRENT_SOURCE_DIR}/../../gpl-3.0.txt
               ${CMAKE_CURRENT_SOURCE_DIR}/../../lgpl-2.1.txt
               DESTINATION share)
+
+# Install proj.db only if PROJ library is found and being used
+if(TARGET PROJ::proj)
```
KoenBussemakerDeltares, line 427: All conan packages are found using find_package commands in the root cmake. Proj should therefore always be available no matter what flavor of config you provide to build.py. You can remove the if.

hrajagers: @KoenBussemakerDeltares That raises an interesting question: for a `d3d4` build, the code itself doesn't have a dependency on PETSc, PROJ, and some other libraries now moved to Conan. Deltares developers may not care about this difference since prebuilt versions of those packages are always available, but if you're outside Deltares building from scratch it can make a significant difference in total build time. Do you indeed intend to imply that the current CMake/Conan configuration not distinguish between those cases ... and building PETSc and PROJ is always needed anyway? If so, has it been considered or would it be possible to tweak the configuration such that it only builds those packages required for your configuration (maybe `all` as default)?

hrajagers:
Converted to draft since latest minor code change failed on the added test case. Analyzing ...

src/cmake/CMakeLists.txt
hrajagers: To be continued after September 1st: UNST-10217.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unc_write_his.F90
hrajagers: Code adjusted.

src/cmake/CMakeLists.txt
FlorisBuwaldaDeltares: I was under the impression we would only require packages for the selected configuration. It would be nice if open source users don't have to compile petsc locally for a d3d4 build

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -11244,6 +11244,43 @@ end subroutine read_mesh2d_face_z
 !> Reads the net data from a NetCDF file.
 !! Processing is done elsewhere.
    subroutine unc_read_net(filename, numk_keep, numl_keep, numk_read, numl_read, ierr)
```
FlorisBuwaldaDeltares, line 11246: git diff is being weird and it's hard for me to descipher what is the intended change here.. is it a simple rename of unc_read_net to unc_read_net_old for the non-UGRID reader??

FlorisBuwaldaDeltares (approved):
looks good to me, however the MR description mentioned UG_WRITE_LYRVAR, which I don't see anywhere in the changelist. I do see many == UG_WRITE_LATLON turned into > 0 instead. Is that the change mentioned? why was it necessary?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
hrajagers: `unc_read_net` did two things: (1) try `unc_read_net_ugrid` and if successful return, and (2) if not successful, read it as an old file. I wanted to add a proj string check at the end of both, so I decided to move (2) to a separate routine called `unc_read_net_old`.

PR #1072, opened 2026-07-13
UNST-10147: skip universalWeir on dry links to prevent NaNs in fu/ru.

arthurvd:
# What was done 

- Prevent NaNs in fu/ru caused by `computeUniversalWeir()`, which would only do early return on "negative" waterdepths, and still do its calculations on exactly dry flow links.
- Check waterdepth using the recommended `equal()` routine instead of deprecated `comparereal()`.
- p.s. A simple `<=` worked well too, but to comply with coding guidelines I used the `equal()` function.
- Issue UNST-10147 contains full diagnosis.
 
# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
Closes UNST-10147.

### DIFF
--- a/src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
+++ b/src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
@@ -94,15 +94,16 @@ module m_Universal_Weir
       ! Global variables
       !
       type(t_uni_weir), pointer, intent(in)       :: uniweir  !< Universal weir object.
-      double precision, intent(  out)             :: fum      !< FU.
-      double precision, intent(  out)             :: rum      !< RU.
+      double precision, intent(inout)             :: fum      !< FU.
+      double precision, intent(inout)             :: rum      !< RU.
       double precision, intent(  out)             :: aum      !< Computed flow area at structure.
       double precision, intent(  out)             :: dadsm    !< Computed flow width at structure.
       double precision, intent(in   )             :: bob0(2)  !< BOB's of the channel
       double precision, intent(in   )             :: s1m2     !< Water level at left side of universal weir.
       double precision, intent(in   )             :: s1m1     !< Water level at right side of universal weir.
       double precision, intent(  out)             :: qm       !< Computed discharge at structure.
-      double precision, intent(inout)             :: u1m      !< Computed flow velocity.
+      double precision, intent(  out)             :: u1m      !< Computed new flow velocity. Note that current velocity will be recomputed first
+                                                              !! based on given fum and rum, in case this universal weir is part of a compound.
       double precision, intent(in   )             :: dxm      !< Length of flow link.
       double precision, intent(in   )             :: dt       !< Time step in seconds.
       logical,          intent(in   )             :: changeStructureDimensions !< Indicates whether the crest level of the weir can be changed when
@@ -133,7 +134,7 @@ module m_Universal_Weir
       !
       !
       u1m =  rum - fum*( s1m2 - s1m1 )
-      qm = aum * u1m
+
       !
       !     Find the flow direction
       if (s1m1  > s1m2) then
@@ -151,8 +152,10 @@ module m_Universal_Weir
       if ((allowedflowdir == 3) .or. (dir == 1 .and. allowedflowdir == 2) .or. (dir == -1 .and. allowedflowdir == 1)) then
          fum  = 0.0d0
          rum  = 0.0d0
-         u1m  = 0.0d0
+         aum  = 0.0d0
+         dadsm = 0.0d0
          qm   = 0.0d0
+         u1m  = 0.0d0
          return
       endif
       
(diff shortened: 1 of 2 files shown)


src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
```diff
@@ -88,20 +88,22 @@ end subroutine deallocUniWeir
...
+      use precision_basics, only: equal
       implicit none
       !
       ! Global variables
       !
       type(t_uni_weir), pointer, intent(in)       :: uniweir  !< Universal weir object.
-      double precision, intent(  out)             :: fum      !< FU.
-      double precision, intent(  out)             :: rum      !< RU.
+      double precision, intent(inout)             :: fum      !< FU.
+      double precision, intent(inout)             :: rum      !< RU.
```
FlorisBuwaldaDeltares, line 98: I do not see anything in the issue description about intent changes of fum and rum? why is the "old" input value of fum and rum relevant for this routine?

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
```diff
@@ -354,10 +360,10 @@ subroutine uniweir_furu(s1m1, s1m2, qstru, qdh1, qdh2, dxm, dt, aum, fum, rum, u
       double precision, intent(in   ) :: dxm      !< Delta X
       double precision, intent(in   ) :: dt       !< Time step
       double precision, intent(in   ) :: aum      !< Flow area at link
-      double precision, intent(inout) :: fum      !< FU at link
-      double precision, intent(inout) :: rum      !< RU at link
+      double precision, intent(  out) :: fum      !< FU at link
+      double precision, intent(  out) :: rum      !< RU at link
```
FlorisBuwaldaDeltares, line 364: same question as above, why did intent change?

FlorisBuwaldaDeltares (commented):
one question, otherwise nice fix :)

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
FlorisBuwaldaDeltares: ah, i have traced it down and it appears that the only reason fum and rum are intent(inout) is because we need the flow velocity at the previous timestep. It would be cleaner to calculate that at the call site, and make u1m the intent(inout) argument and fum and rum both (out)

arthurvd: I see your point, but the ambition is to have all toplevel `compute<StructureType()` subroutines callable as is, without too many prior calculations on the call site with inside knowledge on fu/ru reconstructions.

It's a bit matter of taste, but I'd rather keep the u1m calculation delegated inside the computeUniversalWeir. Ok for you or not?

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
arthurvd: This is boy scout cleanup, not necessary for the bugfix. I spotted that the intents were inaccurate, so possibly giving readers false impressions. So I updated them to reflect what the subroutine code truly assumes about them.

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
FlorisBuwaldaDeltares: yes i agree, having these routines self-contained is also nice. However I do expect some differences in the testbench, the old intent(out) for fum and rum would let the compiler calculate u1m with uninitialized values, did no model results change?

FlorisBuwaldaDeltares (approved):

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
FlorisBuwaldaDeltares: ah i indeed see you have differences above tolerance in a compound structure. that will have to be a new reference

src/utils_gpl/flow1d/packages/flow1d_core/src/universal_weir.f90
```diff
@@ -132,7 +134,7 @@ subroutine ComputeUniversalWeir(uniweir, fum, rum, aum, dadsm, bob0, s1m1, s1m2,
       !
       !
       u1m =  rum - fum*( s1m2 - s1m1 )
-      qm = aum * u1m
+
```
rene-deltares, line 137: I am maybe missing some context just from looking at the diff and the JIRA issue description. But why was this line removed, is `qm` set elsewhere?

UPDATE: Oh, I see. It is set in `uniweir_furu`

rene-deltares (approved):
I checked, and I found no NaN's in the waterbalance variables in the new his file.

PR #1082, opened 2026-07-14
UNST-10125: Fix exit code -1073741819

rene-deltares:
# What was done 

See JIRA issue for more details. The error happens in python import phase. The container and build environments are squeeky clean so python writes lots of small `__pycache__/*.pyc` files. The error code happens there every once in a while. My hope is that by turning off the bytecode file writing that this error will go away. The builds will not be any slower, because we never reuse the bytecodes anyway, we only run TestBench.py once, and then the build directory is removed from the agent, and the files that are not bind-mounted in the container are ephemeral anyway.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/ci/teamcity/Delft3D/windows/test.kt
+++ b/ci/teamcity/Delft3D/windows/test.kt
@@ -131,9 +131,8 @@ object WindowsTest : BuildType({
                     rem Wheels come from the mounted uv cache volume.
                     uv venv C:\venv || (echo [ERROR] uv venv failed! & exit /b 1)
                     call C:\venv\Scripts\activate.bat
+                    set PYTHONVERBOSE=1
                     uv pip sync -v pip/win-requirements.txt || (echo [ERROR] uv pip sync failed! & exit /b 1)
-
-                    echo "Running TestBench.py..."
                     python TestBench.py %%argsList%%
             """.trimIndent()
 
--- a/ci/teamcity/Delft3D/windows/test.kt
+++ b/ci/teamcity/Delft3D/windows/test.kt
@@ -117,8 +117,10 @@ object WindowsTest : BuildType({
                     rem in a clean container, and the build directory is removed every time, so the
                     rem compiled bytecodes are never reused. In addition, we've run into strange
                     rem -1073741819 (0xC0000005) exit codes during bytecode compilations. We decided
-                    rem to turn off the byte code writing because of this.
+                    rem to turn off the byte code writing because of this. Also set PYTHONFAULTHANDLER
+                    rem in case we still get a crash. That should produce a stacktrace on the crashes.
                     set PYTHONDONTWRITEBYTECODE=1
+                    set PYTHONFAULTHANDLER=1
 
                     set argsList=--username %s3_dsctestbench_accesskey% ^
                     --password %s3_dsctestbench_secret% ^
--- a/ci/teamcity/Delft3D/windows/test.kt
+++ b/ci/teamcity/Delft3D/windows/test.kt
@@ -109,39 +109,40 @@ object WindowsTest : BuildType({
             name = "Run TestBench.py"
             id = "RUNNER_testbench"
             workingDir = "test/deltares_testbench/"
-                scriptContent = """
-                    @echo off
+            scriptContent = """
+                @echo off
 
-                    rem The environment and venv are completely clean, so python will write a lot
-                    rem of '.pyc' bytecode files in '__pycache__' directories. Since each build runs
-                    rem in a clean container, and the build directory is removed every time, so the
-                    rem compiled bytecodes are never reused. In addition, we've run into strange
-                    rem -1073741819 (0xC0000005) exit codes during bytecode compilations. We decided
-                    rem to turn off the byte code writing because of this. Also set PYTHONFAULTHANDLER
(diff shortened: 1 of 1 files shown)


ci/teamcity/Delft3D/windows/test.kt
```diff
@@ -106,36 +106,43 @@ object WindowsTest : BuildType({
...
+            scriptContent = """
+                @echo off
+
+                rem Python writes a lot of '.pyc' bytecode files in '__pycache__' directories. This build 
+                rem step runs in a clean container with a clean build directory bind-mounted in every time,
+                rem so the bytecode files are never reused. We've run into -1073741819 (0xC0000005) exit codes
+                rem while python was generating these '*.pyc' files. That's why we've decided to turn off the
+                rem bytecode file writing. In case we still get a crash, we set PYTHONFAULTHANDLER. This should
+                rem produce a stacktrace when python crashes.
+                set PYTHONDONTWRITEBYTECODE=1
```
hrajagers, line 121: Hm, negative logical ... unfortunately, unavoidable here.

FlorisBuwaldaDeltares (approved):
do you think those script comments are useful in a generated script nobody will ever read as an actual file? to me it would seem more logical to put the explanation in kotlin comments above the scriptcontent declaration

ci/teamcity/Delft3D/windows/test.kt
```diff
@@ -106,36 +106,43 @@ object WindowsTest : BuildType({
...
+
+                if "%copy_failed_cases%"=="true" (
+                    set argsList=%%argsList%% --copy-failed-cases
+                )
+
+                rem Create the venv on the container filesystem (C:\venv), NOT the bind-mounted work dir,
+                rem to avoid os error 32 file-lock failures on the mount during install.
+                rem Wheels come from the mounted uv cache volume.
+                uv venv C:\venv
+                if %%ERRORLEVEL%% NEQ 0 exit /b 1
```
hrajagers, line 141: Did you decide to change the syntax, or AI? What is the reasoning ... to avoid going back and forth between two equivalent implementations.

hrajagers (approved):

harmenwierenga:
> do you think those script comments are useful in a generated script nobody will ever read as an actual file? to me it would seem more logical to put the explanation in kotlin comments above the scriptcontent declaration

I think that the comments should be close to the code they are referencing. Which script is generated?

harmenwierenga (approved):

ci/teamcity/Delft3D/windows/test.kt
rene-deltares: When I picked up this issue I didn't know what command was producing the crashes, and since TeamCity reported the large negative exit code I was unsure if the `if errorlevel 1` check was actually caught. I did ask AI, and searched the internet, and they agreed that `if errorlevel 1` in `cmd.exe` means something like: `if errorlevel >= 1`. I have no idea if `cmd.exe` even recognizes negative exit codes, but at least for me, not a batch-aficionado, the `if %errorlevel% neq 0` is seems much more clear, and handles negative exit codes (even if they're hypothetical)

rene-deltares:
> do you think those script comments are useful in a generated script nobody will ever read as an actual file? to me it would seem more logical to put the explanation in kotlin comments above the scriptcontent declaration

These kind of comments to me are extremely useful, because otherwise nobody reading this code will ever understand why we set this obscure PYTHONDONTWRITEBYTECODE environment variable. I hope, for everyone's sake, that everyone who ever touches these files also reads and tries to understand the code in these files or I don't have very high confidence that we'll make our TeamCity pipelines less brittle.

FlorisBuwaldaDeltares:
yes the comments are useful, but my only question was should they be bash script comments or should they be kotlin comments, which file will developers end up reading..

rene-deltares:
> yes the comments are useful, but my only question was should they be bash script comments or should they be kotlin comments, which file will developers end up reading..

Ah, ok, sorry, then I understand you better. I thought with 'generated' you meant that the Kotlin changes were all AI generated or something, but I think you meant that Kotlin takes this big string with an embedded batch script, and generates a 'script' out of it, right?
It has also crossed my mind, but I think the script is a little too big to have just a Kotlin style-comment at the top of it, so I prefer this approach. To me this is all a single source code file, so whether the comment is embedded in a script doesn't matter.

hrajagers:
> > yes the comments are useful, but my only question was should they be bash script comments or should they be kotlin comments, which file will developers end up reading..
> 
> Ah, ok, sorry, then I understand you better. I thought with 'generated' you meant that the Kotlin changes were all AI generated or something, but I think you meant that Kotlin takes this big string with an embedded batch script, and generates a 'script' out of it, right? It has also crossed my mind, but I think the script is a little too big to have just a Kotlin style-comment at the top of it, so I prefer this approach. To me this is all a single source code file, so whether the comment is embedded in a script doesn't matter.

A post-merge additional argument besides the file size: An advantage of having the documentation inside the script itself is that this explanatory information is also available when inspecting the configuration in the TeamCity web interface. A comment inside the Kotlin part wouldn't be visible.

PR #1083, opened 2026-07-15
UNST-9538: Disable long tests (part 1)

rene-deltares:
# What was done 

Based on a recent Windows TestBench results on main, I took the tests that are taking more than 10 minutes and commented them out in the XML, with a note saying how long they took and the Jira issue id.

A separate PR to first disable the tests without putting them in a daily run was explicitly asked for because of the current long running times of the pipelines. The Jira issue UNST-9538 is *not yet* done after merging this, because we still need to put these tests in a 'daily run'. I propose just a single test bench config with all of the cases, including default test cases copied in. Then we can just run TestBench.py on this config. It will take long and occupy a single machine. But I believe we may be able to fit it in a single run on one machine if we take a 1,5 hour timeout. But we'll see.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
@@ -258,6 +258,7 @@ contains
                      if (present(varname2)) then
                         success = ecSetFileReaderProperties(ecInstancePtr, fileReaderId, ec_filetype, filename, refdate_mjd, tzone, ec_second, name, varname=varname, varname2=varname2)
                      else
+                        ! UNST-8900: Data value must be added as extra argument to ecSetFileReaderProperties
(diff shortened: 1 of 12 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -258,6 +258,7 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
                      if (present(varname2)) then
                         success = ecSetFileReaderProperties(ecInstancePtr, fileReaderId, ec_filetype, filename, refdate_mjd, tzone, ec_second, name, varname=varname, varname2=varname2)
                      else
+                        ! UNST-8900: Data value must be added as extra argument to ecSetFileReaderProperties
```
rene-deltares: Whoops, committed in the wrong branch, reverting this one

hrajagers (approved):
List of excluded cases matches what I would expect based on a recent "all" test bench run of "main".

PR #1086, opened 2026-07-15
Small quality of life improvement for devcontainer.json

harmenwierenga:
# What was done 

Explicitly list the third-party-libs tag, so that it is easier to change when working on specific branches


rene-deltares (approved):
I use this sometimes, so looks good

PR #1088, opened 2026-07-15
Clean up ancient and/or unused files in test/deltares_testbench

rene-deltares:
# What was done 

A month ago, I changed the TestBench.py TeamCity build to start using the 'Linux Python' docker image that we made to fix the broken PythonCiTools build. Up until then the testbench unit tests in TeamCity were performed with a `Dockerfile` and `docker buildx bake`. This was a nice experiment, but since nobody but my old self understood it, it's probably best to replace it with something simpler. Even if it is a little slower.

At the time I did not delete the `Dockerfile`, and the `docker-bake.hcl` file because it would mean that everyone who had an old branch (without the latest changes to main) would have to merge main back into their branch, otherwise their TeamCity pipeline would fail. To prevent this, I decided to leave the deletion of these files for a later PR. Now it should be safe to delete them.
 
In addition, I removed some other files that haven't been used in a long time. If you were still using TestBench.exe: I'm sorry, it has to go.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


harmenwierenga (approved):
Nice cleanup, when you can merge you can send a little message to the HMWESS channel perhaps

hrajagers (approved):

PR #1094, opened 2026-07-16
Fm/task/unst 9360 write map to polygon region

FlorisBuwaldaDeltares:
# What was done 

- modify generate_put_var_map.py to use flowgeom maps to correctly write to reduced output set
- add output polygonfile mdu argument
- support multiple polygonfiles
- fix small bugs encountered when making new testcase
- two testcases: 1D2D dar_es_salaam_output_polygon
- modified san francisco bay 3D model to write only the actual san francisco bay
- don't write empty map files
- fix small part of classmap writer (work in progress)
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_types.f90
```diff
@@ -19,8 +19,9 @@ module m_unstruc_netcdf_types
       type(t_ug_meshgeom) :: mesh2D !< Node/edge/face topology and coordinates for the 2D mesh.
       type(t_ug_meshgeom) :: mesh1D !< Node/edge/face topology and coordinates for the 1D mesh.
 
+      logical :: remapping_active = .false. !< True when polygon output reduction requires the mapping arrays below.
```
arthurvd, line 20: This PR is a good time to start introducing separate flowgeom instances for different purposes.
You're currently using the existing `m_unstruc_netcdf_data::flowgeom` variable. I recommend introducing:
`type(t_fm_flowgeom) :: map_flowgeom !< Flow geometry object used for (possibly polygon-cropped) map file output. `

This is already necessary now, because the dfm_volume_tool calls dflowfm.dll's `write_1D_flowgeom_ugrid()`, which calls:
```
call unc_write_1D_flowgeom_ugrid(flowgeom, mapids%id_tsp, ncid)
```
This particular `flowgeom` should *not* be the map-cropped one.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_pol_to_cellmask.f90
```diff
@@ -992,6 +1048,39 @@ end subroutine test_find_cells_crossed_by_polyline_edge_case2
...
+      close (unit, iostat=ierror)
+
+   end subroutine write_polygon_file
+
+   subroutine delete_polygon_file(filename)
+      character(len=*), intent(in) :: filename
+
+      integer :: ierror, unit
+
+      open (newunit=unit, file=filename, status='old', action='read', iostat=ierror)
```
arthurvd, line 1079: Why do you open this file? Would you not do `inquire(file=filename, exist=your_logical)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1826,6 +1823,7 @@ subroutine readMDUFile(filename, istat)
       call prop_get(md_ptr, 'output', 'enableDebugArrays', jawritedebug, success) ! allocate 1d, 2d, 3d arrays to quickly write quantities to map file
       call prop_get(md_ptr, 'output', 'NcNoUnlimited', unc_nounlimited, success)
       call prop_get(md_ptr, 'output', 'NcNoForcedFlush', unc_noforcedflush, success)
+      call prop_get(md_ptr, 'output', 'OutputPolygonFile', md_output_polyfile, success)
```
arthurvd: keyword should at least include "Map", so consider: "MapOutputPolygonFile".
(and please update JIRA issue description with this different name, to avoid future confusion with people reading the issue.)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
```diff
@@ -67,21 +67,42 @@ def generate_rank1(ftype: FortranType) -> str:
...
+      iloc = UNC_LOC_S
+      n1d_write = 0
+      call realloc(work, ndx2d, keepExisting=.false.)
+      do n = 1, ndx2d
+         if (remapping_active) then
+            g = flowgeom%face_map_2D(n)
+         else
+            g = n
+         end if
+         work(n) = values(ktop(g))
```
arthurvd: Suggestion: small elemental utility function `get_mapped_index()`, such that you could do:

```suggestion
         work(n) = values(ktop(get_mapped_index(n, flowgeom%face_map_2D))
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
```diff
@@ -67,21 +67,42 @@ def generate_rank1(ftype: FortranType) -> str:
...
+      end do
+      output_values => work
+   else if (remapping_active) then
```
arthurvd: I would not make this an "else if", but a mere "else", possibly with a nested if:
The `if (write_surface_output)` is about a different aspect than `(else) if (remapping_active)` , so keep the bodies of the IF and ELSE similar.

arthurvd (commented):
Intermediate review, see first remarks. I will only do the generate.py file tomorrow.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_pol_to_cellmask.f90
FlorisBuwaldaDeltares: as discussed, we need to open to get the unit as file deletion is usually done with close, status='delete'

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_types.f90
FlorisBuwaldaDeltares: yes, will do!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
FlorisBuwaldaDeltares: done. The logic remains the same (write_surface handles its own remapping which we don't want to overwrite) but making it a separate if is less confusing

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
FlorisBuwaldaDeltares: done, managed to convince copilot it was a good idea

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
FlorisBuwaldaDeltares: done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
```diff
@@ -507,8 +508,8 @@ module m_unstruc_netcdf_data
       integer, allocatable :: ilink_merge(:) !< like ilink_own, but from the merged restart file
    end type t_unc_merged
 
-   type(t_fm_flowgeom) :: flowgeom !< global flow geometry object to be built by build_flowgeom_2d and build_flowgeom_1d.
-
+   type(t_fm_flowgeom), pointer :: flowgeom_map => null() !< global flow geometry object, possibly cropped by a map output polygon.
```
arthurvd: We should to this initialization in a new `default_unstruc_netcdf_data()` to allow loading multiple MDUs in unit tests/Interacter.

But, there's more module state in this particular module that does the same, so we can do it in a separate cleanup later.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
```diff
@@ -508,8 +508,8 @@ module m_unstruc_netcdf_data
...
+   type(t_fm_flowgeom), target :: flowgeom_full !< global flow geometry object to be built by build_flowgeom_2d and build_flowgeom_1d.
```
arthurvd: Is this clear enough?
```suggestion
   type(t_fm_flowgeom), target :: flowgeom_full !< global flow geometry object to be built by build_flowgeom.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
```diff
@@ -49,39 +49,57 @@ def generate_rank1(ftype: FortranType) -> str:
...
+      case (UNC_LOC_S)
+         call realloc(work, flowgeom_map%ndx_out, keepExisting=.false.)
+         do i = 1, ndx2d
+            work(i) = values(flowgeom_map%face_map_2D(i))
          end do
-      end if
+         do i = 1, n1d_write
+            work(ndx2d + i) = values(flowgeom_map%node_map_1D(i))
+         end do
+      end select
```
arthurvd: Why are only LOC_S and _CN prepared here? And not also LOC_U and _L and _S3D, which are done below close to the writing code.

arthurvd (commented):
More review on your updated code since last Thursday.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
FlorisBuwaldaDeltares: it is a very lazy description, i'll think of something better

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/python/generate_unc_put_var_map.py
FlorisBuwaldaDeltares: hmm this was a leftover from a previous optimization attempt where the shared output_values => work seemed elegant. However you are right, right now it is needlessly inconsistent. Either all variables get preparations or none do. I'll merge these blocks with the big select case further down

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
FlorisBuwaldaDeltares: i've added a small reset routine

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_types.f90
```diff
@@ -21,24 +19,36 @@ module m_unstruc_netcdf_types
...
+contains
+
+   subroutine reset_fm_flowgeom(flowgeom)
+      type(t_fm_flowgeom), intent(inout) :: flowgeom
+
+      integer :: ierr
+
+      ierr = t_ug_meshgeom_destructor(flowgeom%mesh1D)
+      ierr = t_ug_meshgeom_destructor(flowgeom%mesh2D)
+      flowgeom = t_fm_flowgeom()
```
arthurvd, line 51: Aren't you leaking memory here? (any allocated `*_map_*` arrays have not been explicitly deallocated.)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
```diff
@@ -508,13 +508,35 @@ module m_unstruc_netcdf_data
...
+      if (associated(flowgeom_map)) then
+         call reset_fm_flowgeom(flowgeom_map)
+         if (.not. associated(flowgeom_map, flowgeom_full)) then
+            deallocate(flowgeom_map)
+         end if
+      end if
+      nullify(flowgeom_map)
+      call reset_fm_flowgeom(flowgeom_full)
+
+      mapids = default_mapids
```
arthurvd: Would you not rather say this (to avoid the need for a local variable)? (not sure whether it's equivalent):
`mapids = t_unc_mapids()`

arthurvd (commented):
As just discussed, nearly there.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_data.f90
FlorisBuwaldaDeltares: done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_netcdf_types.f90
FlorisBuwaldaDeltares: I have confirmed the assignment automatically deallocates the allocatable member fields of the old flowgeom object

arthurvd (approved):
Looks correct and nicer than before. Ready to merge!

PR #1096, opened 2026-07-17
Fm/task/unst 9482 replace fill value

mihai-pricope:
# What was done 

Replace _FillValue with NaN after reading NetCDF Swan. Afterwards all checks are done agains a NaN value.

# Evidence of the work done 

- [X]	Video/figures 

Diff between results for the processed swan netcdf and original (original has shorts and processed has doubles).
image
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9482

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
@@ -1629,7 +1629,7 @@ contains
          ! TODO: AvD: I'd rather have a full message stack that will combine EC + meteo + dflowfm, and any caller may print any pending messages.
          ! For now: Print the EC message stack here, and leave the rest to the caller.
          ! TODO: RL: the message below is from m_meteo::message, whereas timespace::getmeteoerror() returns timespace::errormessage. So now this message here is lost/never printed at call site.
-         message = dump_ec_message_stack(LEVEL_ERROR, callback_msg)
+         message = dump_ec_message_stack(LEVEL_WARN, callback_msg)
          ! Leave this concluding message for the caller to print or not. (via getmeteoerror())
       end if
       message = 'm_meteo::ec_addtimespacerelation: Error while initializing '''//trim(name)//''' from file: '''//trim(filename)//''''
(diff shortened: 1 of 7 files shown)


src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -2629,9 +2630,16 @@ elemental function cyclic_interpolation(var1, var2, weight1, weight2)
       real(dp) :: delta
       real(dp) :: weightfac
 
-      if (ieee_is_nan(var1) .or. ieee_is_nan(var2)) then
-         cyclic_interpolation = ieee_value(0.0_dp, ieee_quiet_nan)
-         return
+      if (present(dmissValue)) then
+         if (var1 == dmissValue .or. var2 == dmissValue) then
```
rene-deltares: Possible edge case: Can `dmissValue` be `NaN` as well, or no? If so, you can just move the `NaN` checks outside of the `else` branch I think?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1629,7 +1629,7 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
          ! TODO: AvD: I'd rather have a full message stack that will combine EC + meteo + dflowfm, and any caller may print any pending messages.
          ! For now: Print the EC message stack here, and leave the rest to the caller.
          ! TODO: RL: the message below is from m_meteo::message, whereas timespace::getmeteoerror() returns timespace::errormessage. So now this message here is lost/never printed at call site.
-         message = dump_ec_message_stack(LEVEL_ERROR, callback_msg)
+         message = dump_ec_message_stack(LEVEL_WARN, callback_msg)
```
rene-deltares, line 1632: Ah, this was causing that `ERROR ...` line right? How do I interpret this, is this only dumping the messages that have level WARNING or higher?

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
mihai-pricope: Thanks for spotting this. It turned out to be a bit more complicated but I pushed a solution for this.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
mihai-pricope: Yes, we now unwind the whole ec_message stack on WARN and log an Error afterwards. If the whole FM is set to stop on error it will stop there (instead on the first line of ec_message stack)

rene-deltares (approved):

PR #999, opened 2026-06-24
DELWAQ-1234: Update the source code and introduce auxiliary routines

arjenmarkusdeltaresnl:
To demonstrate the use of the auxiliary routines I edited the original example: Write a message to the screen via the logger
Check if the segment is active
I had to add a surprising number of modules to get the precision working.

# What was done 

<a short description with bullets> 

- e.g. Restarts are made more robust 
- e.g. Fixes a bug related to the writing of water depth on the map file 
- e.g. Introduces a new functionality on energy losses at bridge piers
- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
<add video/figures if applicable> 
- [ ]	Clear from the issue description 
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
<add testcase numbers if applicable, Issue number>
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
<add description of changes if applicable, Issue number> 
- [ ]	Not applicable 

# Issue link

PR #1098, opened 2026-07-17
Revert "DELWAQ-1234: Update the source code and introduce auxiliary routines"

hrajagers:
Reverts Deltares/Delft3D#999

Branches of type "all" fail with the message:

Extra files:
share\delft3d\open_process_lib_example\code\decayt.dll
Directory check failed: Missing or extra files detected.


hrajagers:
#1099 implements a better solution: going forward instead of reverting.

PR #1100, opened 2026-07-17
All/task/unst 10162 conan oneapi 2026

harmenwierenga:
# What was done 

- Make it possible to run ConanPackages with OneAPI 2026 for both Linux and Windows
- Detect the intel version in run_conan.py, provide argument for choosing it yourself
- Add conan profiles for OneAPI 2026
- Update Conan recipes to use version of CMake that knows VS2026
- Clean up old cmake that was used for OneAPI 2021

# Issue link UNST-10162


conan/config/settings_user.yml
```diff
@@ -1,8 +1,19 @@
...
 os:
     Linux:
         distro: ["Alma8", "Alma9", "Alma10"]
 
-fortran_compiler: [null, ifx, gfortran, nvfortran, crayftn, nagfor]
+fortran_compiler:
+    null:
+    ifx:
+        version: ["2024.2", "2026.1"]
+    gfortran:
```
KoenBussemakerDeltares: I think we should only add the compilers that we actually support here

doc/compiling_Linux.md
```diff
@@ -231,16 +231,17 @@ This is what `python run_conan.py initialize deltares` does under the hood. The
...
+lockfile [conan.lock](../conan.lock) pins recipe revisions for reproducibility. On Linux we use a single-config generator (`Unix Makefiles`),
 so the build type is selected at both `conan install` and `cmake` time. The third-party packages
 themselves are always built/downloaded as `Release`, while the consumer (CMakeDeps generator) build type is selected via
 `&:build_type=...`.
 
 ```bash
 # 1. Install dependencies (generates CMakeDeps files).
 #    The first call may build packages (or download them from Nexus).
-conan install . --profile:all=delft3d_alma8_intel_2024_v2 \
+conan install . --profile:all=delft3d_alma8_intel_2024_v3 \
```
KoenBussemakerDeltares: Use the 2026 profile, keep 2024 as a fallback and not the default. Same for line 263.

doc/compiling_Windows.md
```diff
@@ -209,18 +211,18 @@ Run `conan install` once per configuration you want to consume:
 ```bat
 :: 1. Install dependencies for all three configurations.
 ::    The first call may build packages (or download them from Nexus). The other two reuse the cache.
-conan install . --profile:all=delft3d_windows_msvc_194_v2 ^
+conan install . --profile:all=delft3d_windows_msvc_194_v3 ^
```
KoenBussemakerDeltares: Same here: point to the newer profile

run_conan.py
```diff
@@ -266,14 +274,45 @@ def upload_new_packages(remote: str, *, ci: bool = False) -> None:
     print(f"\nDone. Uploaded: {uploaded}, skipped: {skipped}")
 
 
-def _get_profile() -> str:
+def _detect_profile() -> str:
```
KoenBussemakerDeltares: I'm not a fan of auto detect mechanisms like this, they tend to be fragile and annoying to debug when things go wrong (both for users and for us). I would much rather like a situation where the user has to provide an additional command line argument in order to switch back to 2024, and take 2026 by default. This also servers as an incentive to update to the new compiler.

We will now be in a transition phase where most people will still be on 2024 and will have to move to 2026. This will be temporary and will only be for say half a year, after which we can expect everyone to be on the new compiler version. We can then just throw away the 2024 command line option.

doc/compiling_Linux.md
harmenwierenga: I am working on supporting the new compiler, but there is still work to be done before we can use it. We have not tested the new compiler yet, it is not yet supported and the linux build does not work at all at the moment. This is a necessary step to make it possible to support the compiler in the future, but I want to incrementally merge the changes necessary for the new compiler. We are not ready to change the default.

run_conan.py
harmenwierenga: Our build.bat has always detected the visual studio version as well, and it works well. Our cmake also detected the intel version through a much more complicated mechanism to support our old CMake mechanism for third party dependencies, and this PR removes that mechanism. I think that this is a quality of life improvement. You can still override it if it would ever go wrong

conan/config/settings_user.yml
harmenwierenga: There are several external and internal projects to support different compilers in Delft3D. Since we don't want to force people to change this too often, I think that it is nice to have an available list. These are really just here to make it possible for conan to calculate hashes based on these names, it has no other functional effect in our codebase

doc/compiling_Windows.md
```diff
@@ -198,8 +198,10 @@ This is what `python run_conan.py initialize deltares` does under the hood. The
 variant additionally removes the Nexus remotes and registers [conan/recipes](../conan/recipes)
 as a `local-recipes-index` remote. See [run_conan.py](../run_conan.py) for details.
 
-The Conan profile is `delft3d_windows_msvc_194_v2` and the lockfile [conan.lock](../conan.lock)
-pins recipe revisions for reproducibility.
+`run_conan.py` detects the active Visual Studio environment and selects
```
rene-deltares: It's a bit unfortunate that with every change of the profile we'll have update all these docs with the new default value. Could we write this in a more future-proof manner? I have little faith that we're disciplined enough to do this every time and end up confusing the mighty morphin power users more than helping them.

run_conan.py
rene-deltares: I must say I'm not a fan either. If people don't want the default, let them override it with exactly the profile they want.
Suppose we do support one of those other fortran compilers in `settings_user.yml` for instance. Then this function will not work anymore. And the code will only get more and more complicated the more profiles we support. I think the inconvenience caused by bugs in this auto-detection code will be much greater than the convenience of not having to pass the `--profile` argument if you don't want the default.

src/cmake/functions.cmake
```diff
@@ -432,76 +432,3 @@ function(create_test test_name)
...
-        if(pair MATCHES "^${key}:")
-            string(REPLACE "${key}:" "" result "${pair}")
-            break()
-        endif()
-    endforeach()
-    set(${value} "${result}" PARENT_SCOPE)
-endfunction()
-
-# Function to return ifort version number
-function(get_intel_version)
```
rene-deltares, line 482: Good riddance

conan/config/settings_user.yml
rene-deltares: I agree with Koen. It would be great if we had support for other toolchains, but we don't have it currently and if we do in the future we can add them to this file then. I get the feeling that if we support one other Fortran compiler we're going to have to change this file anyway (with a sensible version number for isntance). Furthermore we will probably have to make a lot more changes to a lot of other files as well.

rene-deltares: I guess we should clarify what we mean by "support"? We could try to build all of our conan packages with `gfortran`, for example. If we get that working, we could already add a version of `gfortran` to this file and say we support `gfortran` for our conan packages. Even if we don't know how to build the rest of our stuff with `gfortran` yet. 
That's kind of like how with this PR, you added "support" for oneapi 2026, even though we do not build with it yet on TeamCity, and we don't run the TestBenches with those binaries.

doc/compiling_Windows.md
harmenwierenga: I have updated the docs to use general descriptions and the new environment variable so that they don't require constant updating

run_conan.py
harmenwierenga: I have removed the autodetect. To still allow some ergonomics for developers on a non-default toolchain, such as myself at the moment, I ensure that the script also listens to the standard CONAN_DEFAULT_PROFILE environment variable (see https://docs.conan.io/2/reference/environment.html). I have also changed the error message in run_conan.py to tell users to update it if the profile gets a version bump

conan/config/settings_user.yml
harmenwierenga: I have removed them. My point for adding them was that we introduce a new global conan setting, and that really says nothing about which compilers we support. It does have the unintended side effect that we automatically break anyone who tries to build our software with a different compiler for a reason other than that the compiler does not work with our code

rene-deltares (approved):
Thanks!

doc/compiling_Linux.md
```diff
@@ -217,31 +221,25 @@ Note that the default tag used here deviates from the one set in the prerequisit
...
+`Unix Makefiles` generator, so select the consumer build type during both `conan install` and CMake
+configuration. Third-party packages are always built in their `Release` configuration. The CMakeDeps
+generator uses `&:build_type=...` to select the configuration for the consumer, which is Delft3D.
 
 ```bash
 # 1. Install dependencies (generates CMakeDeps files).
 #    The first call may build packages (or download them from Nexus).
-conan install . --profile:all=delft3d_alma8_intel_2024_v3 \
+#    Use `conan profile list` to find the current profile name.
+DELFT3D_CONAN_PROFILE=PROFILE_FROM_CONAN_PROFILE_LIST
```
KoenBussemakerDeltares: Change PROFILE_FROM_CONAN_PROFILE_LIST to something like <Profile from conan list> to clearly indicate that this is something that the user must provide, not a local variable

doc/compiling_Windows.md
```diff
@@ -211,18 +207,20 @@ Run `conan install` once per configuration you want to consume:
 ```bat
 :: 1. Install dependencies for all three configurations.
 ::    The first call may build packages (or download them from Nexus). The other two reuse the cache.
-conan install . --profile:all=delft3d_windows_msvc_194_v3 ^
+::    Replace the value below with the applicable name from `conan profile list`.
+set DELFT3D_CONAN_PROFILE=PROFILE_FROM_CONAN_PROFILE_LIST
```
KoenBussemakerDeltares: Same here

run_conan.py
```diff
@@ -274,57 +267,35 @@ def upload_new_packages(remote: str, *, ci: bool = False) -> None:
...
-        )
-
-    sys.exit(f"ERROR: Unsupported operating system: {os_name}")
+def _get_default_profile() -> str:
+    if platform.system() == "Windows":
+        return DEFAULT_CONAN_PROFILE_WINDOWS
+    return DEFAULT_CONAN_PROFILE_LINUX
 
 
 def _get_profile(profile_override: str | None) -> str:
```
KoenBussemakerDeltares: Rename to `_get_profile_or_default` or something along those lines. This it clear why `profile = _get_profile(args.profile)` is needed instead of just reading the value directly from `args`.

KoenBussemakerDeltares (approved):

PR #1101, opened 2026-07-19
UNST-10154: fix crash for dimensionless realization

mihai-pricope:
# What was done 

Fix a crash with dimensionless realization. This seems to be ok-ish used further with a check:
```
            if (realization_dimid > 0) then
               dim_offset = 1
               nrel = fileReaderPtr%dim_length(dimids(1))
            else
               dim_offset = 0
               nrel = 0
            end if
```

So <=0 value is actually ok
 

# Evidence of the work done 

- [ ]	Video/figures \
- [X]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-10154

### DIFF
--- a/src/utils_lgpl/ec_module/packages/ec_module/src/ec_support.f90
+++ b/src/utils_lgpl/ec_module/packages/ec_module/src/ec_support.f90
@@ -1362,8 +1362,7 @@ contains
                                          x_varid, x_dimid, y_varid, y_dimid, &
                                          z_varid, z_dimid, &
                                          tim_varid, tim_dimid, &
-                                         series_varid, series_dimid, &
-                                         realization_varid, realization_dimid) result(success)
+                                         series_varid, series_dimid) result(success)
       use netcdf
       logical :: success
       integer, intent(in) :: ncid !< NetCDF file ID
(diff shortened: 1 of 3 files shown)


FlorisBuwaldaDeltares (commented):
I don't understand this code or context, but I don't think it's a bad thing to skip this if the dimension is zero. However, ideally we would give an error? why would strcmpi(std_name, realization) succeed but contain an invalid id? it would seem the file was incorrectly generated

mihai-pricope:
> I don't understand this code or context, but I don't think it's a bad thing to skip this if the dimension is zero. However, ideally we would give an error? why would strcmpi(std_name, realization) succeed but contain an invalid id? it would seem the file was incorrectly generated

The NetCDF file header looks like this in this case:
```
	int64 number ;
		number:long_name = "ensemble member numerical id" ;
		number:units = "1" ;
		number:standard_name = "realization" ;
```

It does not look like really well formed but I think it might be some "implicit" default when it contains only one (0).

You would expect something like:
```
	int64 number(number) ;
		number:long_name = "ensemble member numerical id" ;
		number:units = "1" ;
		number:standard_name = "realization" ;
```
and a new dimension ```number``` that contains only one element or something.

JulienGroenenboom (commented):
`nrel` in ec_provider.f90 is adjusted based on realization_dimid, but not used. I suggest removing the lines related to `nrel`.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_support.f90
JulienGroenenboom, line 1417: `nrel` in ec_provider.f90 is adjusted based on realization_dimid, but not used. I suggest removing the lines related to `nrel`.

JulienGroenenboom, line 1417: The only purpose of these `realization` related code seems to be setting dim_offset to 0 or 1 and subsequently `nlay` via `nlay = fileReaderPtr%dim_length(dimids(3 + dim_offset))`.

FlorisBuwaldaDeltares (approved):
seems like we removed an ugly exception, I can't assess whether we missed anything but otherwise looks good

PR #1103, opened 2026-07-20
UNST-10020: Revert #729 - Continue (hidden) support of #values between hashes# in ini-files

JulienGroenenboom:
# What was done 
- Revert #729 
 
# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated: undo the creation of unit test
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10020


KoenBussemakerDeltares (approved):
Too bad we have to revert this but I understand why. Once we have the dflowfm_io library up and running we will no longer be sharing code, which means we can do this properly.

PR #1106, opened 2026-07-21
All/bugfix/unst 10119 fix test model timeouts

aslowwriter:
# What was done 

- removes all old timeout overrides from the xmls as they never have done anything
- adds a default timeout of 5 min
- adds a new override for the tests that need it. I made a good effort to find good timouts that minimise the chance of a false negative without being overly generous, but without hard data this always remains a matter of guess work. Developers should not hesitate to increase the timeout if they find it gives false negatives often, though I hope the'll keep it reasonable (e.g. no 99999 which defeats the purpose) I'll make a little post about it when it's merged, but basically I recommend rounding up the next minute 

Also incorporates suggestions from (and thus supersedes) #1034 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

[UNST-10119](https://issuetracker.deltares.nl/browse/UNST-10119)


aslowwriter:
When these tests run, some will fail because the default timeout is too low for them, I will give them a proper timeout before this is merged.

aslowwriter:
Note to self: on windows the timeout seems to do nothing yet: investigate

tmp.txt
rene-deltares, line 1: I think you accidentally added this file, no?

test/deltares_testbench/src/suite/program.py
```diff
@@ -204,17 +212,70 @@ def __start_process(self, logger):
...
+            "stderr": subprocess.PIPE,
+            "env": env,
+            "cwd": working_dir,
+        }
+        if platform.system() == "Windows":
+            popen_kwargs["creationflags"] = subprocess.CREATE_NEW_PROCESS_GROUP
+        else:
+            popen_kwargs["start_new_session"] = True
+
+        proc = subprocess.Popen(exec, **popen_kwargs)
```
rene-deltares, line 252: If anyone else reading this review is wondering: `subprocess.run` has a `timeout` argument to kill a process when it goes over the max run time. But it does not work on Windows at all. So we have to rely on "good old" `subprocess.Popen` instead

test/deltares_testbench/src/suite/test_case.py
```diff
@@ -30,20 +31,6 @@ def __init__(self, config: TestCaseConfig, logger: ILogger) -> None:
         logger.debug(f"Initializing test case ({self.__config.name}), max runtime : {str(self.__maxRunTime)}")
 
         self.__config.run_file_name = os.path.join(self.__config.absolute_test_case_path, "_tb3_char.run")
```
rene-deltares, line 33: Wow, they made this complicated system with checking if the run-time of the test case did not exceed the "reference run-time" (`_tb3_char.run` contains timing information). I wonder if it ever actually worked?

rene-deltares: Anyway, it doesn't seem to work at all now, so good riddance.

test/deltares_testbench/src/suite/test_case.py
```diff
@@ -143,6 +130,11 @@ def __initializeProgramList__(self, programs: List[Program]):
             program_config.shell = shell
             program_config.case_name = self.__config.name
             program_copy.overwriteConfiguration(program_config)
+            if self.__maxRunTime:
```
rene-deltares, line 133: Ah, OK. So instead of passing the timeout from the test case into the 'run program' methods, you just set the max runtime in the program just before running the case. That works!

test/deltares_testbench/src/suite/test_case.py
rene-deltares: By the way: Is `__findCharacteristicsRuntime__` now unused? You can remove the method in that case

test/deltares_testbench/src/suite/test_set_runner.py
```diff
@@ -301,11 +301,12 @@ def run_test_case(
             if not skip_testcase:
                 logger.test_Result(TestResultType.Exception, str(exception))
 
-        logger.test_finished()
-        if in_use is not None:
-            with idle_process:
-                in_use.value -= config.process_count
-                idle_process.notify_all()
+        finally:
+            logger.test_finished()
```
rene-deltares, line 305: Yeah, this is probably safer. Even though I think the `except` block should catch all exceptions (not even sure tbh)

rene-deltares (changes requested):

rene-deltares (changes requested):
Looks good, just a few very tiny changes requested!

tmp.txt
aslowwriter: I did, I"ve remvoed it

test/deltares_testbench/src/suite/test_set_runner.py
aslowwriter: The difference between a bare except and a finally, is that the finally block should be called regardless of whether an exception is raised, so things are cleaned up whether an exception is raised or we exist cleanly. Do you want me to make any modifacations?

test/deltares_testbench/src/suite/test_case.py
aslowwriter: I had to do it like that bc the timeout needs to be available to the program when it gets run, and I got a little last in the control flow. Regardless I do think it's a nice idea to have a default timeout for *all* programs, which can then be overridden by any testcase.  I think this is nicest in case anyone forgets to specify it

test/deltares_testbench/src/suite/test_case.py
aslowwriter: I've removed it, we'll see what the tests say

test/deltares_testbench/src/suite/test_set_runner.py
rene-deltares: No, it's better with your change

rene-deltares (approved):
Nice, thank you!

PR #1109, opened 2026-07-22
Fm/task/unst 10159 long culvert valverelativeopening

FlorisBuwaldaDeltares:
# What was done 

- valve relative opening for long culverts is now also written to his file when the culvert is dry
- flow direction determination now happens during "find1d2dlongculvertlinks", positive flow direction follows the long culvert polyline coordinates
- added unit test
- limit valve relative opening to between 0 and 1

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -880,13 +934,17 @@ subroutine reduceFlowAreaAtLongculverts()
...
+      real(dp) :: valve_relative_opening
 
       do i = 1, nlongculverts
          if (longculverts(i)%numlinks > 0) then
             L = abs(longculverts(i)%flowlinks(1))
             if (L > 0) then
-               au(L) = longculverts(i)%valve_relative_opening * au(L)
+               valve_relative_opening = min(longculverts(i)%valve_relative_opening, 1.0_dp)
+               valve_relative_opening = max(valve_relative_opening, 0.0_dp)
+               au(L) = valve_relative_opening * au(L)
```
sumihar: This is an important improvement indeed. In the past, the simulation continued to run even with valve_relative_opening > 1, modifying the au.

Is it perhaps better to modify the longculverts%valve_relative_opening directly? I mean, I suppose longculverts%valve_relative_opening (or valstruct?) will be used in writing to his file instead of the local variable valve_relative_opening? Then user can inspect the his file for the valve_relative_opening that is actually used in the computation.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -849,6 +858,51 @@ subroutine longculvertsToProfs(skiplinks)
...
+      end if
+
+      endpoint_is_first_node = equal(xk(k1), longculverts(ilongc)%xcoords(2)) .and. equal(yk(k1), longculverts(ilongc)%ycoords(2))
+      endpoint_is_second_node = equal(xk(k2), longculverts(ilongc)%xcoords(2)) .and. equal(yk(k2), longculverts(ilongc)%ycoords(2))
+
+      if (endpoint_is_first_node .and. .not. endpoint_is_second_node) then
+         longculverts(ilongc)%flow_dir = -1
+      else if (.not. endpoint_is_second_node) then
+         call mess(LEVEL_WARN, 'Cannot match the second coordinate of long culvert '//trim(longculverts(ilongc)%id)// &
+                   ' to either endpoint of its first flow link;')
```
sumihar, line 907: What will happen in this case? Will the simulation continue without longCulvert being activated?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -880,13 +934,17 @@ subroutine reduceFlowAreaAtLongculverts()
...
       do i = 1, nlongculverts
          if (longculverts(i)%numlinks > 0) then
             L = abs(longculverts(i)%flowlinks(1))
             if (L > 0) then
-               au(L) = longculverts(i)%valve_relative_opening * au(L)
+               valve_relative_opening = min(longculverts(i)%valve_relative_opening, 1.0_dp)
+               valve_relative_opening = max(valve_relative_opening, 0.0_dp)
+               au(L) = valve_relative_opening * au(L)
                call getflowdir(L, L_dir)
+               L_dir = longculverts(i)%flow_dir * L_dir
```
sumihar, line 951: I wonder if this is correct. 

I suppose if L_dir output of getflowdir is negative, while allowed_flowdir is FLOWDIR_NEGATIVE, then there should be flow through the longCulvert. 

But this line will make L_dir becomes positive, and hence au(L) = 0 and there is no flow through the longCulvert?

sumihar (commented):
Hi Floris,

thanks for the code changes. I have a few questions. Please have a look.

Many thanks,
Julius

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: you are right, then at least the correct value will be written to the his file. I'll change it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: hmm maybe this should indeed be a level error. However, I'm always hesitant to make errors that the user doesn't know how to fix. In this case, if something went wrong and we cannot reliably determine the flow direction.. it doesn't really matter that much. That was my reasoning but if you think it should be an error i can change it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: hmm let me think. 

L_dir = negative, flow_dir = negative => actual flow is positive
L_dir = positive, flow_dir = negative => actual flow is negative
L_dir = negative, flow_dir = positive => actual flow is negative
L_dir = positive, flow_dir = positive => actual flow is positive

So it seems to me that indeed the multiplication of the flow orientation and the link orientation should be used to determine to block flow or not

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
sumihar: I think, if it doesn't matter much, we better use LEVEL_WARN indeed. 

How I understand it now is that, in this case, the simulation will continue to run with the assumption that flow_dir = 1. I can't oversee now what the consequences could be. For now, we could leave it as it is now. We could pay attention to this when testing.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
sumihar: Thanks for the explanation. I needed to think more carefully. But now I understand it.

sumihar (commented):
Hi Floris,

thanks for you response. The codes look good! We still have to test it though.

Greetings,
Julius

sumihar (approved):
Nice work! I look forward to applying it!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -427,6 +427,8 @@ subroutine replaceCoordinatesInStructures(xcoords, ycoords, structures)
                   end if
                   call prop_set(current, '', 'xCoordinates', xcoords(coordindex:coordindex + ncoords - 1), '')
                   call prop_set(current, '', 'yCoordinates', ycoords(coordindex:coordindex + ncoords - 1), '')
+                  longculverts(i)%xcoords = xcoords(coordindex:coordindex + ncoords - 1)
```
KoenBussemakerDeltares, line 435: What happens if ncoords is 0?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1382,12 +1439,14 @@ subroutine find1d2dculvertlinks(network, longculvert, numcoords)
...
       integer, intent(in) :: numcoords !< number of polyline coordinates
       type(t_longculvert), intent(inout) :: longculvert !< A givin long culvert
-      integer :: i, j, branch_idx, contact_idx, othernode, nodenum, linknum, linkabs, is, ie, jafounds, jafounde, L_net
+      integer :: i, j, branch_idx, contact_idx, othernode, nodenum, linknum, linkabs, is, ie, jafounds, jafounde, L_net, direction_target
       integer, allocatable :: inode(:), inodeGlob(:), jnode(:)
 
       integer :: ierror
 
       longculvert%flowlinks = 0
+      longculvert%flow_dir = 1
```
KoenBussemakerDeltares: Shouldn't this be FLOWDIR_POSITIVE?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -849,6 +858,51 @@ subroutine longculvertsToProfs(skiplinks)
...
+          size(longculverts(ilongc)%xcoords) < 2 .or. size(longculverts(ilongc)%ycoords) < 2) then
+         return
+      end if
+
+      Lnet = longculverts(ilongc)%netlinks(1)
+      if (Lnet <= 0 .or. Lnet > size(kn, dim=2)) then
+         return
+      end if
+
+      k1 = kn(1, Lnet)
```
KoenBussemakerDeltares: I would prefer some more descriptive names. I know it is related to kn / xk / yk but I don't like those names either :P. Same for Lnet.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -849,6 +858,51 @@ subroutine longculvertsToProfs(skiplinks)
...
+         return
+      end if
+
+      endpoint_is_first_node = equal(xk(k1), longculverts(ilongc)%xcoords(2)) .and. equal(yk(k1), longculverts(ilongc)%ycoords(2))
+      endpoint_is_second_node = equal(xk(k2), longculverts(ilongc)%xcoords(2)) .and. equal(yk(k2), longculverts(ilongc)%ycoords(2))
+
+      if (endpoint_is_first_node .and. .not. endpoint_is_second_node) then
+         longculverts(ilongc)%flow_dir = -1
+      else if (.not. endpoint_is_second_node) then
+         call mess(LEVEL_WARN, 'Cannot match the second coordinate of long culvert '//trim(longculverts(ilongc)%id)// &
```
KoenBussemakerDeltares, line 906: So what does this mean for a user? Do they need to undertake any action?

And would this be a severe enough everything-is-broken-situation to throw an error instead?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
```diff
@@ -656,6 +656,56 @@ subroutine test_flow_reverse_head_gives_negative_discharge() bind(C)
...
+      longculverts(1)%allowed_flowdir = FLOWDIR_POSITIVE
+      longculverts(1)%valve_relative_opening = 1.0_dp
+      au(lc_link) = 1.0_dp
+      u1(lc_link) = real(longculverts(1)%flow_dir, dp)
+      call reduceFlowAreaAtLongculverts()
+
+      call f90_expect_true(au(lc_link) > 0.0_dp, &
+                           cstr("positive flow in input-coordinate direction must remain open"))
+
+      call default_longculverts
```
KoenBussemakerDeltares: Missing parenthesis

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
```diff
@@ -656,6 +656,56 @@ subroutine test_flow_reverse_head_gives_negative_discharge() bind(C)
    end subroutine test_flow_reverse_head_gives_negative_discharge
    !$f90tw)
 
+   !$f90tw TESTCODE(TEST, test_longculvert, test_allowed_flowdir_follows_coordinate_order, test_allowed_flowdir_follows_coordinate_order,
+   !> Verifies that direction reconstruction and allowedFlowDir follow the
+   !! input-coordinate order when the local flow-link orientation changes.
+   subroutine test_allowed_flowdir_follows_coordinate_order() bind(C)
```
KoenBussemakerDeltares, line 662: This is a very high level test. It would be nice if you could write a test for `setLongCulvertFlowDirectionSign` directly instead of the indirect effects. You might end up having to write large amounts of test setup code though, which would default the purpose. I leave it up to your judgment :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: thinks go kaboom. in the current program flow any long culvert without a valid set of x-coordinates would have already been rejected by the reader. but technically indeed this routine is making a few validity assumptions

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: agreed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: well that's the thing, i discussed this with Julius, if this happens I think we are a bit screwed. However, i'm not sure what happens in a partitioned model where maybe the long culvert is chopped in two by a partition boundary. That's why i opted for a warning, also because this is completely useless for a user, there is no clear reason why something went wrong or what they should fix. 

Basically I don't want to introduce a new reason why we get a ton of support issues from users who built some wacky model where apparently this can happen and now they can't do their calculations.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: it should, good catch

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
FlorisBuwaldaDeltares: this follows the structure of the surrounding long culvert semi-integration tests. Additionally, indeed the amount of boilerplate needed to construct what is under test here would almost duplicate the actual setup code. 

A direct test for setlongculvertflowdirectionsign can be written, it might seem trivial but that would be a proper unit test

mihai-pricope: I think this type of semi-integration tests might be more useful for now than a pure unit-test.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
mihai-pricope: As we have spoken, let's make this an error because we are not 100% in what scenarios will this happen.  So if this will happen in the wild, we will know the exact scenario and then we can make a more informed decision then.

FlorisBuwaldaDeltares: done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -849,6 +863,51 @@ subroutine longculvertsToProfs(skiplinks)
...
+         return
+      end if
+
+      netnode_1 = kn(1, L_net)
+      netnode_2 = kn(2, L_net)
+      if (netnode_1 <= 0 .or. netnode_2 <= 0) then
+         return
+      end if
+
+      endpoint_is_first_node = equal(xk(netnode_1), longculverts(ilongc)%xcoords(2)) .and. equal(yk(netnode_1), longculverts(ilongc)%ycoords(2))
```
arthurvd: Why not numlinks+1 instead of 2?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -849,6 +863,51 @@ subroutine longculvertsToProfs(skiplinks)
...
+      endpoint_is_second_node = equal(xk(netnode_2), longculverts(ilongc)%xcoords(2)) .and. equal(yk(netnode_2), longculverts(ilongc)%ycoords(2))
+
+      if (endpoint_is_first_node .and. .not. endpoint_is_second_node) then
+         longculverts(ilongc)%flow_dir = -1
+      else if (.not. endpoint_is_second_node) then
+         call mess(LEVEL_ERROR, 'Cannot match the second coordinate of long culvert '//trim(longculverts(ilongc)%id)// &
+                   ' to either endpoint of its first flow link;')
+      end if
```
arthurvd: why check on `.not. endpoint_is_second_node`?
Nothing happens here if `endpoint_is_first_node .and. endpoint_is_second_node`, but you don't want that at all, right?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_longculverts_data.f90
```diff
@@ -55,6 +55,7 @@ module m_longculverts_data
       !< 1 only positive flow
       !< 2 only negative flow
       !< 3 no flow allowed
+      integer :: flow_dir = 1 !< Maps the flow-link direction to the first-to-last coordinate direction.
```
arthurvd: You specifically mean that the first flow link in the long culvert is used as the representative flowlink for the entire long culvert, to base its orientation on, right? So, is this more meaningful:
```suggestion
      integer :: first_flow_link_orientation = 1 !< Maps the flow-link direction to the first-to-last coordinate direction.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
arthurvd: spotted by accident:

```suggestion
      if (branch_idx > 0 .and. network%BRS%size >= branch_idx) then
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1376,18 +1438,21 @@ subroutine find1d2dculvertlinks(network, longculvert, numcoords)
...
       integer, intent(in) :: numcoords !< number of polyline coordinates
       type(t_longculvert), intent(inout) :: longculvert !< A givin long culvert
-      integer :: i, j, branch_idx, contact_idx, othernode, nodenum, linknum, linkabs, is, ie, jafounds, jafounde, L_net
+      integer :: i, j, branch_idx, contact_idx, othernode, nodenum, linknum, linkabs, is, ie, jafounds, jafounde, L_net, direction_target
       integer, allocatable :: inode(:), inodeGlob(:), jnode(:)
 
       integer :: ierror
 
       longculvert%flowlinks = 0
+      longculvert%flow_dir = FLOWDIR_POSITIVE
```
arthurvd: I find this confusing: you're using an enum that is intended for allowedFlowDir values, whereas in your context the `longculvert%flow_dir` is actually something more like a `longculvert%orientation`.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
arthurvd: I now get it, I was confused by "endpoint", which is actually meant as "second_flow_node_is_first_longculvert_node".

arthurvd (commented):
Here are the remarks we just discussed over Teams.
@mihai-pricope: apologies by the way, I overlooked that you took this issue on your review plate this morning already.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: agree, it can just be else

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_longculverts_data.f90
FlorisBuwaldaDeltares: i prefer just "orientation" if that's ok with you.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: renamed for clarity, i've shortened a few things because writing it out in full is too long

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: done, good catch!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: agree. I considered making a separate enum for POSITIVE and NEGATIVE but I decided against it. for something as simple as a orientation -1 and 1 are intuitive enough, because then you also get the implicit understanding that -1 * -1 = 1

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: it is now: netnode_1_is_LC_node_2 & netnode_2_is_LC_node_2

arthurvd (approved):
all good now

PR #1111, opened 2026-07-22
All/task/unst 10120 triagle conan

harmenwierenga:
# What was done 

- Created conan recipe for Triangle
- Cleaned up build files for Triangle in Delft3D third_party_dependencies
- Use the triangle sources from Delft3D GitHub, not from triangle GitHub: the Triangle repo does not provide tricall, which we use, only a tricall test program. We have also made changes to the triangle sources over the years
- Clean up some cmake
- Change triangle code to build with icx 2026 and adhere to the C standard
- Change ConanPackages builds to only rebuild all packages once a week to speed it up

# Issue link UNST-10120


conan/recipes/triangle/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+from conan.tools.scm import Git
+import os
+
+required_conan_version = ">=2"
+
+class TriangleConan(ConanFile):
+    name = "triangle"
+    license = "Custom"
+    description = "A Two-Dimensional Quality Mesh Generator and Delaunay Triangulator."
+    url = "https://github.com/Deltares/Delft3D"
```
KoenBussemakerDeltares, line 15: Add a comment about why we are using our own source and not the official one

conan/recipes/triangle/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+    homepage = "https://www.cs.cmu.edu/~quake/triangle.html"
+    topics = ("triangle", "delaunay", "triangulation", "mesh")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {"shared": [True, False]}
+    default_options = {"shared": False}
+
+    def validate(self):
+        if self.options.shared:
+            raise ConanInvalidConfiguration("triangle does not support shared libraries")
```
KoenBussemakerDeltares, line 35: If we don't support it we might as well remove the option altogether

conan/recipes/triangle/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+        return load(
+            self,
+            os.path.join(self._triangle_source_folder, "triangle_license_info.txt"),
+        )
+
+    def package(self):
+        save(self, os.path.join(self.package_folder, "licenses", "LICENSE"), self._extract_license())
+        copy(self, "triangle.h", src=self._triangle_source_folder, dst=os.path.join(self.package_folder, "include"))
+        copy(self, "*.a", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
```
KoenBussemakerDeltares: Doesn't triangle itself provide an adequate installation method in CMake?

conan/recipes/triangle/config.yml
```diff
@@ -0,0 +1,3 @@
+versions:
+  "1.6-deltares.1":
```
KoenBussemakerDeltares: What is the logic behind this versioning scheme?

conan/recipes/triangle/all/conanfile.py
harmenwierenga: I did not have this at first, but Conan requires the shared option to be defined for 'Library' type packages

conan/recipes/triangle/all/conanfile.py
harmenwierenga: Since I now took full control over the cmake, I have moved the install rules there

conan/recipes/triangle/config.yml
harmenwierenga: 1.6 is the official version. We have changed that source code, so I have added 'deltares' to make it clear that it is not the unpatched official version. Then, we still need a version number that we can increment if we make changes to it, otherwise we get the same package reference in conan.

conan/recipes/triangle/all/conanfile.py
harmenwierenga: Done

KoenBussemakerDeltares (approved):
Another package in conan! Nice job!

src/test/engines_gpl/dflowfm/packages/dflowfm_kernel/CMakeLists.txt
```diff
@@ -4,94 +4,51 @@ file(GLOB source src/*.f90)
 
 add_executable(${executable_name} ${source})
 
+target_link_libraries(${executable_name} PRIVATE
```
rene-deltares, line 7: Nice! were there differences between windows and linux here in the first place?

harmenwierenga: You can see that there was one additional one on Linux, Threads::Threads, that one is linked separately. Most of these are probably not necessary, but I won't check that as part of this PR

ci/teamcity/Delft3D/linux/conanPackages.kt
```diff
@@ -25,13 +25,13 @@ object LinuxConanPackages : BuildType({
...
     name = "Conan packages"
     buildNumberPattern = "%build.vcs.number%"
-
     allowExternalStatus = true
 
     params {
         param("reverse.dep.${LinuxBuildTools.id}.intel_oneapi_version", "2024")
         param("nexus_conan_username", DslContext.getParameter("nexus_conan_username"))
         password("nexus_conan_password", DslContext.getParameter("nexus_conan_password"))
+        param("conan_build_option", "--build-missing")
```
rene-deltares, line 34: If the cache is empty, is there a difference between --build-missing and --rebuild-packages? I think the cache is always empty in this build right? Or have we / are we planning on populating it?

rene-deltares (approved):
Very nice trick to use our own repo in a Conan package. This way at least we don't introduce any new changes. Hopefully we can get rid of this dependency eventually.

ci/teamcity/Delft3D/linux/conanPackages.kt
harmenwierenga: The difference is that --rebuild-packages will ensure that it only looks at the local-recipes remote, not at the nexus remote, while --build-missing will also look at nexus. This is a feature in our run_conan.py. This means that --build-missing will first download the existing packages from nexus and only rebuild the ones that were not found there, and then upload them. --rebuild-packages will always have to build them from scratch, because it only has access to the recipes

PR #1113, opened 2026-07-22
UNST-10151: Write total (instead of background) viscosity to map-file

JulienGroenenboom:
# What was done 
- Write total (instead of background) viscosity to map-file. This bug only occurs when HisInterval > 0
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10151


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/fill_valobs.f90
```diff
@@ -118,9 +118,9 @@ subroutine fill_valobs()
       real(kind=dp), allocatable :: ship_level(:)
       real(kind=dp), allocatable :: cell_z_centers(:)
       real(kind=dp), allocatable :: waq_tmp(:)
-      real(kind=dp), allocatable :: vius(:) !< Flowlink-averaged horizontal viscosity (viu) at s-point
-      integer      , allocatable :: wet_or_dry(:)
-      
+      real(kind=dp), allocatable, save :: vius(:) !< Flowlink-averaged horizontal viscosity (viu) at s-point
```
KoenBussemakerDeltares: As discussed in person, a local save variable is problematic because it can't be reset between runs. This is not really a problem for regular dflowfm runs since that is just a separate process that gets shut down after the run. The unit tests however run sequentially within the same process, and in-between-cleanup is important to ensure that we always start with a clean slate. That is impossible with local save variables since only the enclosing function has access to them.

To resolve this, turn `vius` into a module variable (the `save` keyword is reduntant since it will now be a global variable). Make sure `resetFullFlowModel` resets it to its original state, which is just a matter of deallocating it if it was allocated.

KoenBussemakerDeltares (approved):
Looks good with `vius` as a module variable. The other cleanups are nice too, much cleaner this way.

PR #1116, opened 2026-07-22
All/task/unst 10014 auto generate code for python wrapper

MAfarrag:
# What was done

This branch (UNST-10014) makes the D-Flow FM **Python wrapper auto-generated** from the single
sources of truth — the C header `dflowfm_io_api.h` and `json/mdu.json` — so the typed Python API can
never drift from the C ABI or the MDU schema.

## Auto-generated Python wrapper (`wrappers/python/`)

The old single-file `wrappers/python/dflowfm_io.py` is replaced by a proper `dflowfm_io` package
(`src/dflowfm_io/`, with `base/` and `mdu/` subpackages). The two typed layers are generated by two
scripts under `wrappers/python/scripts/`:

- `generate_model.py` (from the C header) → `mdu/model.py`: the typed **Layer-1 `MduModel`**, one
  get/set method per `mdu_get_*`/`mdu_set_*` function, with per-type value marshalling.
- `generate_schema.py` (from `json/mdu.json`) → `mdu/schema.py`: the typed **Layer-2 `MduSchema`** —
  one class per INI section, one typed property per keyword (22 sections, 703 properties) — plus a
  `KNOWN_PROPERTIES` registry of every dotted key, for schema-existence queries by consumers such as
  hydrolib-core.

Everything else is hand-written, including `base/bindings.py` (a small, rarely-changing ctypes ABI
mirror — `argtypes`/`restype`, the `mdu_severity_t` constants, the `mdu_issue_t` struct), the
`MduDocument` lifecycle, and the `Issue`/`Severity` value types.

**Generated files are not committed.** `mdu/model.py` and `mdu/schema.py` are git-ignored and
produced by the build — consistent with the native library, which is also built and staged, never
committed. The CMake `dflowfm_io_bindings` target regenerates them (it depends on the header, the
JSON, and the generator scripts); a golden test asserts *on-disk == fresh regeneration*, and both
generators fail closed on any C signature or `mdu.json` value type they do not recognise.

## Enum handling (consumes the base branch's split-enum C ABI)

The base branch (UNST-10203) split the MDU enum C ABI into two explicit accessor pairs:
`mdu_get_string_enum`/`mdu_set_string_enum` (by **name**) and `mdu_get_int_enum`/`mdu_set_int_enum`
(by **value**). The generated Python wrapper targets these directly:

- Layer-1 `MduModel` exposes `get_string_enum`/`set_string_enum` and `get_int_enum`/`set_int_enum`.
- Layer-2 `MduSchema` surfaces string enums (which carry labels) **by name** (`str`) and intenums
  **by integer** — driven by the `enum` vs `intenum` value type in `mdu.json`.

The int↔name mapping stays in the C++ core; it is not duplicated in the wrapper.

## Build / packaging

- The native library is staged into the wheel package (`_lib/`) by a dedicated
  `dflowfm_io_stage_lib` target that depends on `dflowfm_io_api`. (It is a dependent target rather
  than a `POST_BUILD` step because CMake forbids attaching `POST_BUILD` to a target defined in
  another directory.) The staging script `stage_native_library.cmake` and the target now live in
  `wrappers/python/`, next to the wheel they feed.
- A `dflowfm_io_wheel` target produces a platform-tagged wheel that bundles the native library, so
  hydrolib-core can `pip install` a self-contained artifact.
- The wheel is **not** part of the default build. It is gated by the CMake option
  `DFLOWFM_IO_BUILD_WRAPPER_PACKAGES` (declared centrally in `dflowfm_io/cmake/options.cmake`,
  default **OFF**), a wrapper-agnostic switch intended to also gate the C# NuGet package later.

## Upstream integration

The branch is merged up to date with the base (`UNST-9973`, including `UNST-10203`/`UNST-10080`) and
the wrapper is adapted to its schema and split-enum ABI.

# How to build

The `dflowfm_io` subproject also builds standalone (outside the Delft3D tree). From
`src/utils_lgpl/dflowfm_io`:

```bash
# 1. Configure once (standalone; requires CMake >= 3.19, a C++20 compiler, Python 3)
cmake -S <path-to>/src/utils_lgpl/dflowfm_io \
      -B <path-to>/src/utils_lgpl/dflowfm_io/out/build/x64-Release \
      -G "Visual Studio 17 2022" -A x64

# 2. Build the native library (staged into wrappers/python/src/dflowfm_io/_lib by dflowfm_io_stage_lib)
cmake --build <build-dir> --target dflowfm_io_api --config Release

# 3. Build the Python wheel ON DEMAND (also builds the DLL and regenerates model.py/schema.py from
#    the header + mdu.json first, then runs `pip wheel`)
cmake --build <build-dir> --target dflowfm_io_wheel --config Release
```

**The Python wheel is NOT part of the default build.** The option `DFLOWFM_IO_BUILD_WRAPPER_PACKAGES`
defaults to **OFF** (the wheel is slow and only needed when publishing for hydrolib-core), so a bare
`cmake --build` builds the DLL but not the wheel. Build the `dflowfm_io_wheel` target explicitly
(step 3), or opt it into the default build with `-D DFLOWFM_IO_BUILD_WRAPPER_PACKAGES=ON`. CI that is
expected to publish the wheel must build that target — a default build will not produce it.

Running the Python wrapper tests (they use the native library staged in `_lib/`, so build first):

```bash
<python> -m pytest src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/tests
```

# Evidence of the work done

- [ ]	Video/figures \
- [ ]	Not applicable

# Tests
- [x] Tests updated \
  115 wrapper tests pass. Includes dedicated generator unit tests under
  `wrappers/python/tests/scripts/` (header parsing / accessor rendering for `generate_model.py`,
  schema rendering for `generate_schema.py`), a golden test asserting the generated files equal a
  fresh regeneration, and Layer-1/Layer-2 coverage of the split-enum accessors (string enums by
  name, intenums by integer).
- [ ]	Not applicable

# Documentation
- [x]	Documentation updated \
  `wrappers/python/README.md` documents the generated code (which modules are generated, by which
  script, and how to regenerate them manually and via the CMake `dflowfm_io_bindings` target).
- [ ]	Not applicable

# Issue link

- UNST-10014 — https://issuetracker.deltares.nl/browse/UNST-10014


src/utils_lgpl/dflowfm_io/dflowfm_io_api/include/dflowfm_io_api/dflowfm_io_api.h
```diff
@@ -52,6 +52,7 @@ DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_get_string(mdu_handle_t handle, co
 DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_get_path(mdu_handle_t handle, const char* key, const char** path_out);
 DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_get_datetime(mdu_handle_t handle, const char* key, int64_t* epoch_out);
 DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_get_enum(mdu_handle_t handle, const char* key, int32_t* enum_out);
+DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_get_enum_name(mdu_handle_t handle, const char* key, const char** name_out);
```
MAfarrag: expose enum names, so that,  doc.schema.time.tUnit is shown as "S" not 3

src/utils_lgpl/dflowfm_io/dflowfm_io_api/dflowfm_io_api.cpp
```diff
@@ -251,6 +252,41 @@ dflowfm_io_result_t mdu_get_enum(mdu_handle_t handle, const char* key, int32_t*
     });
 }
 
+dflowfm_io_result_t mdu_get_enum_name(mdu_handle_t handle, const char* key, const char** name_out)
```
MAfarrag: add a name <-->int conversion

src/utils_lgpl/dflowfm_io/dflowfm_io_api/dflowfm_io_api.cpp
```diff
@@ -379,6 +415,32 @@ dflowfm_io_result_t mdu_set_enum(mdu_handle_t handle, const char* key, int32_t e
     });
 }
 
+dflowfm_io_result_t mdu_set_enum_name(mdu_handle_t handle, const char* key, const char* name)
```
MAfarrag: add a name <-->int conversion

src/utils_lgpl/dflowfm_io/dflowfm_io_api/test/CMakeLists.txt
```diff
@@ -1,13 +1,12 @@
 enable_testing()
 
-cmake_path(SET PYTHON_TEST_DIR NORMALIZE "${CMAKE_CURRENT_SOURCE_DIR}/../wrappers/python/test")
+cmake_path(SET PYTHON_TEST_DIR NORMALIZE "${CMAKE_CURRENT_SOURCE_DIR}/../wrappers/python/tests")
 find_package(Python3 COMPONENTS Interpreter)
 if(Python3_FOUND)
+    # The wrapper loads the DLL from its bundled _lib/, which the dflowfm_io_api POST_BUILD step
+    # stages after every compile, so the test needs no library-location hint.
```
RStolkerDeltares: These comments are a bit too verbose and don't add much value. I'd suggest removing them to keep the code cleaner.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/include/dflowfm_io_api/dflowfm_io_api.h
RStolkerDeltares: I'm not really in favor of adding additional functions for getting/setting enum values. We designed enum handling in the io library such that enums are always treated as integers e.g. `intenum` values in their corresponding integer representation as defined in the mdu.json and `enum` values as the index of the corresponding string value in the mdu.json.

The original design idea was to have the language specific wrappers either expose typed enums or to provide mappings between integer values and their string representations. These typed enums or mappings would need to be generated based on the mdu.json.

The C# wrapper currently generates enum metadata which the UI uses for the integer <-> string mapping. And as far as I know the integer representation is also sufficient for the Fortran kernel but @KoenBussemakerDeltares can comment on that in more detail.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/scripts/generate_bindings.py
```diff
@@ -0,0 +1,367 @@
+"""Generate the Python wrapper's ABI-coupled code from the dflowfm_io_api C header.
```
RStolkerDeltares: I don't think the Python bindings should be generated. The C API is intentionally small and stable. Even if it changes in the future, it's straightforward to maintain the python binding manually. This generator introduces more complexity than the Python bindings themselves. Note that for the same reason the C# bindings are also manually defined on purpose.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/src/dflowfm_io/base/bindings.py
```diff
@@ -0,0 +1,87 @@
+"""GENERATED from dflowfm_io_api.h by scripts/generate_bindings.py - do not edit.
```
RStolkerDeltares: Generated code should not be committed. Instead it should be generated as part of the build, similar to how `MduSchemaGenerated.h` is handled and marked as `GENERATED` in CMake.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/CMakeLists.txt
```diff
@@ -25,6 +25,21 @@ install(TARGETS dflowfm_io_api
     ARCHIVE DESTINATION lib
 )
 
+# Stage the freshly built native library into the Python wheel package (_lib/) so `pip wheel`
+# bundles it. The library is a build artifact: it is git-ignored there and only ever produced by
+# the build, never committed.
```
RStolkerDeltares: These comments are a bit too verbose and don't add much value. I'd suggest removing them to keep the code cleaner.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/stage_native_library.cmake
```diff
@@ -0,0 +1,22 @@
+# Stage a freshly built native library into the Python wheel package's _lib/ directory.
```
RStolkerDeltares, line 1: This CMake script file is specific to the python wrapper and should rather be placed in the python wrapper subdir in it's own `cmake/` folder:
`dflowfm_io_api/wrappers/python/cmake/stage_native_library.cmake`

src/utils_lgpl/dflowfm_io/dflowfm_io_api/CMakeLists.txt
```diff
@@ -25,6 +25,21 @@ install(TARGETS dflowfm_io_api
...
+	"${CMAKE_CURRENT_SOURCE_DIR}/wrappers/python/src/dflowfm_io/_lib")
+
+add_custom_command(TARGET dflowfm_io_api POST_BUILD
+	COMMAND ${CMAKE_COMMAND}
+	-DLIB_DIR=${DFLOWFM_IO_PY_WHEEL_LIB_DIR}
+	-DLIB_FILE=$<TARGET_FILE:dflowfm_io_api>
+	-P ${CMAKE_CURRENT_SOURCE_DIR}/stage_native_library.cmake
+	COMMENT "Staging dflowfm_io_api into the Python wheel package (_lib/)"
+	VERBATIM
+)
```
RStolkerDeltares: It would be cleaner to have this in `wrappers/python/CMakeLists.txt` next to the other Python build steps. I know `POST_BUILD` must be in the target's directory scope but perhaps it can be done with a dedicated staging target that depends on `dflowfm_io_api` and invokes `stage_native_library.cmake`?

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+# on demand:
+#     cmake --build <build-dir> --target dflowfm_io_wheel
+# or include it in the default build by configuring with:
+#     -D DFLOWFM_IO_BUILD_PYTHON_WHEEL=ON
+
+option(
+	DFLOWFM_IO_BUILD_PYTHON_WHEEL
+	"Build the dflowfm_io Python wheel as part of the default build"
+	OFF
+)
```
RStolkerDeltares: This option should go to high-level options file `/dflowfm_io/cmake/options.cmake` and be given a more general name `DFLOWFM_IO_BUILD_WRAPPER_PACKAGES` since it should also enable/disable building the NuGet package for the C# wrapper.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+
+# Regenerate the ABI-coupled Python code from the C header, so it can never drift from the
```
RStolkerDeltares: These comments (and below) are a bit too verbose and don't add much value. I'd suggest removing them to keep the code cleaner.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/README.md
```diff
@@ -0,0 +1,77 @@
+# dflowfm_io
```
RStolkerDeltares, line 1: There's also a README file in the root dir (`dflowfm_io/README.md`) and it would be good to have this documentation either merged into that file or at least have the part of building the wheel in there.

src/utils_lgpl/dflowfm_io/dflowfm_io_api/include/dflowfm_io_api/dflowfm_io_api.h
```diff
@@ -63,6 +64,7 @@ DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_set_string(mdu_handle_t handle, co
 DFLOWFM_IO_API_EXPORT dflowfm_io_result_t mdu_set_path(mdu_handle_t handle, const char* key, const char* value);
```
MAfarrag, line 64: revert

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/fortran/dflowfm_io.f90
```diff
@@ -40,6 +40,7 @@ module dflowfm_io
         procedure :: get_path         => mdu_get_path_f
         procedure :: get_datetime     => mdu_get_datetime_f
         procedure :: get_enum         => mdu_get_enum_f
+        procedure :: get_enum_name    => mdu_get_enum_name_f
```
MAfarrag: revert

src/utils_lgpl/dflowfm_io/dflowfm_io_api/include/dflowfm_io_api/dflowfm_io_api.h
MAfarrag: done

src/utils_lgpl/dflowfm_io/dflowfm_io_api/include/dflowfm_io_api/dflowfm_io_api.h
MAfarrag: reverted

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/src/dflowfm_io/base/bindings.py
MAfarrag: now this is not generated anymore (the generate script that was 400 line is removed), and this module now is hard writted and committed

src/utils_lgpl/dflowfm_io/dflowfm_io_api/CMakeLists.txt
MAfarrag: done

src/utils_lgpl/dflowfm_io/dflowfm_io_api/test/CMakeLists.txt
MAfarrag: done

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/scripts/generate_bindings.py
MAfarrag: removed

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/src/dflowfm_io/base/bindings.py
MAfarrag: generated files now are untracked and ignored in the .gitignore

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/CMakeLists.txt
MAfarrag: removed

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/CMakeLists.txt
MAfarrag: now there is a `dflowfm_io/cmake/options.cmake`  that has the switch on/off for building the wrappers, switches for the csharp and other wrappers also can be added there

src/utils_lgpl/dflowfm_io/dflowfm_io_api/CMakeLists.txt
MAfarrag: python wrapper specific cmake logic is moved to the cmake file inside the python wrapper dir

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/stage_native_library.cmake
MAfarrag: done

src/utils_lgpl/dflowfm_io/dflowfm_io_api/wrappers/python/README.md
MAfarrag: this is the only remaining comment, what exactly should I move to the dflowfm_io/readmd.md and what should i leave in the wrappers/python/readme.md

RStolkerDeltares: It would be nice to just have a short line about building the Python wheel in the dflowfm_io/readmd.md file.

RStolkerDeltares (approved):

PR #1121, opened 2026-07-23
restore mkl finder previously in the petsc cmake. Now check for any m…

FlorisBuwaldaDeltares:
…kl library (not a hardcoded number). Error on multiple found. Add mkl as a dflowfm kernel dependency (functionally equivalent to petsc dependency)

# What was done 

- restore intel mkl imported implib definition in intel_redist
- add dependency on mkl for dflowfm kernel (formerly this was petsc)
- flexible in mkl number, as long as there's only 1 dll in the bin folder.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/CMakeLists.txt
```diff
@@ -198,7 +198,7 @@ if (WIN32)
     if(NOT TARGET impi)
       message(FATAL_ERROR "no valid intel MPI target found, please check your installation.")
     endif()
-    target_link_libraries(${library_name} PRIVATE impi)
+        target_link_libraries(${library_name} PRIVATE impi mkl_sequential)
```
harmenwierenga: Some spurious whitespace

src/third_party_open/intelredist/intel_redist_windows.cmake
```diff
@@ -28,6 +28,30 @@ install(FILES
     ${redist_path}/libiomp5md_db.dll
     DESTINATION bin CONFIGURATIONS debug
 )
+
+if(NOT TARGET mkl_sequential)
+    add_library(mkl_sequential SHARED IMPORTED GLOBAL)
+
+    file(TO_CMAKE_PATH "$ENV{ONEAPI_ROOT}" oneapi_root_cmake)
```
harmenwierenga, line 35: Would it be useful to add an error message if ONEAPI_ROOT cannot be found?

FlorisBuwaldaDeltares: i was under the assumption that we already check for the existence of ONEAPI_ROOT many other times, and also already earlier in our cmake., so that at this point we may assume it exists. If this is indeed the only place that still relies on ONEAPI_ROOT then I should add an error

src/engines_gpl/dflowfm/packages/dflowfm_kernel/CMakeLists.txt
FlorisBuwaldaDeltares: oops

src/third_party_open/intelredist/intel_redist_windows.cmake
FlorisBuwaldaDeltares: I have added a check on the existence of ONEAPI_ROOT in the main cmakelists.txt

src/engines_gpl/dflowfm/packages/dflowfm_kernel/CMakeLists.txt
FlorisBuwaldaDeltares: fixed

harmenwierenga (approved):

PR #1122, opened 2026-07-23
UNST-10106 Fixed duplicate keywords in dflowfm kernel's write mdu logic

KoenBussemakerDeltares:
# What was done 

When working on dflowfm_io-related tests I noticed that the dflowfm kernel always writes certain duplicate properties when writing to an mdu file. `Numlimdt_baorg` is always written twice, `wrihis_runupgauge` is always written three times. Just open any .dia file to confirm this.

image

image

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
@@ -890,10 +890,10 @@ contains
                              'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
       call add_output_config(config_set_his, IDX_HIS_RUG_RUX, &
                              'Wrihis_runupgauge', 'rug_x_coordinate', 'time-varying x-coordinate of shoreline position', '', &
-                             'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
+                             'm', UNC_LOC_RUG)
       call add_output_config(config_set_his, IDX_HIS_RUG_RUY, &
                              'Wrihis_runupgauge', 'rug_y_coordinate', 'time-varying y-coordinate of shoreline position', '', &
-                             'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
+                             'm', UNC_LOC_RUG)
 
       !
       ! HIS: hydraulic structures
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
@@ -3152,9 +3152,7 @@ contains
       if (writeall .or. locsaltmax /= 10.0_dp) then
          call prop_set(prop_ptr, 'numerics', 'LocSaltMax', locsaltmax, 'maximum salinity for case of lock exchange')
       end if
-      if (writeall .or. numlimdt_baorg > 0) then
-         call prop_set(prop_ptr, 'numerics', 'Numlimdt_baorg', Numlimdt_baorg, 'if previous numlimdt > Numlimdt_baorg keep original cell area ba in cutcell')
-      end if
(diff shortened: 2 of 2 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -3152,9 +3152,7 @@ subroutine writeMDUFilepointer(mout, writeall, istat)
       if (writeall .or. locsaltmax /= 10.0_dp) then
          call prop_set(prop_ptr, 'numerics', 'LocSaltMax', locsaltmax, 'maximum salinity for case of lock exchange')
       end if
-      if (writeall .or. numlimdt_baorg > 0) then
-         call prop_set(prop_ptr, 'numerics', 'Numlimdt_baorg', Numlimdt_baorg, 'if previous numlimdt > Numlimdt_baorg keep original cell area ba in cutcell')
-      end if
+      call prop_set(prop_ptr, 'numerics', 'Numlimdt_baorg', Numlimdt_baorg, 'if previous numlimdt > Numlimdt_baorg keep original cell area ba in cutcell')
```
FlorisBuwaldaDeltares, line 3155: you seem to have removed the if (writeall .or. numlimdt_baorg > 0) check, which ensured that this value was only written to the dia file if the value was not default. is that intentional?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -890,10 +890,10 @@ subroutine default_fm_statistical_output()
                              'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
       call add_output_config(config_set_his, IDX_HIS_RUG_RUX, &
                              'Wrihis_runupgauge', 'rug_x_coordinate', 'time-varying x-coordinate of shoreline position', '', &
-                             'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
+                             'm', UNC_LOC_RUG)
       call add_output_config(config_set_his, IDX_HIS_RUG_RUY, &
                              'Wrihis_runupgauge', 'rug_y_coordinate', 'time-varying y-coordinate of shoreline position', '', &
-                             'm', UNC_LOC_RUG, description='Write run-up gauge statistics to his-file')
```
FlorisBuwaldaDeltares, line 896: huh, i would expect you to remove two rug writing calls from writeMDUFilepointer, how does removing these descriptions from the output_config fix the triplicate value in the dia file???

FlorisBuwaldaDeltares (commented):
two questions

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
KoenBussemakerDeltares: That is indeed intentional. This value was written twice, once conditional and once always. I've removed the conditional version and moved the always-version up to keep it closely grouped to `baorgfracmin` in the MDU output.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
KoenBussemakerDeltares: That's because `add_output_config` is a bit silly. It writes the property only if a description is provided. For other values (e.g. `Wrihis_turbulence` or `Wrihis_waves`) you'll see that only one entry has a description. But `wrihis_runupgauge` had a description for all three, so it ended up being written three times.

This mechanism makes no sense and feels very hacky, but I didn't want to blow up the scope any further by refactoring it.

arthurvd: "Silly", "makes no sense", "hacky". Funny how words seem to carry more weight when written down than when spoken in person.

FlorisBuwaldaDeltares (approved):

PR #1124, opened 2026-07-23
UNST-9637 refactor wave initialization

FlorisBuwaldaDeltares:
# What was done 

- Added centralized Wave-7 input requirements derived from active wave processes.
- Current-format external forcing now reads only required wave quantities; inactive fields retain neutral defaults.
- Added validation that reports missing required wave providers at initialization.
- Made direct radiation-stress-only configurations work without unnecessary height/direction input.
- Wave-7 + wavebreakerturbulence is now supported
- radiation stress (waveforce xy only) no longer requires waveperiod just for the force limiter (force becomes unlimited!)
- hwavcom and twavcom are now read-only for FM kernel
- Added focused unit coverage for the offline-wave requirement combinations. 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1534,6 +1534,13 @@ subroutine readMDUFile(filename, istat)
       end if
 
       call prop_get(md_ptr, 'waves', '3Dwavebreakerturbulence', jawavebreakerturbulence) ! Add wave-induced production terms in turbulence modelling: 0 = no, 1 = yes
+      if (kmx <= 1) then
+         jawavebreakerturbulence = WAVE_BREAKER_TURB_OFF ! turn off 3D-only setting
+      end if
+      if (jawave == WAVE_NC_OFFLINE .and. jawavebreakerturbulence > WAVE_BREAKER_TURB_OFF) then
+         call mess(LEVEL_ERROR, 'Wavemodelnr = 7 does not support 3Dwavebreakerturbulence. Set 3Dwavebreakerturbulence = 0.')
```
jreyns, line 1594: Where is this coming from? This should be a user decision

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1556,6 +1563,15 @@ subroutine readMDUFile(filename, istat)
          modind = 0
       end if
 
+      if (jawave == WAVE_NC_OFFLINE) then
+         offline_wave_input_requirements = get_offline_wave_input_requirements(waveforcing, jawaveforces, jawaveStokes, &
+                                                                                jawavestreaming, jawavedelta, &
+                                                                                modind > 0 .and. ftauw > 0.0_dp, &
+                                                                                flow_without_waves)
```
jreyns, line 1623: double check this flow_without_waves req

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
```diff
@@ -1304,28 +1303,8 @@ subroutine set_wave_modelling()
...
-                               & distot, dsurf, dwcap, &
-                               & ndx, 1, hwav, twav, &
-                               & ag, .true., waveforcing, &
-                               & JONSWAPgamma0, sbxwav, sbywav, ierror)
-         end if
-         !
-         call wave_uorbrlabda()
+         call compute_wave_parameters()
          if (kmx == 0) then
-            call wave_comp_stokes_velocities()
```
adrimourits, line 1399: As a result, line 52: `use m_wave_comp_stokes_velocities, only: wave_comp_stokes_velocities` can be removed

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_waveconst.f90
```diff
@@ -20,6 +20,17 @@ module m_waveconst
    integer, parameter :: WAVEFORCING_DISSIPATION_TOTAL = 2
    integer, parameter :: WAVEFORCING_DISSIPATION_3D = 3
 
+   ! Offline wave input quantities. Individual vector components have separate
+   ! flags so initialization can report exactly which provider is missing.
```
adrimourits, line 24: Can you add a comment line with something like `By using powers of 2 as values, parameters can be selected using bitwise masking`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -511,7 +511,7 @@ subroutine readMDUFile(filename, istat)
       use m_fm_icecover, only: fm_ice_read
       use m_f1dimp, only: f1dimppar
       use m_sediment
-      use m_waves, only: hwavuni, twavuni, phiwavuni
+      use m_waves, only: hwavuni, twavuni, phiwavuni, ftauw, offline_wave_input_requirements
```
adrimourits: Not only `ftauw`, but also `fwfac` and `fbreak` need to be added here, I think.

Did that work as expected? I can't imagine. Why did it compile? Why dit CoPilot only add `ftauw`?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
```diff
@@ -94,48 +95,19 @@ subroutine compute_wave_parameters()
...
-      ! SWAN
-      if ((jawave == WAVE_SWAN_ONLINE .or. jawave == WAVE_NC_OFFLINE) .and. .not. flow_without_waves) then
-         if (jawave == WAVE_NC_OFFLINE) then
-            ! HSIG is read from SWAN NetCDF file. Convert to HRMS
-            hwav = hwavcom / sqrt2_hp
-         else
-            hwav = hwavcom
-         end if
-         hwav = min(hwav, gammax * hs)
+      ! Online SWAN coupling
```
adrimourits, line 98: This reorganisation of logics is nice. But it's difficult to check whether the behaviour is the same. I expect that not all options are properly covered by tests. So, do we have enough confidence in CoPilot to accept this? I prefer that @jreyns at least have a look at the changes in this file.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
```diff
@@ -444,33 +444,56 @@ end subroutine fill_open_boundary_cells_with_inner_values_all
...
          ki = link2cell(2, link)
-         hwavcom(kb) = hwavcom(ki)
-         twav(kb) = twav(ki)
-         phiwav(kb) = phiwav(ki)
-         uorbwav(kb) = uorbwav(ki)
-         sxwav(kb) = sxwav(ki)
-         sywav(kb) = sywav(ki)
-         mxwav(kb) = mxwav(ki)
-         mywav(kb) = mywav(ki)
+         if (wave_input_is_required(offline_wave_input_requirements, WAVE_INPUT_SIGNIFICANT_HEIGHT)) then
```
adrimourits: This long list of if-statements inside the do loop might cause a performance issue. But there is a workaround (define all parameters). Do you think it's worth to put a do-loop inside each if-statement? I'm not a fan of that solution.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field.f90
```diff
@@ -329,4 +329,101 @@ subroutine test_validate_nonlegacy_operand_does_not_warn() bind(C)
    end subroutine test_validate_nonlegacy_operand_does_not_warn
    !$f90tw)
 
+   !$f90tw TESTCODE(TEST, test_init_spatial_field, test_offline_wave_requirements_radiation_stress, test_offline_wave_requirements_radiation_stress,
```
adrimourits, line 332: Nice UnitTests; no complex initialization needed, no allocation issues.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_offline_wave_external_forcing.f90
```diff
@@ -0,0 +1,232 @@
...
+
+      call create_scalar_bc_file(PERIOD_BC_FILE, 'waveperiod', 's', 4.0_dp, 6.0_dp)
+      call create_scalar_bc_file(FORCE_X_BC_FILE, 'xwaveforce', 'N/m2', 1.0_dp, 3.0_dp)
+      call create_scalar_bc_file(FORCE_Y_BC_FILE, 'ywaveforce', 'N/m2', -2.0_dp, 2.0_dp)
+      call create_offline_wave_ext_file(POSITIVE_EXT_FILE, include_y_force=.true.)
+      threshold_abort = LEVEL_FATAL
+      call setup_offline_radiation_stress_case(POSITIVE_EXT_FILE)
+
+      iresult = DFM_NOERR
+      call init_new(iresult)
```
adrimourits, line 174: `init_new` is doing allocations. I expect some cleaning is needed at the end of this test. Also in next test.

adrimourits (approved):
Nice

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_waveconst.f90
FlorisBuwaldaDeltares: will do!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
FlorisBuwaldaDeltares: hmm excellent question, I will investigate

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
FlorisBuwaldaDeltares: maybe it means that this is a good moment to cover all options with tests? in my experience we shouldn't blindly trust copilot's code if there's gaps in what's tested haha

FlorisBuwaldaDeltares: Do you have an idea of what's currently tested and where the gaps are?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
FlorisBuwaldaDeltares: hmm good point, hadn't thought of that. My first instinct is: this doesn't happen every timestep, only during an external forcing update. I think the extra conditionals will not cause any significant performance issues.

Additionally if i make offline_wave_input_requirements an intent(in) dummy argument, the compiler knows it is invariant and can hoist the conditional out of the loop. That might be worth a commit.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
adrimourits: Finding out what tests are there is not trivial. My first guess is that the only tests related to using a SWAN output file directly as input for FM are the two testcases in e02_f041_SWAN-tau_to_WAQ.

FlorisBuwaldaDeltares: SWAN-tau_to_WAQ tests flow_without_waves, i know that as i had to fix them when they were broken.

Additionallym, the fm testbench contains e02_f044_c01_waveforcing_1, 2 and 3. , which test wavemodel = 7 and waveforcing is 1, 2 and 3 respectively.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
```diff
@@ -180,6 +152,61 @@ subroutine compute_wave_parameters()
...
+            call wave_uorbrlabda()
+         else
+            hwav = 0.0_dp
+            twav = 0.0_dp
+            mxwav = 0.0_dp
+            mywav = 0.0_dp
+            sbxwav = 0.0_dp
+            sbywav = 0.0_dp
+
+            if (wave_input_is_required(offline_wave_input_requirements, WAVE_INPUT_PERIOD)) then
```
jreyns, line 203: why would you need this?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
```diff
@@ -722,21 +722,7 @@ end subroutine set_all_wave_parameters
    !> Read only the offline wave quantities required by the active configuration.
    subroutine set_offline_wave_parameters()
 
-      hwavcom(:) = 0.0_dp
```
jreyns: Put this back please, this is important for flooding and drying

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
```diff
@@ -730,75 +719,37 @@ subroutine set_all_wave_parameters()
 
    end subroutine set_all_wave_parameters
 
-!> set wave parameters for jawave == 7 (offline wave coupling) and waveforcing == 1 (wave forces via radiation stress)
-   subroutine set_parameters_for_radiation_stress_driven_forces()
```
jreyns, line 783: these reinitialisations are important to get correct model results with flooding and drying. There is no knowledge in the external wave model about the state of the flow model. I don't want stale values somewhere in my grid because it happens not to be covered by a new external field. Please put them back

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
jreyns: these can be replaced with m_waveconst

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
```diff
@@ -180,6 +152,61 @@ subroutine compute_wave_parameters()
...
+               forcing_for_transformation = waveforcing
+            end if
+
+            call transform_wave_physics_hp(hwavcom, phiwav, twavcom, hs, &
+                                           sxwav, sywav, mxwav, mywav, &
+                                           distot, dsurf, dwcap, &
+                                           ndx, 1, hwav, twav, &
+                                           ag, .true., forcing_for_transformation, &
+                                           JONSWAPgamma0, sbxwav, sbywav, ierror)
+            ! Wave-7 receives a wave field already computed by the external
```
jreyns: hwav = min(hwav, gammax * hs) needed here

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
```diff
@@ -94,48 +95,19 @@ subroutine compute_wave_parameters()
...
 
-      ! SWAN
-      if ((jawave == WAVE_SWAN_ONLINE .or. jawave == WAVE_NC_OFFLINE) .and. .not. flow_without_waves) then
-         if (jawave == WAVE_NC_OFFLINE) then
-            ! HSIG is read from SWAN NetCDF file. Convert to HRMS
-            hwav = hwavcom / sqrt2_hp
-         else
-            hwav = hwavcom
-         end if
-         hwav = min(hwav, gammax * hs)
```
jreyns, line 105: we need this, probably after transform wave physics, but for sure for all the online swan couplings. this is called every flow timestep. so first one is there, that is okay, but I think it is missing from the offline branch. hwavcom should be read with dtuser, but transformed and limited every dts

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
```diff
@@ -4459,23 +4462,28 @@ function fm_ext_force_name_to_ec_item(trname, sfname, waqinput, constituent_name
...
          map_write_settings%wav_twav = 1
+         call register_offline_wave_input_provider(WAVE_INPUT_PERIOD)
       case ('dir', 'wavedirection')
          itemPtr1 => item_dir
          dataPtr1 => phiwav
          map_write_settings%wav_phiwav = 1
+         call register_offline_wave_input_provider(WAVE_INPUT_DIRECTION)
          ! wave height needed as the weighting factor for direction interpolation
          itemPtr2 => item_hrms
          dataPtr2 => hwavcom
```
jreyns, line 4518: call register... for wave height not needed?

src/utils_lgpl/deltares_common/packages/deltares_common/src/transform_wave_physics.f90
jreyns, line 124: if you define the subroutine, why not use it? I am also puzzled as to why we need this period conversion in a default branch where everything is set to zero

jreyns (changes requested):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
FlorisBuwaldaDeltares: I'm no expert but in subroutine "wave_fillsurdis" which is called with WAVE_NC_OFFLINE and jawavebreakerturbulence on does not have a branch handling the WAVE_NC_OFFLINE case. copilot concluded this is an implementation gap and it's best to disable the combination for now.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_update.f90
FlorisBuwaldaDeltares: In a previous iteration i reset all wave parameters during an external forcings update. However testcases relied on this, as the references seemed to need the stale values remaining. I'm open with doing a full reset but then consistently, which will involve updating references.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
FlorisBuwaldaDeltares: I had it, but then my testcases were red. The conclusion was we cannot limit hwav an additional time without impacting results. THat's why i removed it (and this comment is here explaining why). Do you disagree with the reasoning? because then there must be a sleeping bug somewhere.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
FlorisBuwaldaDeltares: Technically we cannot register wave height here, we have no guarantee wave height has been succesfully supplied at this point. Unless wave_input_direction is one of those coupled quantities where every datafile giving input data for direction always also contains waveheight?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
FlorisBuwaldaDeltares: In the case 

Image

the external forcing supplies only waveperiod, and wave force x+y. 

Setwavfu still needs the converted period to depth-limit the supplied forces

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
FlorisBuwaldaDeltares: I think this is correct, flowwithoutwaves still calls wave_uorbrlabda() which requires wave significant height and period.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
FlorisBuwaldaDeltares: I double checked, the old codepath applied the limiter before "transform_wave_physics_hp" which effectively overwrote it. Doing it after actually enables the limiter, and thus changes testcase. If you think the old behavior was a bug i'll update the references and restore the statement

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
FlorisBuwaldaDeltares: i think this is the same as the issue above? the old code didn't apply the limiter for offline calculations, if i add it here i get different results. But it seems not applying the limiter was a sleeping bug, so i'll put it back and update the references. online swan should properly be limited now?

src/utils_lgpl/deltares_common/packages/deltares_common/src/transform_wave_physics.f90
FlorisBuwaldaDeltares: I think i explained it above, this is the flowwithoutwaves branch which still needs force xy and period. And I think the routine is used?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
FlorisBuwaldaDeltares: m_sediment also exports m_waves in its entirety. That is something to fix, but i'd prefer to do it somewhere else. for now i'll add fwac and fbreak too

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
FlorisBuwaldaDeltares: done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_flowinit.f90
FlorisBuwaldaDeltares: done!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
FlorisBuwaldaDeltares: i've made them locals so the compiler mayh optimize freely

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_offline_wave_external_forcing.f90
FlorisBuwaldaDeltares: i added some teardown

src/utils_lgpl/deltares_common/packages/deltares_common/src/transform_wave_physics.f90
jreyns: not in the transform_physics routine just below it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
jreyns: if wave height is not supplied, the directional interpolation is incorrect and the model should crash

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
jreyns: i think online swan is okay. Don't update the refs yet, I want to run these offline cases manually again with this branch

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute_waves/compute_wave_parameters.f90
jreyns: so my remark below; I think it was indeed missing from the previous code version (or omitted for a reason I cannot remember). I will manually check results.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
FlorisBuwaldaDeltares: yes, after the external forcing initialization is done, during "finalize()" `validate_offline_wave_input_providers` will be called. If the required input for the wave config is not present the simulation will abort with an error. 
That's why I think it's correct to register the wave height only if it's in the ext file as an actual quantity, not as a bonus when input_direction is present.

src/utils_lgpl/deltares_common/packages/deltares_common/src/transform_wave_physics.f90
FlorisBuwaldaDeltares: oh, i think someone didn't apply YAGNI, i'll remove the unused _sp routine

FlorisBuwaldaDeltares: apparently _sp isn't called in an fm solution, but if you build fm_suite i do find a function calling it

jreyns: @FlorisBuwaldaDeltares the _sp version is needed for the single precision delft3d4 builds!

jreyns (approved):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
jreyns: in wave_fillsurdis, do:

      select case (jawave)
      case (WAVE_SWAN_ONLINE .or. WAVE_NC_OFFLINE)
         surdis = dsurf(k) + dwcap(k)
      case (WAVE_SURFBEAT)
         if (roller > 0) then
            surdis = DR(k)
         else
            surdis = D(k)
         end if

jreyns: note that these dissipation fields should then be provided in the coupling, so there should be a check using jawavebreakerturbulence

FlorisBuwaldaDeltares: Done!

jreyns (approved):

rene-deltares (approved):
Floris was the last committer, approving so it is mergable :)

PR #1127, opened 2026-07-24
Read initial sup sediment concentration; add output for transp formula 1,2,4; disable undocumented transp formula 3

hrajagers:
# What was done 

1. Correctly read the initial sediment concentration from
- old external forcings file
- initial conditions file
- newexternal forcings file
into the constituents array instead of the sed array ... the latter isn't used anymore by D-Morphology.

2. Add optional dedicated output to the transport formulas of Engelund-Hansen (formula 1), Meyer-Peter and Mueller (formula 2), and general formula (formula 4) ... useful for debugging, by default switched off

3. Disable the Swanby / Ackers-White formula 3 (not documented in user manual, and potential inconsistency between passed sediment diameter and expected sediment diameter). New issue UNST-10195 to investigate and document.

Switched to all branch replacing #839 since new commits where added that extend the scope beyond fm.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
Three new test cases \
`e02_dflowfm/f022_morphology_2D/c110_sand_settling_OldExtForce` \
`e02_dflowfm/f022_morphology_2D/c111_sand_settling_IniField` \
`e02_dflowfm/f022_morphology_2D/c112_sand_settling_NewExtForce`
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9899


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
```diff
@@ -881,11 +881,11 @@ function resolve_initial_3d_target(quantity, target_location_type, target_array_
...
             return
          end if
-         first_index = iconst - ISED1 + 1
-         target_array_3d => sed
+         first_index = iconst
+         target_array_3d => constituents
 
       case ('initialsediment')
-         if (jased <= 0) then
+         if (jased <= 0 .or. stm_included) then
```
jreyns: i think stm_included implies jased==4

src/utils_gpl/morphology/packages/morphology_kernel/src/eqtran.f90
```diff
@@ -227,6 +227,8 @@ subroutine eqtran(sig, thick, num_layers_grid, ws, ltur, &
    logical :: equi_conc ! equilibrium concentration given (instead of susp. transport rate)
    logical :: sbc_total ! total bed load given (instead of m,n components)
    logical :: sus_total ! total suspended load given (instead of m,n components)
+   !
+   real(fp), parameter :: CHEZY_MIN = 1.0e-6_fp ! minimum Chezy value
```
jreyns, line 231: is this even physical?

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb1.f90
```diff
@@ -72,18 +71,24 @@ subroutine tranb1(utot, d50, c, h, npar, &
       ag = par(1)
       delta = par(4)
       acal = par(11)
-      !rk      = par(12) ! obsolete
+      !rk = par(12) ! obsolete
```
jreyns: delete

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb3.f90
```diff
@@ -83,29 +82,30 @@ subroutine tranb3(utot, d35, c, h, npar, &
       ag = par(1)
       delta = par(4)
       acal = par(11)
-      rk = par(12)
+      !rk = par(12) ! obsolete
```
jreyns: delete

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb3.f90
```diff
@@ -3,7 +3,7 @@ module m_tranb3
 
 contains
 
-   subroutine tranb3(utot, d35, c, h, npar, &
+   subroutine tranb3(utot, d35, chezy, water_depth, npar, &
```
jreyns, line 6: if this formula is not used, why not remove it from the mor module?

jreyns (changes requested):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_inifields.f90
hrajagers: Yes. This check aims to do an early return if you're looking at a case in which `initialsediment` is specified, but the model configuration doesn't actually support sediment. The `initialsedfrac` case above corresponds to `jased == 4`. This case corresponds to `jased == 1`, `2` and `3`. For all other cases, it should fail and return early. I realized that the `jased <= 0` check wouldn't exclude the `jased == 4` case. I'll change the check to focus on jased only.

src/utils_gpl/morphology/packages/morphology_kernel/src/eqtran.f90
hrajagers: Nope. It wasn't intended to be physical. It's just for avoiding divisions by zero. I moved this limiter from the individual transport formulas up to eqtran. A more physical limit would be a Chezy value of 19.425 since that corresponds to a k_N value equal to the water depth h. k_N values larger than that don't strictly make sense .... but I know that they sometimes are used. One should even wonder whether a k_N value close to the water depth scale is physically correct to be used in this way ... For consistency, I keep it as-is. I have created UNST-10212 to evaluate a larger "more physical" limit.

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb1.f90
hrajagers: I had kept the line to explain why par(12) is skipped. I'll remove the line here, but I'll add a more explicit comment in `rdtrafrm`.

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb3.f90
hrajagers: Ackers-White is a commonly used formula in other software. I want to check it and re-enable: UNST-7974.

src/utils_gpl/morphology/packages/morphology_kernel/src/tranb3.f90
hrajagers: removed. Comment added in rdtrafrm.

hrajagers (commented):
3 comments processed by adjusting the code. 2 comments referred to other issues.

jreyns (approved):

PR #1128, opened 2026-07-24
All/bugfix/unst 10036 oneapi 2026 code issues

harmenwierenga:
# What was done

- Fixed multiple bugs in flowgeom_mask that did not initialise num_elements and then properly allocate the mask for multiple location types. This caused unit test_init_spatial_fields_integration.test_initialverticalsalinityprofile to fail with OneAPI 2026
- Throw proper error if an unknown location type is encountered
- Fixed bug in test_init_spatial_fields_integration.test_initialwaterlevel_static_field_populated_at_init where ndx is used before it is set
- Fixed bugs in deltares_common_c code, stronger type checking caught those

# Issue link UNST-10036


rene-deltares (approved):
Nice. After this the Testbench 🤞

thomas-pijls (approved):
Good fix, also very nice that you added a case default to `construct_mask`

PR #1132, opened 2026-07-27
Tc/bugfix/unst 9695 cmake visual studio integration

harmenwierenga:
# What was done 

- For some reason the cmake and visual studio installation did not create a proper integration, and the compilers could not be find. Use a proper release version of cmake now, and also ensure that we use a proper visual studio installer that matches with the latest visual studio version. The installer already installs the latest version of visual studio, so the main difference is that the installer itself, which also provides vswhere and other info about the installation, is also up-to-date.
- Remove set-env-vs2022.cmd after set-env.cmd has been part of main for more than a week. Older branches will have to merge main.

# Issue link UNST-9695


willem-deltares (approved):
I did sometimes see that the stable in https://aka.ms/vs/stable/vs_community.exe is really a pointer at microsoft  to the last stable. Seen from our point of view it is better to use 18 instead of stable (or capture this in a variable).

Using of cmake-x,y,z is indeed better than using cmake-x.y.z-rcw

harmenwierenga:
> I did sometimes see that the stable in https://aka.ms/vs/stable/vs_community.exe is really a pointer at microsoft to the last stable. Seen from our point of view it is better to use 18 instead of stable (or capture this in a variable).
> 
> Using of cmake-x,y,z is indeed better than using cmake-x.y.z-rcw

Yes, I would prefer that too. https://aka.ms/vs/17/release/vs_community.exe does exist, but https://aka.ms/vs/18/release/vs_community.exe does not. It could be a conscious policy from Microsoft, their page says that Visual Studio Enterprise and Professional offer installers for fixed versions, but Community does not: https://learn.microsoft.com/en-us/visualstudio/releases/2026/release-history#release-dates-and-build-numbers

PR #1135, opened 2026-07-27
UNST-10062 Add source/sink cleanup to resetfullflowmodel and use it i…

harmenwierenga:
…n precice test adapter unit tests.

Also use functions to find the expected cell number instead of hard-coded ones

- unstruc_model.f90:resetModel() is made responsible for resetting extfile_new_list
- Added a function for cleaning up source/sink administration, call it from resetfullflowmodel
- flow_geominit should check each variable separately instead of deallocating in bulk, that way it becomes useful for unit tests as well, which may not initialise all variables consistently
- Some small fixes to properly clean up the tests
- Group of tests was rerun with --shuffle

# Issue link UNST-10062


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_geominit.f90
```diff
@@ -529,7 +529,55 @@ subroutine flow_geominit(iphase) ! initialise flow geometry
...
+      end if
+      if (allocated(wu_mor)) then
+         deallocate (wu_mor)
+      end if
+      if (allocated(wu1D2D)) then
+         deallocate (wu1D2D)
+      end if
+      if (allocated(hh1D2D)) then
+         deallocate (hh1D2D)
       end if
```
FlorisBuwaldaDeltares: are these all pure deallocates? if they are allocated later in the routine i find it nicer to use a realloc call instead of: 
if allocated then
deallocate
end if
allocate
call aerr

that's 6 lines per variable

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/source_sink.f90
```diff
@@ -108,6 +109,23 @@ module m_source_sink
...
+         self = SourceSinks()
+      end select
+
+      if (allocated(source_sink_all_discharges)) then
+         deallocate (source_sink_all_discharges)
+      end if
+      if (allocated(source_sink_reduction)) then
+         deallocate (source_sink_reduction)
+      end if
+   end subroutine clear_source_sinks
```
FlorisBuwaldaDeltares: nice! I think however that for other derived types around the codebase we usually use the 'dealloc' interface. However since this is a type-bound procedure maybe that's a good enough reason to deviate from this. Still, we don't use 'clear' much, convention is either "deallocate/dealloc" or "reset"

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/resetfullflowmodel.f90
```diff
@@ -144,7 +146,9 @@ subroutine resetFullFlowModel()
 
       call default_modelbounds()
 
+      call doclose(mext)
```
FlorisBuwaldaDeltares: oh no.. ideally we wouldn't have to call this because any routine that reads mext closes it after it is done :( but I guess this is more robust, it is just ugly

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_precice_adapter.f90
```diff
@@ -48,23 +48,33 @@ end subroutine cleanup_netcells
...
       call grid_helper%make_square_grid( &
          bottom_left_x=0.0_dp, bottom_left_y=0.0_dp, side_length=10.0_dp, &
          rows=1, columns=2, array_size_margin=2 &
          )
       call flow_geominit(0)
 
+      call init_cell_geom_as_polylines()
+      expected_sink_cell = point_find_netcell(5.0_dp, 5.0_dp)
+      expected_source_cell = point_find_netcell(15.0_dp, 7.0_dp)
+      call cleanup_cell_geom_polylines()
```
FlorisBuwaldaDeltares, line 76: glad to see my routine being used :)

FlorisBuwaldaDeltares (approved):
thanks for looking into this! i've got a few comments, nothing major

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_geominit.f90
harmenwierenga: Done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/source_sink.f90
harmenwierenga: I have renamed it to dealloc, to not clash with the built-in deallocate

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/resetfullflowmodel.f90
harmenwierenga: You're right, I have removed it here for now, possible errors should be fixed in the external forcings code

FlorisBuwaldaDeltares (approved):

PR #1136, opened 2026-07-28
UNST-9695 Use proper visual studio 2026 link instead of latest stable…

harmenwierenga:
…, remove redundant version check

# Issue link UNST-9695


willem-deltares (approved):
This will fix it!

PR #1137, opened 2026-07-28
Bugfix for temperatures below zero degrees

goede2606:
# What was done 

- Correct several bugs in `heatun.f90` and `m_fm_icecover.f90` ... some related to the generalization of `0 degC` for fresh water to `t_freeze` for variable salinity. Updated three testcases in this context.
- Updated `partition.f90` to avoid touching the `longculverts` variable when unallocated.
- Added output option for heat fluxes `qh_air2ice` and `qh_ice2wat` to map- and his-files.
- Extended soil temperature model to depth-averaged models; added soil temperature to map-file in case it's used and thermodynamic fluxes are written to the output file.

# Tests 
- [x] Tests updated: e02_f006_c176, e02_f020_c034, e02_f020_c035.

# Documentation  
TODO: check equations in the user manual

# Issue link
UNST-10166


hrajagers:
Three ice test cases updated (`e02_f006_c176` is a 2D model with prescribed salinity, `e02_f020_c034` is the same model with simulated salinity, `e02_f020_c035` is a 3D version thereof). The results of all three simulations are very similar with respect to the simulated ice cover. The new simulations have a bit less ice (see figure: blue = new, red = previous) because the heat flux from ice to water (`qh_ice2wat`) is changed from `rhow * SPECIFIC_HEAT_WATER * c_tz * (-0.01)` to `rhow * SPECIFIC_HEAT_WATER * c_tz * min(-0.01, t_freeze - tempwat)`.

image

goede2606:
Review by @goede2606 of commit c343a83:

-  in m_fm_icecover.f90 subroutine preprocess_icecover has been improved in the way as agreed between @hrajagers en @goede2606. The calculation of qh_ice2at(n) has improved.
- in heatun.f90 the calculation of  solar_radiation_soil_heat_flux and  heatsrc0(k_bot) has been improved.

So, I fully agree with all code changes.

goede2606:
Review by @goede2606 of commit 323a616

In the NetCDF output of the MAP output files the quantities qh_air2ice, qh_ice2wat and tbed for the air-to-ice heatflux, the ice-2-water heatflux and for the bed temperature has been added correctly.

So, I fully agree with all code changes.

goede2606:
Review by @goede2606 of commit 363b4a2:

In the NetCDF output of the HIS output files the quantities qh_air2ice, qh_ice2wat and tbed for the air-to-ice heatflux, the ice-2-water heatflux and for the bed temperature has been added correctly.

So, I fully agree with all code changes in the corresponding five subroutines.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
```diff
@@ -314,7 +314,7 @@ subroutine preprocess_icecover(n, Qlong_ice, tempwat, saltcon, wind)
 
       ! Compute freezing point
       t_freeze = freezing_temperature(saltcon)
-      t_ref = celsius_to_kelvin(0.0_fp) ! 0.0 or t_freeze ?
+      t_ref = celsius_to_kelvin(t_freeze)
```
FlorisBuwaldaDeltares, line 317: is t_freeze still needed after celsius to kelvin conversion? t_freeze makes the code more readable than t_ref, if the answer is no, you could consider t_freeze = celsius_to_kelvin(freezing_temprature(saltcon)),

but it's a trivial change, not worth starting a whole pipeline over

goede2606:
Hoi @rene-deltares, hoi Rene, 

Van de onderstaande vier commits van Bert heb ik de laatste drie gereviewed. Dit betreft commits ten behoeve van ijsmodellering. (Nb. De eerste commit betreft long culverts en die zal door Floris gereviewed worden.)

In het issue ([Comparing main...fm/bugfix/UNST-10166_fix_for_temperatures_below_zero_degrees · Deltares/Delft3D](https://github.com/Deltares/Delft3D/compare/fm/bugfix/UNST-10166_fix_for_temperatures_below_zero_degrees)) heb ik voor elke commit mijn review commentaar toegevoegd. Voor alle drie is dit precies conform verwachting , zoals Bert en ik dat ook onderling besproken hebben.
Zou jij dit nu als reviewer willen goedkeuren, omdat ik hiertoe geen permissie heb? 

Alvast bedankt hiervoor en met vriendelijke groet,

Erik de Goede

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
```diff
@@ -6866,6 +6886,9 @@ subroutine unc_write_map_filepointer(imapfile, tim, jaseparate) ! wrimap
                   end if
 
                   call definencvar(imapfile, id_Qtot(iid), nf90_double, idims, 'Qtot', 'total heat flux', 'W m-2', 'FlowElem_xcc FlowElem_ycc')
+                  if (soiltempthick > 0.0_dp) then
+                     call definencvar(imapfile, id_tbed(iid), nf90_double, idims, 'tbed', 'Temperature of the bed', 'degC', 'FlowElem_xcc FlowElem_ycc')
+                  end if
```
FlorisBuwaldaDeltares, line 6891: why was this part of this PR? doesn't have anything to do with ice as far as I can see?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -354,32 +355,36 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
...
+                  salinity = constituents(isalt, n)
+               else
+                  salinity = constituents(isalt, k_top)
+               end if
+            else
+               salinity = backgroundsalinity
+            end if
+            t_freeze = freezing_temperature(salinity)
 
+            if (ice_thickness(n) > MIN_ICE_SNOW_THICKNESS .or. (water_temperature_in_cell < t_freeze + 0.1_fp .and. air_temperature(n) < 0.0_fp)) then
```
FlorisBuwaldaDeltares: why a margin of .01? also I think according to our coding guidelines should be a named constant

FlorisBuwaldaDeltares (approved):
few small questions, otherwise all looks good

rene-deltares (approved):
I approved on Erik's behalf

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
hrajagers, line 240: We don't apply `dexp` in 2D, so we are adding all solar radiation to the water column ... and maybe some to the bed. That's double counting (if bed temperature is included) and should ideally be changed. If we apply `dexp` always, this may influence many 2D thermal models (leading typically to lower temperatures since we never included the bed temperature in such models until now). Maybe we need a switch depending on the inclusion of bed temperature ... would we then also like to do that in 3D? Created UNST-10334 to investigate this further after the 2027.01 code freeze.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
hrajagers: The ice thickness is very sensitive to this value. We have planned a session to investigate this further. The current 0.01 value is consistent with the previous code. It could be related a background heat of the bed in shallow systems ... or a 0.2 ppt error in the salinity (locally the salinity might be higher because it is expelled from the ice when freezing). I will change all floats into PARAMETERS in the next ice update.

hrajagers (approved):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
hrajagers: We use `t_freeze` variable both in `heatun` and here. Both variables are currently defined in celsius ... consistency is also worth something. Maybe we should introduce `t_freeze_kelvin` next to `t_freeze_celsius` ... or rename all temperature variables consistent with their unit, or choose one unit and stick to it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_io/unstruc_netcdf.f90
hrajagers: I tried to use the bed temperature instead of the magic `-0.01` value in the computation of `qh_ice2wat` and then wanted to check the `tbed` value, but for the upcoming release it's better to stick to the `-0.01` ... and the option to write out `tbed` is an improvement anyway.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
hrajagers: You wrote `.01` ... I thought that you were referring to the `-0.01` in the `qh_ice2wat` computation a couple of lines lower. I now realize that you may have referred to the `0.1` in the selected line. We used to have `water_temperature_in_cell < 0.1_fp` and the current code is consistent with that for fresh water. I think that this margin is not actually needed.

FlorisBuwaldaDeltares: oh oops. Well we can leave the margin investigation for the next issue

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/m_fm_icecover.f90
FlorisBuwaldaDeltares: well i see many other variables here that could use some renaming, so fine to also leave that for later

FlorisBuwaldaDeltares (approved):

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_statistical_output.f90
```diff
@@ -1674,6 +1674,14 @@ subroutine default_fm_statistical_output()
                              'Wrihis_snow_temperature', 'snow_temperature', 'snow temperature', 'temperature_in_surface_snow', &
                              'K', UNC_LOC_STATION, nc_attributes=atts(1:1), description='Write snow temperature to his-file', &
                              nc_dim_ids=station_nc_dims_2D)
+      call add_output_config(config_set_his, IDX_HIS_QH_AIR2ICE, &
+                             'Wrihis_heatflux_air_to_ice', 'qh_air2ice', 'Heat flux from air to ice', '', &
+                             'W m-2', UNC_LOC_STATION, nc_attributes=atts(1:1), description='Write heat flux from air to ice to his-file', &
+                             nc_dim_ids=station_nc_dims_2D)
+      call add_output_config(config_set_his, IDX_HIS_QH_ICE2WAT, &
+                             'Wrihis_qh_ice2wat', 'qh_ice2wat', 'Heat flux from ice to water', '', &
```
hrajagers: Similar change needed as for previous quantity to be consistent with `icecover_input_module`:
```suggestion
                             'Wrihis_heatflux_ice_to_water', 'qh_ice2wat', 'Heat flux from ice to water', '', &
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/heatun.f90
```diff
@@ -354,32 +365,38 @@ subroutine heatun(n, time_in_hours, nominal_solar_radiation)
...
+                  salinity = constituents(isalt, k_top)
+               end if
+            else
+               salinity = backgroundsalinity
+            end if
+            t_freeze = freezing_temperature(salinity)
 
+            qh_air2ice(n) = 0.0_fp
+            qh_ice2wat(n) = 0.0_fp
+            if (ice_thickness(n) > MIN_ICE_SNOW_THICKNESS .or. (water_temperature_in_cell < t_freeze + 0.1_fp .and. air_temperature(n) < 0.0_fp)) then
```
hrajagers: Setting `qh_air2ice` and `qh_ice2wat` to 0.0_fp whenever we don't go into this if-clause, causes a thin layer of ice to remain. The previous code that kept the old values would use the old heat fluxes which would melt the ice. Maybe we can get rid of these thresholds and margins.
```suggestion
            if (ice_thickness(n) > 0.0_fp .or. (water_temperature_in_cell < t_freeze .and. air_temperature(n) < 0.0_fp)) then
```

hrajagers: Setting `qh_air2ice` and `qh_ice2wat` to zero instead keeping the old value has had an unexpected side effect: keeping the old value caused the remaining ice to be melted (extrapolating the previous time steps). I agree that setting those fluxes to zero when there is no ice, is nicer but then we probably need to drop the threshold value for `ice_thickness`.
```suggestion
            qh_air2ice(n) = 0.0_fp
            qh_ice2wat(n) = 0.0_fp
            if (ice_thickness(n) > 0.0_fp .or. (water_temperature_in_cell < t_freeze .and. air_temperature(n) < 0.0_fp)) then
```

hrajagers (changes requested):
* Inconsistent `wrihis` flag.
* Setting ice thickness to 0 too early prevents last layer of ice not to melt.

FlorisBuwaldaDeltares (approved):

hrajagers:
I've increased the tolerances for two ice test cases: the ice temperature is now about 0.1 degree different at one moment during the simulation (see figure, red = reference, blue = new). Looking at the variability of the ice temperature that is surprisingly little, but we allowed at most 0.001 degree difference. Increased the allowed deviation to 0.2 degree and didn't update the references. Also reduced the sensitivity of the checks for thickness to 1 cm and water temperature to 0.01 degree (larger water body, so should change less than the ice temperature).

image

hrajagers (approved):

rene-deltares (approved):
The tolerance increase seems reasonable to me!

PR #1139, opened 2026-07-28
UNST-8900: time independent windspeed factor

rene-deltares:
# What was done 

Add support for `[Spatial]` block in new ext file with a `dataValue`. This lets you multiply a quantity with a constant value (with an option to use a polygon so this only happens in a certain region.

UPDATE 2026-08-14: I added a test case to the test bench that uses a data value multiplier for both solar radiation and wind speed. It is e02_f006_c080. The test case is a copy from a small heat flux model (e02_f020_c040) with minor modifications to use the data value multiplier. Also a polygonal mask is applied to the multipliers. The 'before-after' comparison of the map file looks like this:

windspeed_solar_diff

You can see that the solar radiation value ratios of the map files (after / before) is exactly the data value factor bounded by the polygon. For wind speed you can see that the adjacent cells to the polygonal region are also modified a bit. This is apparently because the windxy quantities are calculated on the cell edges, and the cell centers are interpolated

# Evidence of the work done 

- [x]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated
  - e02_f006_c080 (modified version of e02_f020_c040)
  - 3 unit tests in 'test_ec_module_data_value.f90'
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -80,6 +80,7 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
       logical, optional, intent(in) :: quiet !< When .true., in case of errors, do not write the errors to screen/dia at the end of the routine.
       character(len=*), optional, intent(in) :: varname !< variable name within filename
       character(len=*), optional, intent(in) :: varname2 !< variable name within filename
+      real(hp), optional, intent(in) :: data_value !< Data value used for multiplying quantities with a constant factor.
```
arthurvd:
```suggestion
      real(hp), optional, intent(in) :: data_value !< Scalar constant data value used instead of filename/forcingfile.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -255,7 +256,10 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
                   if (present(dtnodal)) then
                      success = ecSetFileReaderProperties(ecInstancePtr, fileReaderId, ec_filetype, filename, refdate_mjd, tzone, ec_second, name, dtnodal=dtnodal / 86400.0_dp, varname=varname)
                   else
-                     if (present(varname2)) then
+                     if (present(data_value)) then
```
arthurvd: I think Fortran allows you to pass on optional arguments without present() checks, if they are both optional in the caller and callee. So the three call be low can be done at once:

```
                     success = ecSetFileReaderProperties(ecInstancePtr, fileReaderId, ec_filetype, filename, refdate_mjd, tzone, ec_second, name, data_value=data_value, varname=varname, varname2=varname2)
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1438,6 +1448,8 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
       case ('solarradiation')
          if (ec_filetype == provFile_netcdf) then
             sourceItemName = 'surface_downwelling_shortwave_flux_in_air'
+         else if (ec_filetype == provFile_datavalue) then
+            sourceItemName = 'solarradiation'
```
arthurvd: Which piece of code required this change? I'd hope that the `else` case (or even any random name) would suffice.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -1091,6 +1097,9 @@ module function init_spatial_fields(block_ptr, base_dir, file_name, group_name)
                                                 method, oper, varname=variable_name, tgt_item1=ec_item, tgt_data1=target_data)
                else if (target_location_type == UNC_LOC_S3D) then
                   res = read_3d_sigma_field(quantity, target_x, target_y, mask, kx, forcing_file, filetype, method, oper, variable_name, ec_item, target_data)
+               else if (comparereal(input%data_value, 0.0_dp) /= 0) then
+                  res = ec_addtimespacerelation(quantity, target_x, target_y, mask, kx, forcing_file, filetype, &
+                                                method, oper, data_value=input%data_value, tgt_item1=ec_item, tgt_data1=target_data)
```
arthurvd: 0.0 should be an allowed value too. Check instead on a local logical that says whether the data_value was provided in user input or not.
Or, alternatively, set data_value to dmiss and check on that (that only means we do not support the user to specify -999.0, which is more acceptable.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
```diff
@@ -65,6 +65,7 @@ module m_spatial_field
       integer :: oper = OPERAND_OVERRIDE !< Operand enum, derived from operand_string, defaulting to OPERAND_OVERRIDE.
       integer :: method = -1 !< FM interpolation method enum, derived by validate_spatial_field_input. -1 = not yet derived.
       integer :: filetype = -1 !< FM file type enum, derived by validate_spatial_field_input. -1 = not yet derived.
+      real(dp) :: data_value = 0.0_dp !< Time and space independent value, used for multiplying quantities with a constant factor.
```
arthurvd: See remark above regarding default dmiss (or undefined)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/time_space_parameters.f90
```diff
@@ -169,6 +170,8 @@ function convert_file_type_string_to_integer(string) result(file_type)
          file_type = BCASCII
       case ('curvigrid')
          file_type = CURVI
+      case ('datavalue')
+         file_type = DATAVALUE
```
arthurvd: We do not want to require from the user that `dataFileType` is specified. It is OK to have an enum value that says file_type = DATAVALUE, but it should not come from input.
When `dataFile`/`dataFileType` is not provided, but `dataValue` is, then the filetype should be automatically set.

In short: this case can safely be removed here.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
```diff
@@ -223,13 +226,32 @@ function validate_spatial_field_input(input, file_name, group_name, base_dir) re
          return
       end if
 
-      if (len_trim(input%forcing_file) == 0) then
-         write (msgbuf, '(5a)') 'Incomplete block in file ''', trimmed_file_name, ''': [', trimmed_group_name, ']. Field ''forcingFile'' is missing.'
+      input%filetype = convert_file_type_string_to_integer(input%forcing_file_type)
+      if (input%filetype == FILE_TYPE_UNKNOWN) then
```
arthurvd: Only do these conversion + checks when `dataFileType` is given. When user input only contains `dataValue` auto-set the filetype.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/cutcell_list.f90
```diff
@@ -85,7 +85,10 @@ subroutine cutcell_list(n12, jamasks)
 
       jastored = 0
 
-      inquire (FILE=md_cutcelllist, EXIST=JAWEL)
+      JAWEL = .false.
+      if (len_trim(md_cutcelllist) > 0) then
```
rene-deltares, line 89: Before I made this change, over a hundred test cases failed with Release binaries. `md_cutcelllist` is a 1024 length string of blanks. `JAWEL` is `0x11111111` after the `inquire` (reproduced it with RelWithDebugInfo binaries). I have no idea what is going on here, but the `if` statement fixed it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1437,8 +1446,10 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
       case ('solarradiation')
          if (ec_filetype == provFile_netcdf) then
             sourceItemName = 'surface_downwelling_shortwave_flux_in_air'
-         else
+         else if (ec_filetype == provFile_arcinfo .or. ec_filetype == provFile_curvi) then
             sourceItemName = 'sw_radiation_flux'
+         else
+            sourceItemName = 'solarradiation'
```
FlorisBuwaldaDeltares: the nice thing about https://issuetracker.deltares.nl/browse/UNST-10067 is that I think we will be able to remove these

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1028,6 +1031,12 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
             if (.not. success) then
                goto 1234
             end if
+         else if (ec_filetype == provFile_datavalue) then
+            sourceItemId = ecFindItemInFileReader(ecInstancePtr, fileReaderId, 'windxy')
+            success = (sourceItemId /= ec_undef_int)
+            if (.not. success) then
+               goto 1234
+            end if
```
FlorisBuwaldaDeltares, line 970: idem

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -1133,6 +1136,9 @@ module function init_spatial_fields(block_ptr, base_dir, file_name, group_name)
...
```
FlorisBuwaldaDeltares: is it an option to always provide data value, and inside addtimespacerelation to ignore it if it is missing? then we don't need an extra call site here.. i'm already unhappy about the fact we have a separate one for variable_name and 'bcascii'..

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_ec_module_data_value.f90
```diff
@@ -0,0 +1,257 @@
...
+      call F90_EXPECT_TRUE(ok, cstr("ec_addtimespacerelation failed for windx"))
+
+      quantity_name = "windy"
+      ok = ec_addtimespacerelation(quantity_name, x, y, mask, 1, "global", &
+         BCASCII, SPACEANDTIME, OPERAND_OVERRIDE, forcingfile=BC_FILE)
+      call F90_EXPECT_TRUE(ok, cstr("ec_addtimespacerelation failed for windy"))
+
+      quantity_name = "windxy"
+      ok = ec_addtimespacerelation(quantity_name, x, y, mask, 1, "", &
+         DATAVALUE, JUSTUPDATE, OPERAND_MULTIPLY, data_value=WINDSPEEDFACTOR)
```
FlorisBuwaldaDeltares, line 95: oh, the other tests call init_spatial_field, but maybe doing a direct addtimespacerelation is not so bad

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -4133,6 +4135,59 @@ subroutine ConvertToSparseIndices(n_points, indices, n_rows, ia, ja)
 
    end subroutine ConvertToSparseIndices
 
+   !> Converter for the 'datavalue' provider.
+   !! The source item holds a single, time- and space-independent scalar value.
+   !! This converter broadcasts that scalar to every element of every target item,
+   !! applying the configured operand (typically multiply) and honouring the
+   !! optional target mask. It handles the common case where the target has
+   !! multiple components (e.g. windx + windy for the 'windxy' quantity), which
+   !! ecConverterUniform does not support for a scalar source.
```
FlorisBuwaldaDeltares: did you write this yourself or did copilot help? I find that without explicit instruction he's always waaay to verbose for apidocs. Generally we keep function descriptions to 1-2 sentences without going into all the details. But if you prefer this description that's ok, just checking

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -4133,6 +4135,59 @@ subroutine ConvertToSparseIndices(n_points, indices, n_rows, ia, ja)
...
+   function ecConverterDataValue(connection, timesteps) result(success)
```
FlorisBuwaldaDeltares, line 4233: I lack the knowledge of EC module to properly review this.. if it works it works but hopefully @arthurvd can chime in here

src/utils_lgpl/ec_module/packages/ec_module/src/ec_module.f90
```diff
@@ -378,6 +378,7 @@ module m_ec_module
 
    interface ecCreateInitializeBCFileReader
       module procedure ecProviderCreateInitializeBCFileReader
+      ! module procedure ecUnitProviderCreateInitializeBCFileReader
```
FlorisBuwaldaDeltares: why is this commented line here?

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -125,6 +125,22 @@ function ecProviderCreateInitializeBCFileReader(instancePtr, forcingfile, locati
...
+   !    integer, intent(in) :: forcingfileUnit
+   !    character(len=*), intent(in) :: location
+   !    character(len=*), intent(in) :: quantity
+   !    real(dp), intent(in) :: k_refdat !< kernel ref date
+   !    real(dp), intent(in) :: k_tzone !< kernel time zone
+   !    integer, intent(in) :: k_tsunit !< kernel timestep unit (1=sec, 2=min, 3=hour)
+   !    integer, intent(out) :: fileReaderId !< unique fileReader id
+   !    character(len=*), optional, intent(in) :: funtype !< matching function in the BC-block header
+   ! end function ecUnitProviderCreateInitializeBCFileReader
+
```
FlorisBuwaldaDeltares: I think this one can go too, right?

FlorisBuwaldaDeltares (approved):
looks good! I do have to say it's harder than I'd have hoped to set up a new ecProvider and a new ecConverter.. hopefully the second reviewer can look at those critically

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
arthurvd: Indeed, over time we need to get rid of most of these special cases.
For now, an alternative could be to precede the `select case (str_tolower(trim(target_name)))` on line 701 with:
```
      if (ec_filetype == provFile_datavalue) then
      ! ... jump to the default handling
```
This is still important within this PR, because we'd like the datavalue to be supported across all quantities, so not leading to errors inside the incomplete select case handling.
As a test: does the dataValue work as well for `quantity = frictionCoefficient`?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_ec_module_data_value.f90
```diff
@@ -0,0 +1,257 @@
...
+
+      ! Assert
+      ok = ec_gettimespacevalue(ecInstancePtr, item_windx, IREFDATE, tzone, tunit, 50.0_dp)
+      call f90_expect_near(wx(1), 1.0_dp, 1.0e-6_dp, cstr("windxy_x@50")) ! 1.0 is halfway between -1.0 and 3.0
+
+      ok = ec_gettimespacevalue(ecInstancePtr, item_windy, IREFDATE, tzone, tunit, 50.0_dp)
+      call f90_expect_near(wy(1), -1.0_dp, 1.0e-6_dp, cstr("windxy_y@50")) ! -1.0 is halfway between -4.0 and 2.0
+
+      ok = ec_gettimespacevalue(ecInstancePtr, item_windxy_x, IREFDATE, tzone, tunit, 50.0_dp)
```
arthurvd: Just checking: what would happen if you call this one again, or even more subtle:

```suggestion
      ok = ec_gettimespacevalue(ecInstancePtr, item_windxy_x, IREFDATE, tzone, tunit, 50.0_dp)
      ok = ec_gettimespacevalue(ecInstancePtr, item_windxy_y, IREFDATE, tzone, tunit, 50.0_dp) ! <-- NEW
```
Are the results still correct? No multiplication with 0.8 twice?

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
arthurvd: Yes, see remarks below.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -4133,6 +4135,59 @@ subroutine ConvertToSparseIndices(n_points, indices, n_rows, ia, ja)
...
+         targetMask => connection%targetItemsPtr(i)%ptr%elementSetPtr%mask
+         if (connection%converterPtr%targetIndex /= ec_undef_int) then
+            jmin = connection%converterPtr%targetIndex
+            jmax = connection%converterPtr%targetIndex
+         else
+            jmin = 1
+            jmax = connection%targetItemsPtr(i)%ptr%elementSetPtr%nCoordinates
+         end if
+
+         call check_undefined_values_for_operand(connection%converterPtr%operandType, targetField%arr1dPtr(jmin:jmax), status)
```
arthurvd: The only thing is that if *any* target value is now undefined, EVEN outside of the targetMask, then this method will still error and return early.
But, Thomas created that subroutine intentionally to handle an entire target array, so that we don't have to call that function *inside* the do-loop below.

Maybe in the case of masking that is ok?

```suggestion
         if (.not. present(targetMask)) then
               call check_undefined_values_for_operand(connection%converterPtr%operandType, targetField%arr1dPtr(jmin:jmax), status)
         else
            status = .true.
            ! Handle undefined check in masked j-loop below
         end if
```

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -4133,6 +4135,59 @@ subroutine ConvertToSparseIndices(n_points, indices, n_rows, ia, ja)
...
+         if (.not. status) then
+            return
+         end if
+
+         do j = jmin, jmax
+            if (associated(targetMask)) then
+               if (targetMask(j) == 0) then
+                  cycle
+               end if
```
arthurvd: To make the previous remark complete, see this fragment:
```suggestion
               else
                  call check_undefined_values_for_operand(connection%converterPtr%operandType, [targetField%arr1dPtr(j)], status)
                  if (.not. status) then
                     return
                  end if
               end if
```

I don't like it too much, open for better suggestions.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -254,6 +270,7 @@ recursive function ecProviderInitializeFileReader(instancePtr, fileReaderId, fil
       real(dp), optional, intent(in) :: dtnodal !< Nodal factors update interval
       character(len=*), optional, intent(in) :: varname !< variable name within filename
       character(len=*), optional, intent(in) :: varname2 !< variable name 2 within filename
+      real(dp), optional, intent(in) :: data_value !< Data value file reader
```
arthurvd:
```suggestion
      real(dp), optional, intent(in) :: data_value !< Constant data value (still creates a tEcFileReader, but without underlying data file)
```

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -957,6 +992,85 @@ end function ecProviderCreateUniformItems
 
    ! =======================================================================
 
+   !> Create a source Item holding a single time- and space-independent constant.
+   !! Used for the 'dataValue' forcingFileType, where a Spatial ext-block specifies
+   !! a scalar 'dataValue' that is combined with a target quantity. The value is stored in both
+   !! sourceT0 and sourceT1 fields so that time interpolation always yields it.
```
arthurvd: Don't use FM-specific terminology, EC-module should be(come more) kernel agnostic.
```suggestion
   !! Used for the 'dataValue' forcingFileType, where no data file is given, but instead
   !! a scalar 'dataValue'. It can typically (but not per se) be combined on top of another
   !! provider+connection, using an operand. The value is stored in both sourceT0 and
   !! sourceT1 fields so that time interpolation always yields it.
```

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -957,6 +992,85 @@ end function ecProviderCreateUniformItems
...
+      field1Id = ecInstanceCreateField(instancePtr)
+      if (.not. ecFieldCreate1dArray(instancePtr, field1Id, 1)) then
+         return
+      end if
+
+      itemId = ecInstanceCreateItem(instancePtr)
+      if (.not. ecItemSetRole(instancePtr, itemId, itemType_source)) then
+         return
+      end if
+      if (.not. ecItemSetType(instancePtr, itemId, accessType_fileReader)) then
```
arthurvd, line 1035: Code is correct, but it is subtle that we are using the `accesstype_fileReader` for something without a data file. Let's explicitly document that it's a deliberate choice.
```suggestion
      ! Note: even though dataValue has no underlying data file, we still have a tEcFileReader acting as provider.
      if (.not. ecItemSetType(instancePtr, itemId, accessType_fileReader)) then
```

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -957,6 +992,85 @@ end function ecProviderCreateUniformItems
...
+         return
+      end if
+      
+      item => ecSupportFindItem(instancePtr, itemId)
+      ! Store the constant value in both time slots so the temporal interpolator returns `data_value` for any t.
+      item%sourceT0FieldPtr%arr1dPtr(1) = data_value
+      item%sourceT0FieldPtr%timesteps = 0.0_dp
+      item%sourceT1FieldPtr%arr1dPtr(1) = data_value
+      item%sourceT1FieldPtr%timesteps = huge(0.0_dp)
+      item%quantityPtr%vectorMax = 1
```
arthurvd, line 1057: is this out of safety? Or truly needed? Which code fragment elsewhere relies on this?

And, related, why is `provfile_dataValue` not in `ecFileReaderReadNextRecord()`.

I actually like that, but where do you prevent a call to `ecFileReaderReadNextRecord()`?

These last remarks are maybe easier answered in a brief call.

arthurvd (changes requested):
Nice work and it seems to work well.
I have several remarks, mainly small stuff.
Maybe a brief call will help getting this merged soon.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
rene-deltares: Seems to work! 👍

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_ec_module_data_value.f90
rene-deltares: Yes, you get `WINDSPEEDFACTOR**2`. Just checking what happens if you a separate datavalue to `windx` and `windy` instead.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
rene-deltares: Ah, I ran into this technicality: `dmiss` from `m_missing` is not a compile time constant. But I can work around this in `read_spatial_input`.

FlorisBuwaldaDeltares: we used to have dmiss_neg for this exact reason but Thomas removed it :(

src/utils_lgpl/ec_module/packages/ec_module/src/ec_module.f90
rene-deltares: Ah, this was left over from an attempt to have `ec_module` to read data from an already opened file (hence the "Unit" in the name).
Creating files in unit tests gives me bad vibes. I did not get it working this time but... one day I will figure it out.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
rene-deltares: trufax

src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
rene-deltares: In the case of a masked array I now gather all of the 'masked-in' array indices in a separate array and effectively moved the if out of the do.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
rene-deltares: `sourceT0FieldPtr%arr1DPtr(1)` is read in `ec_converter`, in the `ecConverterDataValue` function to retrieve the `data_value`.
Both `sourceT0FieldPtr%timesteps` and `sourceT1FieldPtr%timesteps` are read by `ecItemUpdateSourceItem` in `ec_item`. There a comparison is done between `timesteps%mjd()` and `item%sourceT1FieldPtr%timesteps` (which is huge) so the function returns early there.

Indeed, it seems like `sourceT1FieldPtr%arr1dPtr(1)` is never read.

The only calls to `ecFileReaderReadNextRecord` I can find are in `ecItemUpdateSourceItem`. But these calls are never reached in case of a data value source item because of this line: https://github.com/Deltares/Delft3D/blob/d6f3556ceb1846cc9b3b256d4ae9080a9576b181/src/utils_lgpl/ec_module/packages/ec_module/src/ec_item.f90#L538

```
 else if (comparereal(timesteps%mjd(), item%sourceT1FieldPtr%timesteps) == -1) then
     ! requested time is before T1
     success = .true.
     return
 endif
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
rene-deltares: Unfortunately, it turns out that `windxy` still requires special handling, because the "default behavior" did not work in cases where there are two target items. But `solarradiation`, `windx` and `windy` did work!

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -3448,6 +3555,10 @@ function ecProviderInitializeTimeFrame(fileReaderPtr, k_refdate, k_timezone, k_t
          else
             success = .true.
          end if
+      case (provFile_datavalue)
```
rene-deltares, line 3558: Note to self: Is this really needed or a 'copilot-ism'

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -697,7 +695,14 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
       sourceItemId_4 = 0
       source_connection_created = .false.
 
-      select case (str_tolower(trim(target_name)))
+      selector = str_tolower(trim(target_name))
+      if (ec_filetype == provFile_datavalue .and. .not. associated(targetItemPtr2)) then
```
rene-deltares, line 699: Question to reviewers: How repelled are you by my h4x0r skillz?

FlorisBuwaldaDeltares: hmm i have a feeling this might not be necessary but i have to think about it a bit..

arthurvd: @FlorisBuwaldaDeltares : it is truly necessary: the fallback option outside of the select case only handles a *single* target item, whereas some vector quantities have two targetitems.
For example, check the special case for 'windxy':
```
         if (success) then
            success = ecAddConnectionTargetItem(ecInstancePtr, connectionId, item_windxy_x)
         end if
         if (success) then
            success = ecAddConnectionTargetItem(ecInstancePtr, connectionId, item_windxy_y)
         end if
```
Let's discuss this after standup.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
arthurvd: OK, I understand, but if the time-checking *does* occur (so, your `huge` value is needed), then I find it much safer to keep the current code, that is: keep the sourceT1 field assignment as well.

arthurvd (commented):
Hi @rene-deltares, I checked all your changes and resolved most conversations. One is left for discussion with @FlorisBuwaldaDeltares this morning.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
FlorisBuwaldaDeltares: very nice!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
FlorisBuwaldaDeltares: I now see why this is necessary. However this is overly complex: this line is ONLY for windxy support, it does not make other 2-datapointer quantities automagically supported. Therefore I would suggest that we replace `.not. associated(targetItemPtr2))` with `.and. selector /= 'windxy'`. Then at least it is clear that this is a quantity specific if-statement

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
FlorisBuwaldaDeltares, line 1337: huh why can those be removed? did i forget them in the previous PR??

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
```diff
@@ -213,42 +221,46 @@ function validate_spatial_field_input(input, file_name, group_name, base_dir) re
...
          call err_flush()
          return
       end if
 
-      if (len_trim(input%forcing_file_type) == 0) then
-         write (msgbuf, '(5a)') 'Incomplete block in file ''', trimmed_file_name, ''': [', trimmed_group_name, ']. Field ''forcingFileType'' is missing.'
-         call err_flush()
-         return
-      end if
+      if (comparereal(input%data_value, dmiss) /= 0) then
```
FlorisBuwaldaDeltares, line 230: I prefer the equals function but for a simple dmiss check comparereal will do

arthurvd (approved):
We discussed this PR with @rene-deltares, @FlorisBuwaldaDeltares and @arthurvd and angreed it is now ready for merging.

Work left for a future improvement issue:
* go by all remaining special `case`s in `ec_addtimespacerelation()` and try and understand why or whether they are still necessary. As part of the that, the `selector = ''` trick from this PR can be reconsidered then.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
FlorisBuwaldaDeltares: after our discussion for future reference: this line is not only for windxy support, this line is to support all 'specials' that are scalar (like for example 'rainfall') in a generic way, instead of adding an explicit 'datavalue' case there.

Technically I would prefer keeping the divide between special (everything explicit) or simple generic quantity sharper, it is architecturally cleaner and less confusing.

However for now I'm ok with this line. A clarifying comment could be added that this is only for the scalar specials, not for scalar generic quantities

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
FlorisBuwaldaDeltares: oh it's a git diff artifact

FlorisBuwaldaDeltares (approved):
it was quite a job but i'm happy how it turned out :)

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
rene-deltares: It was necessary, I was confused by the `diff` in GitHub.

PR #1141, opened 2026-07-28
UNST-9695 Add waq_definition target to link libraries of part_data_f,…

harmenwierenga:
… because it uses m_waq_precision

# Issue link UNST-9695


harmenwierenga:
Created a tc PR to get the collect step fixed more quickly: https://github.com/Deltares/Delft3D/pull/1142

PR #1149, opened 2026-07-30
UNST-10213  Use of 32 bit integers for SWAN time axis crashes D-Waves

jreyns:
# What was done 

Increased data size for time axis on nc hotfile writer in SWAN to int64 to accomodate runs in the very far future
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
test case to SWAN testbed: P:\dflowfm\maintenance\JIRA\10000-10999\10213
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link

UNST-10213

### DIFF
--- a/src/utils_lgpl/nefis/packages/nefis/CMakeLists.txt
+++ b/src/utils_lgpl/nefis/packages/nefis/CMakeLists.txt
@@ -33,6 +33,14 @@ add_library(${library_name} ${headers}
                             ${source}
                             ${ini_version_file})
 
+# The d3d4-suite also builds nefis_dll with OUTPUT_NAME nefis. On Windows,
+# that DLL's import library would otherwise overwrite this static archive
+# (nefis.lib) in a multi-configuration build directory. Keep the target name
+# unchanged so consumers still link to the static library target.
+if (WIN32)
+    set_target_properties(${library_name} PROPERTIES OUTPUT_NAME nefis_static)
+endif()
+
 # Set additional dependencies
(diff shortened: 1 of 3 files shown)


src/utils_lgpl/nefis/packages/nefis/CMakeLists.txt
```diff
@@ -33,6 +33,14 @@ add_library(${library_name} ${headers}
                             ${source}
                             ${ini_version_file})
 
+# The d3d4-suite also builds nefis_dll with OUTPUT_NAME nefis. On Windows,
```
adrimourits, line 36: I don't expect a nefis-related-change in a CMakeLists.txt file. Is this intended as part of this issue?

adrimourits (commented):
One question about the CMakeLists change.

src/utils_lgpl/nefis/packages/nefis/CMakeLists.txt
jreyns: i ran into this trying to build a linux container, as this model is supposed to run on the cluster.

hrajagers: I've seen nefis dynamic and static library builds overwrite each other's objects ... causing Windows to always rebuild nefis, so you may have a point here. However, I'm a bit surprised that the need for a WIN32 specific statement would be identified while working inside a linux container.

adrimourits (approved):

PR #1150, opened 2026-07-30
UNST-10214 Fix global state in fm unit tests unrelated to precice

harmenwierenga:
# What was done 

Attempt to clean up global state for fm unit tests.

- Added feature to F90TW to register googletest listener which runs at the end of every test invocation, so that we can call resetfullflowmodel at the end of every test.
- Ensure that bubblescreen data is also defaulted when external forcings are defaulted
- Separate the defaulting and deallocation of source/sink data into two functions
- Reallocate xu, yu, and blu separately so that unit tests don't always have to set them all
- Fix the number of polygons that is counted
- Ensure that read_polyline_coordinates returns all data, and does not tell callers to read the global scratch buffers that may become outdated depending on the call order
- Add more fields to cleanup_grid
- Ensure that resetfullflowmodel is called before inidat in test setup
- Initialise Network data with default values in 1d code
- Ensure that the copying from the temporary, previous dia file to the new dia file only happens when they do not point to the same file on disk. In unit tests, this could lead to tests in infinite loops that would keep duplicating lines in the dia file while copying previous lines to the end of the file

I timed the runtime of the unit tests on Windows in Release.
Full ctest run three times:
- main: 26.13 / 26.26 / 25.77 s
- branch: 27.36 / 27.17 / 26.76 s
Directly run test_dflowfm_kernel_gtest.exe three times:
- main: 5.78 / 5.84 / 5.77 s
- branch: 7.21 / 9.07 / 7.48 s

So the total set of tests appears to be around 1 to 2 seconds slower than previously in release. We can discuss whether this is too much
# Issue link UNST-10214


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -1329,6 +1331,15 @@ subroutine add_longculvert_crosssection(network, flowlink, link_id, cs_def_id)
...
+      line2cross_size = flowlink
+      if (associated(network%adm%line2cross)) then
+         line2cross_size = max(line2cross_size, ubound(network%adm%line2cross, 1))
+      end if
+      call realloc_line2cross(network%adm%line2cross, [line2cross_size, 3], stat)
+      if (stat /= 0) then
+         call SetMessage(LEVEL_ERROR, 'Failed to reallocate line2cross for long culvert')
+         return
+      end if
       network%adm%line2cross(flowlink, :) = t_chainage2cross(c1=icrs, c2=icrs, f=1.0_dp, distance=0.0_dp)
```
FlorisBuwaldaDeltares, line 1343: hmm i'd hope that other initialization would handle line2cross reallocation and that the long culvert subroutine can assume it is properly sized

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_geominit.f90
```diff
@@ -571,7 +571,13 @@ subroutine flow_geominit(iphase) ! initialise flow geometry
...
+         deallocate (xu)
+      end if
+      if (allocated(yu)) then
+         deallocate (yu)
+      end if
+      if (allocated(blu)) then
+         deallocate (blu)
       end if
       allocate (xu(lnx), yu(lnx), blu(lnx), stat=ierr)
       call aerr('xu(lnx), yu(lnx) , blu(lnx)', ierr, 3 * lnx)
```
FlorisBuwaldaDeltares, line 583: this can be replaced with 3 reallocs

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,6 +98,7 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
             exit
          end if
 
+         i_poly = i_poly + 1
```
FlorisBuwaldaDeltares, line 101: do you know why this statement needed to be moved?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/rest_f90/m_read_location_info.f90
```diff
@@ -7,13 +7,14 @@ module m_read_location_info
 !> Generic reader for polyline coordinates from either a locationFile (pli/pliz) or inline keys
    !! (numCoordinates + xCoordinates + yCoordinates + optionally zCoordinates).
    !! Can be reused by source/sinks, long culverts, or any other feature that needs polyline input.
-   subroutine read_polyline_coordinates(block_ptr, object_id, file_name, base_dir, group_name, x_coordinates, y_coordinates, z_coordinates, num_columns, is_successful)
+   subroutine read_polyline_coordinates(block_ptr, object_id, file_name, base_dir, group_name, x_coordinates, y_coordinates, z_coordinates, num_columns, is_successful, fourth_coordinates)
```
FlorisBuwaldaDeltares, line 10: i like the changes but.. why were they necessary??

FlorisBuwaldaDeltares (approved):
overall good fixes i think! just a few questions

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
harmenwierenga: The early 'exit' above means that the polygon is not valid, but i_poly was still incremented. That meant that the i_poly used afterward this do while loop was one too large sometimes.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/rest_f90/m_read_location_info.f90
harmenwierenga: There are a couple of global scratch arrays (xpl, ypl, zpl, dzl) that are used to temporarily store the data. This function then promises to fill that data properly. However, the fourth coordinate (used for a z-range, so bottom and top z are the third and fourth coordinate) was still read from the global scratch array after. That was okay in this case, because the function reading dzl was called directly after read_polyline_coordinates, but there was technically no promise that dzl was still filled with the proper values. The unit tests could hit a wrong order of function calls, and it is better to be consistent between all coordinates.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
harmenwierenga: I have removed this change, it was no longer necessary and you're right that this was unfortunate

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_geominit.f90
harmenwierenga: Done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set(test_target test_dflowfm_kernel_gtest)
 f90twtest(${test_target}
     CFILES test_dflowfm_kernel.cpp
     F90FILES
+        src/dflowfm_test_cleanup.f90
         src/test_dflowfm_kernel_example.f90 # This must be the first Fortran file, as it is referenced in test_dflowfm_kernel.cpp!!
```
FlorisBuwaldaDeltares: is this comment outdated?

src/third_party_open/f90tw/f90tw-main/README.md
```diff
@@ -65,4 +65,8 @@ F90TWTEST(test_name
 )
 ```
 
+An executable can define `F90TW_TEST_TEARDOWN()` in the C++ file of the test
+before including `f90tw_gtest.h` to run a cleanup function after every test.
+This uses GoogleTest's test event listener and also runs after a failed test.
```
FlorisBuwaldaDeltares, line 70: so do we already do this for a test executable? it's a bit complex but i like the idea, but I think if we merge this at least one executable should be calling this so it stays working

FlorisBuwaldaDeltares (approved):
looks good, thanks for fixing! two small questions

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/CMakeLists.txt
harmenwierenga: Good catch, it was placed in the wrong location. I have moved and updated the comment

src/third_party_open/f90tw/f90tw-main/README.md
harmenwierenga: Yes, see src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/test_dflowfm_kernel.cpp. It defines the macro as a funtion call, and the function is defined in Fortran.

FlorisBuwaldaDeltares (approved):

PR #1156, opened 2026-08-03
All/task/unst 9694 add pre cice to conan

AabhasSrivastava:
# What was done 

Added precice and libraries that preCICE depends on in conan package

- Added boost with minimal options in conan
- Added libxml2, eigen in conan
- Added preCICE 3.4.1 in conan
- Added a readme on how to add package to conan
- Update CmakeLists to use conan versions/paths
- Removed preCICE, Boost, Eigen, and libxml2 from the third_party_open folder
 - Removed preCICE, VTK, ASTE, Boost, Eigen, and libxml2 from the the third party docker container. ASTE is not used much and would add too much of a burden, we can see if it can be installed using a python package manager in the devcontainer when necessary

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [x]	Documentation updated \
- [ ]	Not applicable 

# Issue link UNST-9694


conan/README.md
adrimourits, line 1: Nice. Clear description.

src/engines_gpl/fbc/packages/FBCTools/src/CMakeLists.txt
```diff
@@ -115,8 +115,7 @@ add_library(FBCTools_static STATIC
             ${PROJECT_SOURCE_DIR}/../../version/fbc_version.h
 )
 set_target_properties (FBCTools_static PROPERTIES FOLDER engines_gpl/fbc)
-target_link_libraries(FBCTools_static PRIVATE ${LINK_LIBRARIES} ${Boost_LIBRARIES})
-target_include_directories(FBCTools_static PRIVATE ${Boost_INCLUDE_DIR})
+target_link_libraries(FBCTools_static PRIVATE ${LINK_LIBRARIES} Boost::filesystem Boost::date_time)
```
adrimourits, line 118: Good that only the Boost-libraries are added that are really used.

adrimourits (approved):
No remarks at all

ci/python/ci_tools/dimrset_delivery/all-testbench-binaries.json
```diff
@@ -396,10 +428,6 @@
             "file": "bin\\libpetsc.dll",
             "issuedTo": "Stichting Deltares"
         },
-        {
-            "file": "bin\\libxml2.dll",
```
harmenwierenga, line 400: We now link libxml statically into precice I see, that is nice.

ci/python/ci_tools/dimrset_delivery/all-testbench-binaries.json
```diff
@@ -5,27 +5,59 @@
...
             "issuedTo": "Stichting Deltares"
         },
         {
-            "file": "bin\\boost_program_options-vc143-mt-x64-1_90.dll",
+            "file": "bin\\boost_filesystem.dll",
             "issuedTo": "Stichting Deltares"
         },
         {
-            "file": "bin\\boost_thread-vc143-mt-x64-1_90.dll",
+            "file": "bin\\boost_log_setup.dll",
```
harmenwierenga, line 28: At some point I started using the CMake that Boost provides for its find package. This cmake is not fully correct though, it declares certain internal header-only dependencies between boost packages as INTERFACE_LINK_LIBRARIES. However, CMake will then also install the corresponding dlls, even though we only use the headers. I created a hack to remove those superfluous dlls from the relevant INTERFACE_LINK_LIBRARIES. In src/third_party_open/precice/CMakeLists.txt (which you removed here, that's good) I called:

```
    strip_boost_header_only_deps(
        TARGETS      Boost::thread Boost::log Boost::log_setup Boost::filesystem Boost::program_options
        EXCLUDE_DEPS Boost::atomic Boost::chrono Boost::container Boost::regex Boost::serialization Boost::exception
    )
```
This strip_boost_header_only_deps is defined in src/cmake/functions.cmake. We could still call this function to see if we can reduce the installed dlls back to the original set. I do like the clearer names

ci/python/ci_tools/dimrset_delivery/all-testbench-binaries.json
harmenwierenga, line 1: I am missing precice.dll in the installed list. We probably need to add some info to the package info of the precice recipe so that it is properly picked up by our install mechanisms on both Windows and Linux

harmenwierenga: Oh wait, I get it. precice.dll is already part of the list, it did not change. No issue here!

ci/python/ci_tools/dimrset_delivery/all-testbench-binaries.json
harmenwierenga: I have looked into this more. Apparently, the Linux so files really DO depend on these other libraries. It is only on Windows that the dependencies are not necessary. I think that it goes a bit far to add this hack just to clean up a few dlls. I will remove the (now uncalled) function from functions.cmake.

Something that does need fixing is the precice recipe: it currently just says that it requires boost, which means that it will automatically tell cmake that anything that uses precice requires ALL of boost's dynamic libraries. We have made boost a lot smaller in the recipe, but this is still painful. We can override this by declaring which boost components we actually need in the cppinfo of the precice recipe. I have made this change on this branch.

erwin-mulder (approved):
LGTM

PR #1164, opened 2026-08-05
UNST-10233: inform user about new defaults

JulienGroenenboom:
# What was done 
- Add interface and subroutines to inform users about recently changed defaults (when relevant)
 

# Evidence of the work done 
- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10233


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -2438,6 +2457,14 @@ subroutine readMDUFile(filename, istat)
          end if
       end if
 
+      call notify_recent_default_changes( &
```
KoenBussemakerDeltares: I think we can keep the code a bit cleaner if we warn right after parsing the values. That way you don't need to create additional variables to store whether a property was parsed or not. Instead you can just directly call `notify_default_change`. If you pass the `success` value to it then you don't even need to use an if statement, the function will handle that for you. It makes the whole thing a bit more elegant and less easy to forget setting/checking something.

You do give up the everything-is-in-one-place formatting in the .dia file, but IMO that is just a nice to have and not really worth the additional code bloat.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -4221,6 +4251,152 @@ subroutine set_output_time_vector(md_tvfil, ti_tv, ti_tv_rel)
...
+      call notify_default_change_impl( &
+         chapter, keyword, release_version, &
+         trim(new_default), trim(user_value), &
+         keyword_is_specified, keyword_is_relevant, &
+         .not. strcmpi(trim(user_value), trim(new_default)), .true.)
+
+   end subroutine notify_default_change_char
+
+   subroutine notify_default_change_impl(chapter, keyword, release_version, new_default, user_value, &
+                                         keyword_is_specified, keyword_is_relevant, &
```
KoenBussemakerDeltares: I wonder if the `keyword_is_relevant` is overcomplicating things a bit. Isn't it better to just always give a heads up? I'm generally in favor of less conditional magic and just consistently reporting back such things.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -4221,6 +4251,152 @@ subroutine set_output_time_vector(md_tvfil, ti_tv, ti_tv_rel)
...
+         default_value_text = '"'//trim(new_default)//'"'
+         user_value_text = '"'//trim(user_value)//'"'
+      else
+         default_value_text = trim(new_default)
+         user_value_text = trim(user_value)
+      end if
+
+      if (.not. keyword_is_specified) then
+         msgbuf = 'Keyword ['//trim(chapter)//'] '//trim(keyword)//' is not specified. Since release '//trim(release_version)// &
+                  ', the default value is '//trim(default_value_text)//'. Results may differ from older releases.'
```
KoenBussemakerDeltares, line 4335: A warning is generally a non-critical signal that the user needs to take action. Do we want to suggest user action here? Maybe "Consider specifying the value" ?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
JulienGroenenboom: I see your point. Presently it is at the end to make sure other inputs (presently 'kmx' and 'mapformat'), that are used for the conditional check (`keyword_is_relevant`), are already processed.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
JulienGroenenboom: It is making this slightly more complicated, but otherwise 1D/2D-modellers get (irrelevant) messages about 3D-settings. The INFO/WARNING could be made more generic ('if applicable to your model') instead of trying to have this logic in our code. (and then your previous suggestion would also be possible)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
JulienGroenenboom: You recommend switching from WARNING to INFO?

KoenBussemakerDeltares: I think a warning is justified as it has the potential to alter your model output, which is quite a serious consequence. Info messages also tend to be quite spammy so I can image that most users will never read them, or might even hide them using the log level. I fear that many users ignore warnings too, but that's a different story and beyond our control :)

I do think it is nice to give the user a way out of this warning message in case they would like to get rid of it, so that's why I would add the suggestion to specify it in the mdu file.

KoenBussemakerDeltares (approved):
Looks nice and clean now. The last things remaining are minor details so you got my thumbs up. Mainly the Warning vs Info thing.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
KoenBussemakerDeltares: Nice!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
JulienGroenenboom: I also discussed this point with @arthurvd and decided to lower the severity to an INFO-message.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -4221,6 +4248,105 @@ subroutine set_output_time_vector(md_tvfil, ti_tv, ti_tv_rel)
...
+                                         keyword_is_specified, values_differ, quote_values)
+
+      character(len=*), intent(in) :: chapter
+      character(len=*), intent(in) :: keyword
+      character(len=*), intent(in) :: release_version
+      character(len=*), intent(in) :: new_default
+      character(len=*), intent(in) :: user_value
+      logical, intent(in) :: keyword_is_specified
+      logical, intent(in) :: values_differ
+      logical, intent(in) :: quote_values
```
arthurvd: Docstring:
```suggestion
   !> Inform user about changed default values, if applicable. To make users aware
   !! they're not using the currently recommended settings.
   !!
   !! When user has not provided a particular keyword, always inform about the changed
   !! default. When user has provided a particular keyword, only inform if that value
   !! is not the same as the current default. 
   subroutine notify_default_change_impl(chapter, keyword, release_version, new_default, user_value, &
                                         keyword_is_specified, values_differ, quote_values)

      character(len=*), intent(in) :: chapter !< Block name in MDU under which this keyword appeared.
      character(len=*), intent(in) :: keyword !< Keyword name.
      character(len=*), intent(in) :: release_version !< Release version in which the default changed.
      character(len=*), intent(in) :: new_default !< Current default value (string representation)
      character(len=*), intent(in) :: user_value !< Value provided in user input (string representation)
      logical, intent(in) :: keyword_is_specified !< Whether or not a non-empty value was in user's input at all.
      logical, intent(in) :: values_differ !< Whether or not the user's and default value differ.
      logical, intent(in) :: quote_values !< Whether or not to quote the values in the printed messages.
```

arthurvd (commented):
Here's a suggestion for missing docstrings.

arthurvd (approved):

PR #1172, opened 2026-08-06
UNST-10208: Abort the computation when the entrainment factor is negative

adrimourits:
# What was done 

- Added: abort statement
- Added: example NF2FF file
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 
- [x] Tested with NF2FF file `NF2FF__negative_entrainment.xml` 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
[UNST-10208](https://issuetracker.deltares.nl/browse/UNST-10208)


adrimourits:
Hi @AabhasSrivastava , in case you have some leftover-time, and you're in the mood, you can review this mini-PR.

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -260,6 +261,11 @@ namespace pre_c_sumo
             for (std::size_t sink_index = 1; sink_index < sinks.size(); sink_index++)
             {
                 double delta_s = sinks[sink_index].entrainment - sinks[sink_index - 1].entrainment;
+                if (delta_s < 0.0)
```
AabhasSrivastava, line 263: It might be better to use try-catch here. That way, we can also test this functionality with unit test.

AabhasSrivastava: Also, what would happen if delta_s == 0? Do we still follow the same path?

AabhasSrivastava (dismissed):
I would recommend using try-catch and add unit test around it.
Also, is delta_s==0 accepted? There maybe a case with 40 sinks and all having same entrainment - would that be an acceptable input for FM?

src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
adrimourits: Done: try-catch and unitTest.

I'm not sure about delta_s==0. It's not wrong. I can imagine that it can happen.
delta_s checks are not in the DIMR implementation. Therefore I propose to allow delta_s==0 for now. It's up to Wilbert to decide on this.

erwin-mulder: So we are using exceptions to signal errors now? This is not wrong, but we used the whole 'expected' mechanism to catch errors everywhere else. Handling errors using exceptions and value-or-error styles are competing concepts and if I have to be strict, I feel we should stick to one or the other and not start mixing them inside one project.

adrimourits: There are also two `throw`s in `ConnectedSinkSources::write_to_precice`. Do they also need to be changed into `expected`? Do you propose to change all `throw`s as part of this issue or create a new issue?

adrimourits: Separate issue: UNST-10243

erwin-mulder (approved):

PR #1173, opened 2026-08-06
UNST-8322 Fix xymiss misinterpretations for writing nested coordinates

jreyns:
# What was done 

Take into account missing values and ambiguous boundary points in determining open boundaries for nesting in D-Waves

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
closes UNST-8322


jreyns:
image

Grid with cutouts now also handled correctly

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
```diff
@@ -57,35 +57,459 @@ subroutine write_bnd(xc        ,yc        ,mc        ,nc        , &
...
+            if (.not. up .and. .not. lower_left .and. .not. lower_right) then
+                return
+            end if
+        endif
+        swan_valid_boundary_point = .true.
+    end function swan_valid_boundary_point
+
+    subroutine validate_active_topology(status)
+        integer, intent(out) :: status
+        integer :: component_count
```
adrimourits: Proposal: rename to `region_count`

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
adrimourits: Can you add a general description about what is done here (check boundaries, move points a bit if needed) and why (SWAN can not handle ....)

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
```diff
@@ -57,35 +57,459 @@ subroutine write_bnd(xc        ,yc        ,mc        ,nc        , &
...
+    real(kind=hp), parameter :: TOL = 1.0e-6_hp
+    ! Move each location by a small fraction of its incoming edge. This makes
+    ! every child boundary point belong to one nesting segment only; SWAN can
+    ! otherwise see a point at the end of both adjacent segments when
+    ! BOUN NEST is CLOSED (most visibly at corners and local edge kinks).
+    real(kind=hp), parameter :: BOUNDARY_OFFSET_FRACTION = 0.001_hp
 !
 !! executable statements -------------------------------------------------------
 !
     if (inest>1) then
```
adrimourits, line 155: Nice. I like this top-level-subroutine (good readable), delegating all the actual work in sub-subroutines.

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
```diff
@@ -57,35 +57,459 @@ subroutine write_bnd(xc        ,yc        ,mc        ,nc        , &
...
+              kcs(boundary_m(previous_point),boundary_n(previous_point)), &
+              xc(boundary_m(next_point),boundary_n(next_point)), &
+              yc(boundary_m(next_point),boundary_n(next_point)), &
+              kcs(boundary_m(next_point),boundary_n(next_point)))
        enddo
        close(lunbot)
+       deallocate(boundary_m, boundary_n)
     endif
+contains
+    logical function valid_bnd_point(x, y, mask, missing)
```
adrimourits, line 207: I like doxygen descriptions being added to all functions/subroutines. Copilot can generate them.

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
```diff
@@ -57,35 +57,459 @@ subroutine write_bnd(xc        ,yc        ,mc        ,nc        , &
...
 !
-    integer           :: i
-    integer           :: j
+    integer           :: current_point
     integer           :: lunbot
+    integer           :: next_point
+    integer           :: num_boundary_points
+    integer           :: previous_point
+    integer           :: topology_status
+    integer, allocatable :: boundary_m(:)
```
adrimourits: Minor detail: using `dimension in front of the `::` is preferred: `integer, dimension(:), allocatable :: boundary_m`

adrimourits (approved):
Looks nice. All my remarks are minor: ignore them if you prefer.

The only annoying thing:
It seems that we're working around a SWAN issue by modifying the boundary points in D-Waves. What are your ideas about solving it properly in SWAN?

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
jreyns: done

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
jreyns: done

src/engines_gpl/wave/packages/io/src/write_bnd_loc.f90
jreyns: done

jreyns:
Hi @adrimourits 

This PR provides a much more capable nesting administration in D-Waves I think, as we now can:
- Trace the actual active-grid perimeter instead of assuming a simple rectangular outer boundary as we did before.
- Follow cutouts and irregular boundaries, including the additional grid row around missing points.
- Support grids with an entirely inactive outer ring.
- Avoid undefined coordinates at inactive kcs points.
- Detect disconnected regions, isolated cells, and unsupported topology explicitly.
- Produce an ordered, closed boundary consistent with SWAN’s list of valid boundary points.

I think that these improvements are specific to D-Waves because here we know the child grid, its kcs mask, and which coordinates are valid to pass on to SWAN, who has no idea about this, but just gets a list of numbers from D-Waves.

In essence, only the small coordinate offset we use now is specifically compensating for a SWAN defect in double ownership of corner points for closed boundaries, afaics.

marcio-albernaz:
@adrimourits, thanks for the review. 
I see that Johan already covered all points. I just made the testbench green again. Please have a final look and approve, if you agree with all points. NB. I see that Adri already approved it, but since I pushed the test-bench for a failing case, I can't approve/close it now. :)

Regarding the comment about SWAN, I think that this has more to do with the nesting approach from D-Waves than SWAN itself. SWAN already supports the NEST output-input to be used as boundary and we just need to smartly define the output/input point locations, as Johan just improved. From the user point-of-view this is huge improvement as we have had several (silent) problems in the past.

adrimourits (approved):

PR #1178, opened 2026-08-07
UNST-10106 Fixed datetime properties containing underscores in testbench mdus

KoenBussemakerDeltares:
# What was done 

Now that we have DVC I wanted to try batch updating a little bit. There were 20 testbench tests that contained for example `"RestartDateTime = 20030105_000000"`. This underscore is not supported by the dflowfm_io library and we want to remove it.

# Issue link


thomas-pijls (approved):
Great work

PR #1179, opened 2026-08-07
UNST-9110: fix wave forcing custom varnames

mihai-pricope:
# What was done 

Fix varnames for wave forcing

# Evidence of the work done 

- [ ]	Video/figures \
- [X]	Not applicable 

# Tests 
- [X] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-9110


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -850,18 +850,6 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
...
-         ! the name of the source item created by the file reader will be the same as the ext.force. var name
-         if (.not. present(varname)) then !> these variables will crash without a varname
-            write (msgbuf, '(3a)') 'm_meteo::ec_addtimespacerelation: ''dataVariableName'' is required for quantity ''', &
-               trim(target_name), ''' but was not provided. Add dataVariableName= to the [Parameter] block.'
-            call err_flush()
-            goto 1234
-         end if
-         ! TODO: UNST-9110: this is actually introduces a bug: the identification of the source item should be consistent with this
-         ! code here and the code in m_ec_provider::ecProviderCreateNetcdfItems()
-         sourceItemName = varname
```
arthurvd, line 863: For future reference: I now understand why this new code works.
Originally, sourceItemName was set to the user-provided custom name (e.g., 'hs'), and this name was used to call
```
sourceItemIds = ecFindItemsInFileReader(.., sourceItemName) ! so, we're searching by user custom name
```
But, the actual item inside the filereader was created with a quantity that has 
```
ecQuantitySet(instancePtr, quantityId, name=quantityName, ... ! but we've created the filereader using quantityname='wavesignificantheight'
```
This then resulted in items not found-errors.

Now that we step away from the "name-based" getting of source items, and rely on the `select ... case default`, this is used instead:
```
         if (fileReaderPtr%nitems >= 1) then
            sourceItemId = fileReaderPtr%items(1)%ptr%id
```
And this piece of code always finds the item in the file, regardless of its name.
So, as long as we're not using combined quantities (such as for 'airpressure_windx_windy'), this approach will work for the scalar wave quantities.

arthurvd (approved):
Change looks good and I understand why it works. Ready to merge.

arthurvd (approved):

PR #1183, opened 2026-08-11
fix: temporary disable dvc diff job

aslowwriter:
the DVC diff job is taking up too many TC resources, so we need to disable it until we can diagonse and fix that. 
This PR disables this job by adding a condition that should never be true (so we don't lose the trigger logic etc)

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link
DEVOPSDSC-989 / DEVOPSCICD-24

### DIFF
--- a/ci/teamcity/Delft3D/trigger.kt
+++ b/ci/teamcity/Delft3D/trigger.kt
@@ -281,6 +281,12 @@ object Trigger : BuildType({
             conditions {
                 doesNotContain("teamcity.build.triggeredBy", "Snapshot dependency")
                 startsWith("teamcity.build.branch", "pull")
+
+		// only way to disable a step as far as I'm aware
+		// it's taking up too many TC resources, so this is disabled until we can 
+		// diagnose and fix that
+		equals("env.disabled", "false") 
+		
             }
 
             scriptContent = """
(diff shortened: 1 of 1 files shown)


ci/teamcity/Delft3D/trigger.kt
```diff
@@ -281,6 +281,12 @@ object Trigger : BuildType({
             conditions {
                 doesNotContain("teamcity.build.triggeredBy", "Snapshot dependency")
                 startsWith("teamcity.build.branch", "pull")
+
```
ahmad-el-sayed: I think you can use `enabled = false` under `name` attribute.

ahmad-el-sayed (changes requested):

ci/teamcity/Delft3D/trigger.kt
ahmad-el-sayed: this would disable the step

ahmad-el-sayed (approved):
Step successfully disabled.

Image

PR #1186, opened 2026-08-12
UNST-10248: reading also the locationfile

jjnoort:
# What was done 

See issue description

# Evidence of the work done 

- [ ]	Video/figures \
- [x]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10248


FlorisBuwaldaDeltares (commented):
hmm does find_netcells_for_structures also check anything with that polyline file? the only thing you added was a check on its existence.. i'd expect also somewhere where we check whether the structures overlap?

jjnoort:
As far as I understand, the code that follows prevents that partition boundaries cross a structure. It was reported to me, that this code with v3 structure files, didn't work anymore, because the inifile now contains locationFile in stead of polylineFile. So this was an easy fix.

FlorisBuwaldaDeltares:
so if i understand correctly the locationfile has the same content as a polyline file, and thus this routine can read it? it's only the name that changed?

FlorisBuwaldaDeltares (approved):

FlorisBuwaldaDeltares (approved):

PR #1187, opened 2026-08-12
UNST-10210: Resize the sources and sinks mesh

erwin-mulder:
# What was done 
 - Resize the sources and sinks mesh to number of items received from NF2FF files on every exchange.
 -  FM side should already be adjusting the size read. 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10210


src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
```diff
@@ -184,6 +185,20 @@ namespace pre_c_sumo
             ConnectedSinkSources connected_sink_sources =
                 convertNFtoConnectedSinkSources(csumo_settings.value(), nf2ff_readers);
 
+            if (participant.isTimeWindowComplete())
+            {
+                participant.resetMesh(sources_sinks_mesh);
```
adrimourits, line 199: Is it worth checking whether the number of entries has changed (and do nothing if it is the same)?

adrimourits (approved):
Is that all???

src/tools_gpl/pre_c_sumo/src/pre_c_sumo_lib.cpp
erwin-mulder: The FM side does the query all the time too. I don't see why we need to add complexity here unless it causes performance or other kinds of problems.

PR #1188, opened 2026-08-12
Fm/task/unst 10068 generic bc nc support

FlorisBuwaldaDeltares:
# What was done 

- use generic connection fallback for all "simple" quantities. Instead of a specific sourceItemName, the EC module connects its first source item to the target.
- generic fallback checks if file reader, source and target exist
- all "special" quantities retain explicit handling. Rainfall does funky things, and also multi-target quantities
- new boolean `source_connection_created` to determine whether the fallback is necessary
- matrix (!) unit tests for bcascii and .nc support, to test all meteo quantities mentioned in the issue!

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -821,7 +827,7 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
          else if (ec_filetype == provFile_curvi) then
             sourceItemName = 'curvi_source_item_1'
          else
-            call mess(LEVEL_FATAL, 'm_meteo::ec_addtimespacerelation: Unsupported filetype for quantity rainfall.')
+            call mess(LEVEL_FATAL, 'm_meteo::ec_addtimespacerelation: Unsupported filetype for special quantity rainfall.')
```
arthurvd: Why do we tell the user that rainfall is a "special" quantity? (2x)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -1622,4 +1700,321 @@ end subroutine test_field1d_global_value_applied_to_frictioncoefficient
...
+      character(len=*), intent(in) :: filename
+      integer :: ncid, time_dimid, x_dimid, y_dimid, time_varid, x_varid, y_varid
+      integer :: variable_ids(NUM_SCALAR_METEO_CASES)
+      integer :: i, ierr
+      real(dp) :: times(2), x_coord(2), y_coord(2), values(2, 2, 2)
+
+      call check_scalar_meteo_netcdf(nf90_create(filename, NF90_CLOBBER, ncid), 'create NetCDF file')
+      call check_scalar_meteo_netcdf(nf90_def_dim(ncid, 'time', 2, time_dimid), 'define time dimension')
+      call check_scalar_meteo_netcdf(nf90_def_dim(ncid, 'x', 2, x_dimid), 'define x dimension')
+      call check_scalar_meteo_netcdf(nf90_def_dim(ncid, 'y', 2, y_dimid), 'define y dimension')
```
arthurvd, line 1743: Spatially uniform meteo ("global") is not supposed to be defined on a grid, just like the .bc timeseries is global.

Of course we continue to support gridded meteo in NetCDF too. So this unit test is useful, but it's somewhat confusing to have it inside this PR for spatially uniform input. I'd keep it for now, but we'll need a matrix tests for .nc without x,y as well.

arthurvd (commented):
I did a quick incomplete review, have two small remarks to begin with, see comments.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
FlorisBuwaldaDeltares: you are right, i was intending to communicate to the user that for generic quantities all filetypes are automatically supported, and for the special quantities only specific filetypes, but on second thought this is only confusing. I will remove it

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: oh, yes, i didn't notice spatially uniform netcdf support was a blocker for this issue. I was under the assumption that here I just had to implement bc and nc support.
So i'd say we test this now to check whether we can make these netcdf connections, and then once we support spatially uniform input we can add the matrix test for that, and hopefully it will just work automatically since the only difference is inside the EC module?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -20,6 +20,22 @@ module test_init_spatial_fields_integration
...
+      'p140209', 'msl', 't2m', 'tcc', 'd2m', 'rhum', 'slhf', 'strd', 'ssr', 'ssrd', &
+      'sshf', 'tauu', 'tauv', 'u10', 'v10']
+   character(len=64), parameter :: SCALAR_METEO_STANDARD_NAMES(NUM_SCALAR_METEO_CASES) = [character(len=64) :: &
+      'air_density', 'air_pressure', 'air_temperature', 'cloud_area_fraction', 'dew_point_temperature', 'relative_humidity', &
+      'surface_upward_latent_heat_flux', 'surface_net_downward_longwave_flux', 'surface_net_downward_shortwave_flux', &
+      'surface_downwelling_shortwave_flux_in_air', 'surface_upward_sensible_heat_flux', &
+      'surface_downward_eastward_stress', 'surface_downward_northward_stress', 'eastward_wind', 'northward_wind']
+   real(dp), parameter :: SCALAR_METEO_VALUES(NUM_SCALAR_METEO_CASES) = [ &
+      1.2_dp, 101325.0_dp, 20.0_dp, 0.4_dp, 10.0_dp, 60.0_dp, 120.0_dp, 80.0_dp, 200.0_dp, 250.0_dp, &
+      50.0_dp, 0.1_dp, 0.2_dp, 3.0_dp, 4.0_dp]
```
thomas-pijls: This is style / coding guidelines discussion. In the [coding guidelines](https://dpcbuild.deltares.nl/repository/download/DsDoc_DscGuidelines_DscDevelopmentProcessGuidelines/.lastSuccessful/pdf/*.pdf/development_process_documents/DSC_development_coding_guidelines_fortran.pdf) it is stated that we want to declare arrays using `dimension(n) ::` instead of `:: your_array(n)`. 

In some PR's of you Floris I often see the last style of array declaration. I'm not against this, however I was wondering to which style we want to adhere (or allow both). Maybe this is also more a team-level discussion :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -42,6 +58,7 @@ subroutine teardown_minimal_grid()
       if (allocated(yz)) deallocate (yz)
       if (allocated(kcs)) deallocate (kcs)
       if (allocated(rain)) deallocate (rain)
+      if (allocated(air_pressure)) deallocate (air_pressure)
```
thomas-pijls, line 61: To continue, the coding guidelines state that we want to write if statements over multiple lines. This is especially helpful during debugging.

However, for these deallocations I do not mind making them single line. But this is also more a style / team-level discussion

thomas-pijls (commented):
Some general style, coding guideline comments I was wondering about

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: in general i agree with you.. but for if allocated deallocate statements.. i sincerely hope nobody will ever debug them and i don't like making them take up 3 lines each :/

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: you are right, i will adjust it, i should pay more attention to those things

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
thomas-pijls: Yeah precisely, so I agree that for deallocations (and other very simple operations) single line if statements are allowed :)

mihai-pricope:
For the addtimespacerelation refactoring:

I think this change has a bigger probability to crash for an unforseen use case.  The upside, is that the crash will be very clean and we can easily fix it. 

Of course this code is way better than what we had before.  So I will say let's merge it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: done!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1562,30 +1389,90 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
...
+            if (associated(targetItemPtr2) .or. associated(targetItemPtr3) .or. associated(targetItemPtr4)) then
+               call mess(LEVEL_ERROR, 'm_meteo::ec_addtimespacerelation: Generic source connection supports one target item only; quantity '''//trim(target_name)//''' requires special handling.')
+               goto 1234
+            end if
+
+            ! Use the first provider item to identify the logical source quantity,
+            ! then retain all matching items for nested source data.
+            sourceItemName = fileReaderPtr%items(1)%ptr%quantityPtr%name
+            sourceItemIds = ecFindItemsInFileReader(ecInstancePtr, fileReaderId, sourceItemName)
+            if (.not. allocated(sourceItemIds)) then
```
mihai-pricope: this should be:
 if (size(sourceItemIds) == 0) then

it turns out you can't return an unallocated allocatable pointer as a function return.

Was fixed here:
https://github.com/Deltares/Delft3D/pull/1179/changes

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1562,30 +1389,90 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
             ! with nesting there can be more than one source item now. But the first is always the main one
             ! The second is made for nesting to be able to interpolate z-values in time
             sourceItemIds = ecFindItemsInFileReader(ecInstancePtr, fileReaderId, sourceItemName)
-            if (size(sourceItemIds) == 0) then
```
mihai-pricope, line 1565: here too, this should be:
if (size(sourceItemIds) == 0) then

it turns out you can't return an unallocated allocatable pointer as a function return.

mihai-pricope (changes requested):
Only a minor change, I actually fixed in main and it seems it's not picked up here.

mihai-pricope (approved):
Nice refactoring for addtimespcerelation!

PR #1192, opened 2026-08-13
Fm/task/unst 10067 scalar netcdf

FlorisBuwaldaDeltares:
# What was done 

- Added automatic detection of time-dependent scalar NetCDF variables in ec_provider.F90.

- Variables with only the time dimension become scalar source items, uses the existing elmSetType_scalar representation.
- Added scalar NetCDF reading in ec_filereader_read.F90.
- Skip spatial weight construction for scalar sources.
- Time-interpolates scalar NetCDF values.
- Broadcasts one scalar across all target coordinates, reusing part of the datavalueprovider
- Supports multiple scalar sources mapped to multiple target items, such as windxy.
- Full scalar NetCDF matrix covering 15 meteo quantities.
- Full gridded NetCDF matrix covering the same quantities.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/utils_lgpl/ec_module/packages/ec_module/src/ec_converter.f90
```diff
@@ -4228,8 +4300,70 @@ subroutine ConvertToSparseIndices(n_points, indices, n_rows, ia, ja)
...
+      else
+         n_coordinates = connection%targetItemsPtr(target_item_index)%ptr%elementSetPtr%nCoordinates
+         if (n_coordinates <= 0 .or. modulo(size(targetField%arr1dPtr), n_coordinates) /= 0) then
+            call set_ec_message("ERROR: ec_converter::ecConverterApplyScalarToTargetItem: Target Field size is inconsistent with its coordinates.")
+            return
+         end if
+         first_coordinate = 1
+         last_coordinate = n_coordinates
+         values_per_coordinate = size(targetField%arr1dPtr) / n_coordinates
+      end if
```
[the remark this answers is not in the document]
FlorisBuwaldaDeltares: good point, fixed

src/utils_lgpl/ec_module/packages/ec_module/src/ec_filereader_read.F90
```diff
@@ -768,6 +768,49 @@ function ecNetcdfReadNextBlock(fileReaderPtr, item, t0t1, timesndx) result(succe
...
+                  return
+               end if
+
+               valid_field = (fieldPtr%arr1dPtr(1) /= dmiss_nc)
+               if (.not. valid_field .and. timesndx < fileReaderPtr%tframe%nr_timesteps) then
+                  timesndx = timesndx + 1
+               else
+                  valid_field = .true.
+               end if
+            end do
```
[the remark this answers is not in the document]
FlorisBuwaldaDeltares: you are right, we should indeed not advance the shared timesndx

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -2952,7 +2954,11 @@ function ecProviderCreateNetcdfItems(instancePtr, fileReaderPtr, quantityName, v
             end do
          end if ! has non-empty coordinates attribute
 
-         if (fgd_id < 0 .or. sgd_id < 0) then
+         is_scalar_source = (tim_dimid > 0 .and. ndims == 1 .and. dimids(1) == tim_dimid .and. &
+                             fgd_id < 0 .and. sgd_id < 0)
```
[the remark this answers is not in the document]
FlorisBuwaldaDeltares: good catch, suggestion applied

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
rene-deltares, line 1: I'm not requesting changes, but I feel like some of the tests (not the new ones in particular, and also my own ones) are getting a bit too low-level. Because this file is getting very long, as are the test subroutines.

It would be great if we could, for instance, create standard compliant netcdf files, without having to hardcode the types, dimensions and all of the required attributes. I think a helper could be great here, and it may be useful in other test modules, and future tests. Also I'd like to find a way to encapsulate the global state more.

Other than that minor nitpick the tests look good to me

src/utils_lgpl/ec_module/packages/ec_module/src/ec_filereader_read.F90
```diff
@@ -803,16 +836,6 @@ function ecNetcdfReadNextBlock(fileReaderPtr, item, t0t1, timesndx) result(succe
 
          valid_field = (col1 == 0 .and. row1 == 0)
          do while (.not. valid_field)
-            ! - Read a scalar data block.
-            if (item%elementSetPtr%nCoordinates == 0) then
```
rene-deltares, line 807: I guess the removed block of code here is never reached now, because it is handled by the new if statement above?

src/utils_lgpl/ec_module/packages/ec_module/src/ec_provider.F90
```diff
@@ -2755,6 +2757,8 @@ function ecProviderCreateNetcdfItems(instancePtr, fileReaderPtr, quantityName, v
       name = ''
       ndims = 0
       rotate_pole = .false.
+      is_scalar_source = .false.
+      has_horizontal_coordinates = .false.
```
rene-deltares, line 2761: I like thess

rene-deltares (approved):
Ok, I went through and tried to understand the changes and to me they look good.

src/utils_lgpl/ec_module/packages/ec_module/src/ec_filereader_read.F90
FlorisBuwaldaDeltares: yes!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: I agree with you. If you're okay with it let's pick that up in a followup issue, because I indeed feel maybe test_init_spatial_fields_integration is getting a bit big and should be split-up. Also, I think some of the tests that were a good addition in the past might have outlived their usefulness.

PR #1201, opened 2026-08-18
INCIDENT: Increase maxRunTime amount of several new test cases, that were merged just after maxRunTime was enforced.

rene-deltares:
There were some new test cases added that had the 'maxRunTime' property set too low (they time out every time).
Also it doesn't make sense to set the maxRunTime lower than 480. The default is 300 and if any test case needs more time than that it makes sense to set it at least three minutes above five minutes, because the maxRunTime is just there to avoid processes from hanging indefinitely. The timeout is not meant as a check on performance degradation.

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


FlorisBuwaldaDeltares (approved):
i didn't check each individual case to see if it was indeed flaky, but i trust this is an improvement!

PR #1206, opened 2026-08-20
don't let dzTop overrule the stretchType for non z-layer models

hrajagers:
# What was done 

D-Flow FM distinguishes for the vertical layer distribution the following options:
- `STRETCH_UNIFORM` (0)
- `STRETCH_USER` (1)
- `STRETCH_EXPONENTIAL` (2) and
- `STRETCH_UNI_OVER_EXP` (-1) -- undocumented.
The last one was the default initialization and was enforced whenever `dzTop` was specified, even though it's not compatible with sigma-layers where it effectively fell back to `STRETCH_UNIFORM`.

In the new situation, `STRETCH_UNIFORM` is the default for all non z-layer models; the `STRETCH_UNI_OVER_EXP` option cannot be selected.
The `STRETCH_UNI_OVER_EXP` option is the default for z-layer models and is also used (for z-layer models) when `STRETCH_UNIFORM` is requested (if we change this behaviour, we will have to update test case e02_f018_c030 whicih specifies `STRETCH_UNIFORM` but uses `STRETCH_UNI_OVER_EXP`).

Some refactoring of the code related to the processing of the `stretchType` for the vertical layer distribution -- effectively no code change except for the change mentioned above. Renamed the variable `iStrchTyp` to `stretch_type` and consistently used parameters for the different options. `STRETCH_UNDEFINED` (-999) added to allow for initialization depending on user selected layer-type.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

UNST-10265

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
@@ -773,16 +773,16 @@ contains
       call prop_get(md_ptr, 'geometry', 'ZlayTop', zlaytop)
       call prop_get(md_ptr, 'geometry', 'StretchType', stretch_type)
       
-      if (layertype /= LAYTP_Z) then
-         if (stretch_type == STRETCH_UNDEFINED .or. stretch_type == STRETCH_UNI_OVER_EXP) then
-            stretch_type = STRETCH_UNIFORM
-         end if
-      else
+      if (layertype == LAYTP_Z) then
          if (stretch_type == STRETCH_UNDEFINED) then
             stretch_type = STRETCH_UNI_OVER_EXP
          elseif (dztop > 0.0_dp .and. stretch_type /= STRETCH_UNI_OVER_EXP) then
             call mess(LEVEL_WARN, 'The stretchType is reset to -1 because a strictly positive dzTop is specified.')
             stretch_type = STRETCH_UNI_OVER_EXP
+         end if
+      else
+         if (stretch_type == STRETCH_UNDEFINED .or. stretch_type == STRETCH_UNI_OVER_EXP) then
+            stretch_type = STRETCH_UNIFORM
          end if
       end if
 
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
@@ -480,7 +480,7 @@ contains
       !                         javakeps,                                                             &
       !                         fixedweirtopwidth, fixedweirtopfrictcoef, fixedweirtalud, ifxedweirfrictscheme,  &
       !                         Tsigma, jarhoxu,                                                      &
-      !                         iStrchType, STRCH_UNIFORM, STRCH_USER, STRCH_EXPONENT, STRCH_FIXLEVEL, laycof
+      !                         stretch_type, STRCH_UNIFORM, STRETCH_USER, STRETCH_EXPONENT, laycof
 
       use m_globalparameters, only: sl
       use m_flowgeom !,              only : wu1Duni, bamin, rrtol, jarenumber, VillemonteCD1, VillemonteCD2
@@ -749,8 +749,8 @@ contains
       kmx = 0
       call prop_get(md_ptr, 'geometry', 'Kmx', kmx)
 
-      call prop_get(md_ptr, 'geometry', 'Layertype', Layertype)
-      if (Layertype /= LAYTP_SIGMA) then
+      call prop_get(md_ptr, 'geometry', 'Layertype', layertype)
+      if (layertype /= LAYTP_SIGMA) then
          mxlayz = kmx
       end if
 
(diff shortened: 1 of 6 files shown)


marcio-albernaz:
well done, Bert! Increadible that this was hidden in the code for such a long time.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -480,7 +480,7 @@ subroutine readMDUFile(filename, istat)
       !                         javakeps,                                                             &
       !                         fixedweirtopwidth, fixedweirtopfrictcoef, fixedweirtalud, ifxedweirfrictscheme,  &
       !                         Tsigma, jarhoxu,                                                      &
-      !                         iStrchType, STRCH_UNIFORM, STRCH_USER, STRCH_EXPONENT, STRCH_FIXLEVEL, laycof
+      !                         stretch_type, STRCH_UNIFORM, STRETCH_USER, STRETCH_EXPONENT, laycof
```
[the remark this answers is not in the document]
hrajagers: 1) This comment has been updated. Also included newly used parameters.
2) The comments in lines 780 and 2864 refer to user interaction. The user doesn't know about these internal names. We should introduce strings instead of magic numbers, but that's a change beyond the current scope that is not overruling the user set stretch_type unless we're in a z-layer model.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/flow_allocflow.f90
```diff
@@ -359,45 +359,50 @@ subroutine flow_allocflow()
...
+               ! Fill layers based on dzslay computed above
                zslay(0, j) = 0.0_dp
                do k = 1, mx
                   zslay(k, j) = zslay(k - 1, j) + dzslay(k, j)
                end do
 
-            else if (laytyp(j) == 2) then
+            else if (laytyp(j) == LAYTP_Z) then
 
                call realloc(zslay, uindex=[mx, mxlaydefs], lindex=[0, 1], stat=ierr, keepexisting=.false.) ! nr of layer distributions
```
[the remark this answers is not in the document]
hrajagers: Removed this `realloc` call. Update the initial `realloc` call to include initialization to `dmiss`.

hrajagers (commented):
Addressed Copilot review comments.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -771,10 +772,19 @@ subroutine readMDUFile(filename, istat)
...
-
-      if (Dztop > 0.0_dp) then ! hk claims back original functionality
-         iStrchType = -1
+      call prop_get(md_ptr, 'geometry', 'StretchType', stretch_type)
+      
+      if (layertype == LAYTP_Z) then
+         if (stretch_type == STRETCH_UNDEFINED) then
+            stretch_type = STRETCH_UNI_OVER_EXP
+         elseif (dztop > 0.0_dp .and. stretch_type /= STRETCH_UNI_OVER_EXP) then
+            call mess(LEVEL_WARN, 'The stretchType is reset to -1 because a strictly positive dzTop is specified.')
```
FlorisBuwaldaDeltares: Does the user know -1? also "reset" is a strange verb to use here. Also the term 'strictly' feels like a copilot-ism, Why not something like:

"warning: a positive dztop requires stretchType uniform over exponential. input value: //stretch_type // is ignored" 

or something like that.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -771,10 +772,19 @@ subroutine readMDUFile(filename, istat)
...
       call prop_get(md_ptr, 'geometry', 'ZlayBot', zlaybot)
       call prop_get(md_ptr, 'geometry', 'ZlayTop', zlaytop)
-      call prop_get(md_ptr, 'geometry', 'StretchType', iStrchType)
-
-      if (Dztop > 0.0_dp) then ! hk claims back original functionality
-         iStrchType = -1
+      call prop_get(md_ptr, 'geometry', 'StretchType', stretch_type)
+      
+      if (layertype == LAYTP_Z) then
+         if (stretch_type == STRETCH_UNDEFINED) then
```
FlorisBuwaldaDeltares: I think the rest of the keywords here don't use stretch_type == UNDEFINED, but simply use the success status of prop_get to determine whether the user supplied the keyword.

However I think this is quite clear too so no need to change it, just be aware it is now inconsistent.

FlorisBuwaldaDeltares (approved):
Hi Bert, nice work! two tiny suggestions but up to you if you want to do anything with them.

Additionally we could add a unit test for this but I guess the integration tests already cover it?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
hrajagers: I hadn't realized that inconsistency, but you're correct. Since I had to change the other line anyway, I've now changed it such that success flag is used.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
hrajagers: The default option -1 is unfortunately not documented (I expect that that was a choice rather than oversight - to check). However, not indicating that the user gets something else that he/she selected, is not an option. I have added the "uniform over exponential" for some guidance.

I added the 'strictly' to explicitly indicate that a zero value isn't affecting the behaviour. I have now dropped it.

hrajagers (commented):
Code updated in line with review suggestions.

FlorisBuwaldaDeltares (approved):

hrajagers:
@FlorisBuwaldaDeltares I didn't reply yet to "Additionally we could add a unit test for this but I guess the integration tests already cover it?". Since the check is in the middle of `unstruc_model` it's a bit hard to test this in a unit test (especially without IO). Since we'll soon have a major makeover of the mdu-file reading, such test would likely have to be refactored again. The integration tests partially cover the new code. I will create a sigma-layer test and add it via a separate pull request. Issue reopened.

PR #1208, opened 2026-08-20
UNST-9805: Added extra check routine to detect if an NF2FF file is complete

erwin-mulder:
Added extra check routine to detect if an NF2FF file is complete, by checking for the end tag.

# What was done 

We added extra function that opens a file and checks if the end tag `</NF2FF>` is contained in the file.
 
# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-9805


src/tools_gpl/pre_c_sumo/src/coupling_steps.cpp
```diff
@@ -130,14 +132,28 @@ namespace pre_c_sumo
         }
     }
 
+    bool is_complete_nf2ff_file(std::filesystem::path filename)
```
AabhasSrivastava, line 135: Wouldn't it be nicer to use pugixml to parse the file and check
- the load file is successful
- check that the root element is NF2FF

It would then autmatically validate the xml structure. It is possible that someone just puts NF2FF closing line - that does not mean that the file is correct (pugixml should verify that the file structure is valid - do not know if it has such an option).
Of course parsing through pugixml would fail if the file is still being written - but maybe we can see/view some sort of file lock? In that case our wait loop to retry reading should keep trying - but maybe we should then add a time out , like 1000 seconds as default, but configurable either via commandline or from the csumo-config.xml.

AabhasSrivastava: bool is_complete_nf2ff_file(const std::filesystem::path& filename)
{
    pugi::xml_document doc;
    pugi::xml_parse_result result = doc.load_file(filename.string().c_str());

    if (!result)
    {
        return false;
    }

    auto root = doc.first_child();
    return root && std::string_view(root.name()) == "NF2FF";
}

Copilot suggested code - if using pugixml.

erwin-mulder: While possible, it is a load of compute to check basically the existence of a single tag.

AabhasSrivastava: Agreed.

AabhasSrivastava (approved):
Looks good. We can point the timeout variable to Wilbert so he can change it when building it. (and if required - create an issue to make it dynamic from cli or csumo-config)

PR #1210, opened 2026-08-20
UNST-9955: Precision of rounding cll files

thomas-pijls:
# What was done 

- Functional change: line 706 and 707 of `calibration.f90`, used comparereal to account for machine precision errors in comparison, included a tolerance to check for that scales linearly with the number of area fractions
- Cleaned up comments in `calibration.f90`
- Cleaned up `scannr.f90` and put the subroutine in the new module `m_scannr`
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_utils/calibration.f90
```diff
@@ -761,7 +702,7 @@ subroutine read_cllfile(md_cllfile, clddata, phase)
             clddata%rttar(icll) = rfield(5)
             clddata%linar(icll) = mcurec
             clddata%sumar(L) = clddata%sumar(L) + clddata%rttar(icll)
-            if (clddata%sumar(L) > 1.0_dp) then
+            if (comparereal(clddata%sumar(L), 1.0_dp) == 1) then
```
FlorisBuwaldaDeltares: please use equals instead of comparereal, it is newer and more robust:

Image

FlorisBuwaldaDeltares: oh wait comparereal returns whether the number is bigger (signified with the == 1)

so it would be equivalent to `a > b .and. .not. equal(a, b)`

FlorisBuwaldaDeltares: I suppose then comparereal can stay, but just be aware

FlorisBuwaldaDeltares (approved):
looks good, nice to do some extra cleanup too

PR #1211, opened 2026-08-21
Speed up big enclosure polygons using binning algorithm

FlorisBuwaldaDeltares:
# What was done 

-added a binning algorithm to reduce workload for large (for dcsm: 90k) enclosure polygons. Since every cell scans every polygon, the current brute-force raycasting implementation is very efficient for many small polygons and extremely inefficient for a single super large one

EDIT:

-refactored towards an OOP implementation. Instead of a global polyline cache, we now have a polyline geometry object with a nice default constructor and type-bound procedures.
-Also a separate netcell geometry, and the binning implementation also has its own type, constructor and procedures.
 

### Problem

Point-in-polygon normally checks every polygon edge for every model cell:

$$
\text{work} \approx \text{cells} \times \text{polygon edges}
$$

For the DCSM enclosure, this meant checking approximately 93,000 edges for almost every cell because the enclosure bounding box covers the complete model.

### Approach

Enclosure polygons are divided into horizontal latitude bins. Each bin stores only the edges whose vertical extent intersects that bin. This significantly reduces the edges that need to be checked for intersection by the raycast algorithm. In this example, the binning excludes e2 and e4 from consideration because the ray will never cross them anyway:

image

For a query point, its latitude identifies one bin. The existing ray-casting algorithm then examines only the edges stored in that bin.

```mermaid
flowchart LR
    A[Cell center] --> B[Find latitude bin]
    B --> C[Retrieve candidate edges]
    C --> D[Existing ray-casting calculation]
    D --> E[Inside or outside]
```

The polygon geometry is not simplified or approximated.

### Construction

```mermaid
flowchart TD
    A[Read enclosure polygons] --> B[Choose latitude bins]
    B --> C[Determine bins crossed by each edge]
    C --> D{Index storage bounded?}
    D -- No --> E[Use original full scan]
    D -- Yes --> F[Store edge numbers per bin]
    F --> G[Use binned ray casting]
```

Edges are inserted in their original polygon order. Therefore, the indexed ray cast processes its candidate edges in the same order as the original implementation.

### Safety Fallback

An edge crossing several bins is stored once in every crossed bin. Normally, short coastline edges occur in only one or two bins.

For pathological polygons containing many long vertical edges, storage could grow excessively. If the average exceeds eight memberships per edge, index creation is abandoned and the original full scan is used.

This fallback affects performance and memory use only, not results.

### Scope

Binning is enabled only for `GridEnclosureFile` processing. Dry-area polygons continue using the original implementation because their smaller bounding boxes already reject most cells efficiently.

### Validation

The implementation was checked by running the original and binned algorithms independently on the complete DCSM model and comparing their resulting masks.

- Masks: identical
- Original enclosure processing: approximately 196 seconds
- Binned enclosure processing: approximately 1.2 seconds

# Evidence of the work done 
before:
image

after:

image

added a full comparison test that reported 0 mismatches for DCSM-FM_100m.mdu
image

image

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
@@ -28,6 +28,7 @@
 !-------------------------------------------------------------------------------
 
 module m_cellmask_from_polygon_set
+   use iso_fortran_env, only: int64
    use m_missing, only: jins, dmiss
    use precision, only: dp
 
@@ -49,19 +50,30 @@ module m_cellmask_from_polygon_set
    real(kind=dp), allocatable :: x_poly_max(:), y_poly_max(:) !< Polygon bounding box max coordinates, (dim = polygons)
    real(kind=dp), allocatable :: polygon_type(:) !< Polygon type, positive or dmiss = drypoint , negative = enclosure (dim = polygons)
    integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
+   logical, allocatable :: edge_index_enabled(:) !< Whether each polygon uses the binned edge index.
+   integer, allocatable :: i_poly_bin_start(:), i_poly_num_bins(:) !< Start and number of bins for each polygon.
+   integer, allocatable :: edge_bin_offsets(:), edge_indices(:) !< CSR offsets and ordered local edge indices.
+   real(kind=dp), allocatable :: y_poly_bin_scale(:) !< Scale from polygon y-coordinate to bin index.
    logical :: cellmask_initialized = .false. !< Flag indicating if cellmask data structures have been initialized for safety
    logical :: enclosures_present = .false. !< Flag indicating if any enclosures are present in the polygon dataset
+
+   integer, parameter :: min_edges_for_index = 256 !< Small polygons are cheaper to scan directly.
+   integer, parameter :: max_edge_bins = 1024 !< Maximum number of latitude bins per polygon.
+   integer(kind=int64), parameter :: max_memberships_per_edge = 8_int64 !< Memory/work cap for vertically long edges.
+   real(kind=dp), parameter :: min_predicted_speedup = 2.0_dp !< Required estimated improvement before indexing.
 
 contains
 
    !> Initialize module-level cellmask polygon data structures, such as the bounding boxes, cache and iistart/iiend
    ! this keeps the actual calculation routines elemental.
-   subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly)
+   subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly, num_query_points, x_query, y_query)
       use m_alloc
       use geometry_module, only: get_startend
 
       integer, intent(in) :: polygon_points !< Number of polygon points
       real(kind=dp), intent(in) :: x_poly(polygon_points), y_poly(polygon_points), z_poly(polygon_points) !< Polygon coordinate arrays
+      integer, intent(in), optional :: num_query_points !< Number of points that will be classified.
+      real(kind=dp), intent(in), optional :: x_query(:), y_query(:) !< Points that will be classified, used to estimate index benefit.
 
       integer :: i_point, i_start, i_end, i_poly
 
@@ -122,6 +134,10 @@ contains
       call realloc(i_poly_start, polygons, keepExisting=.true.)
       call realloc(i_poly_end, polygons, keepExisting=.true.)
       call realloc(polygon_type, polygons, keepExisting=.true.)
+
+      if (present(num_query_points) .and. present(x_query) .and. present(y_query)) then
+         call init_polygon_edge_index(num_query_points, x_query, y_query)
+      end if
 
       ! check if there are any enclosure polygons
       do i_poly = 1, polygons
@@ -220,6 +236,24 @@ contains
       if (allocated(i_poly_end)) then
          deallocate (i_poly_end)
       end if
+      if (allocated(edge_index_enabled)) then
+         deallocate (edge_index_enabled)
+      end if
+      if (allocated(i_poly_bin_start)) then
+         deallocate (i_poly_bin_start)
+      end if
+      if (allocated(i_poly_num_bins)) then
+         deallocate (i_poly_num_bins)
+      end if
+      if (allocated(edge_bin_offsets)) then
+         deallocate (edge_bin_offsets)
+      end if
+      if (allocated(edge_indices)) then
+         deallocate (edge_indices)
+      end if
+      if (allocated(y_poly_bin_scale)) then
+         deallocate (y_poly_bin_scale)
+      end if
 
       polygons = 0
       cellmask_initialized = .false.
@@ -235,17 +269,199 @@ contains
       integer, intent(in) :: i_poly !< Polygon index
       logical :: is_inside !< Result
 
-      integer :: i_start, i_end, n_points
+      integer :: i_bin, i_start, i_end, n_points
+      logical :: use_edge_index
 
       ! Get bounds for this polygon from module arrays
       i_start = i_poly_start(i_poly)
       i_end = i_poly_end(i_poly)
       n_points = i_end - i_start + 1
 
-      ! Call the shared optimized algorithm with array slice
-      is_inside = pinpok_raycast(x, y, xpl_cache(i_start:i_end), ypl_cache(i_start:i_end), n_points)
+      use_edge_index = .false.
+      if (allocated(edge_index_enabled)) then
+         use_edge_index = edge_index_enabled(i_poly)
+      end if
+
+      if (use_edge_index) then
+         i_bin = i_poly_bin_start(i_poly) + get_edge_bin(y, y_poly_min(i_poly), y_poly_bin_scale(i_poly), i_poly_num_bins(i_poly)) - 1
+         is_inside = pinpok_raycast(x, y, xpl_cache(i_start:i_end), ypl_cache(i_start:i_end), n_points, &
+                                   edge_indices(edge_bin_offsets(i_bin):edge_bin_offsets(i_bin + 1) - 1))
+      else
+         is_inside = pinpok_raycast(x, y, xpl_cache(i_start:i_end), ypl_cache(i_start:i_end), n_points)
+      end if
 
    end function pinpok_elemental
+
+   !> Build latitude-bin edge indices only where the observed query distribution predicts a worthwhile reduction.
+   subroutine init_polygon_edge_index(num_query_points, x_query, y_query)
+      integer, intent(in) :: num_query_points !< Number of points that will be classified.
+      real(kind=dp), intent(in) :: x_query(:), y_query(:) !< Points that will be classified.
+
+      integer :: bin_edge_counts(max_edge_bins), bin_query_counts(max_edge_bins), bin_write_positions(max_edge_bins)
+      integer :: bins, edge, edge_end, edge_start, i_bin, i_poly, point
+      integer :: total_bins, write_position
+      integer(kind=int64) :: memberships, total_memberships
+      real(kind=dp) :: candidate_work, direct_work, indexed_work
+
+      allocate (edge_index_enabled(polygons), i_poly_bin_start(polygons), i_poly_num_bins(polygons), y_poly_bin_scale(polygons))
+      edge_index_enabled = .false.
+      i_poly_bin_start = 0
+      i_poly_num_bins = 0
+      y_poly_bin_scale = 0.0_dp
+
+      total_bins = 0
+      total_memberships = 0_int64
+
+      do i_poly = 1, polygons
+         edge_start = i_poly_start(i_poly)
+         edge_end = i_poly_end(i_poly)
+         edge = edge_end - edge_start + 1
+         if (edge < min_edges_for_index .or. y_poly_max(i_poly) <= y_poly_min(i_poly)) then
+            cycle
+         end if
+
+         bins = min(max_edge_bins, max(16, edge / 4))
+         y_poly_bin_scale(i_poly) = real(bins, dp) / (y_poly_max(i_poly) - y_poly_min(i_poly))
+
+         memberships = count_edge_memberships(i_poly, bins)
+         if (memberships > max_memberships_per_edge * int(edge, int64)) then
+            cycle
+         end if
+
+         call count_edges_per_bin(i_poly, bins, bin_edge_counts)
+         bin_query_counts(1:bins) = 0
+         do point = 1, min(num_query_points, size(x_query), size(y_query))
+            if (x_query(point) < x_poly_min(i_poly) .or. x_query(point) > x_poly_max(i_poly) .or. &
+                y_query(point) < y_poly_min(i_poly) .or. y_query(point) > y_poly_max(i_poly)) then
+               cycle
+            end if
+            i_bin = get_edge_bin(y_query(point), y_poly_min(i_poly), y_poly_bin_scale(i_poly), bins)
+            bin_query_counts(i_bin) = bin_query_counts(i_bin) + 1
+         end do
+
+         direct_work = real(sum(bin_query_counts(1:bins)), dp) * real(edge, dp)
+         candidate_work = 0.0_dp
+         do i_bin = 1, bins
+            candidate_work = candidate_work + real(bin_query_counts(i_bin), dp) * real(bin_edge_counts(i_bin), dp)
+         end do
+         indexed_work = real(edge + sum(bin_query_counts(1:bins)), dp) + 2.0_dp * real(memberships, dp) + candidate_work
+
+         if (direct_work < min_predicted_speedup * indexed_work) then
+            cycle
+         end if
+
+         edge_index_enabled(i_poly) = .true.
+         i_poly_num_bins(i_poly) = bins
+         total_bins = total_bins + bins
+         total_memberships = total_memberships + memberships
+      end do
+
+      if (total_bins == 0) then
+         deallocate (edge_index_enabled, i_poly_bin_start, i_poly_num_bins, y_poly_bin_scale)
+         return
+      end if
+
+      allocate (edge_bin_offsets(total_bins + 1), edge_indices(int(total_memberships)))
+      i_bin = 1
+      write_position = 1
+      do i_poly = 1, polygons
+         if (.not. edge_index_enabled(i_poly)) then
+            cycle
+         end if
+         bins = i_poly_num_bins(i_poly)
+         call count_edges_per_bin(i_poly, bins, bin_edge_counts)
+         i_poly_bin_start(i_poly) = i_bin
+         do edge = 1, bins
+            edge_bin_offsets(i_bin) = write_position
+            write_position = write_position + bin_edge_counts(edge)
+            i_bin = i_bin + 1
+         end do
+      end do
+      edge_bin_offsets(total_bins + 1) = write_position
+
+      do i_poly = 1, polygons
+         if (.not. edge_index_enabled(i_poly)) then
+            cycle
+         end if
+         bins = i_poly_num_bins(i_poly)
+         do i_bin = 1, bins
+            bin_write_positions(i_bin) = edge_bin_offsets(i_poly_bin_start(i_poly) + i_bin - 1)
+         end do
+
+         edge_start = i_poly_start(i_poly)
+         edge_end = i_poly_end(i_poly)
+         do edge = 1, edge_end - edge_start + 1
+            call edge_bin_range(i_poly, edge, bins, bin_edge_counts(1), bin_edge_counts(2))
+            do i_bin = bin_edge_counts(1), bin_edge_counts(2)
+               edge_indices(bin_write_positions(i_bin)) = edge
+               bin_write_positions(i_bin) = bin_write_positions(i_bin) + 1
+            end do
+         end do
+      end do
+
+   end subroutine init_polygon_edge_index
+
+   !> Count the storage needed by a polygon edge index without allocating or iterating over memberships.
+   function count_edge_memberships(i_poly, bins) result(memberships)
+      integer, intent(in) :: i_poly, bins !< Polygon index and number of bins.
+      integer(kind=int64) :: memberships !< Total edge-to-bin memberships.
+
+      integer :: bin_first, bin_last, edge, num_edges
+
+      memberships = 0_int64
+      num_edges = i_poly_end(i_poly) - i_poly_start(i_poly) + 1
+      do edge = 1, num_edges
+         call edge_bin_range(i_poly, edge, bins, bin_first, bin_last)
+         memberships = memberships + int(bin_last - bin_first + 1, int64)
+         if (memberships > max_memberships_per_edge * int(num_edges, int64)) then
+            return
+         end if
+      end do
+   end function count_edge_memberships
+
+   !> Count candidate edges in each latitude bin for one polygon.
+   subroutine count_edges_per_bin(i_poly, bins, bin_counts)
+      integer, intent(in) :: i_poly, bins !< Polygon index and number of bins.
+      integer, intent(out) :: bin_counts(:) !< Edge count per bin.
+
+      integer :: bin_first, bin_last, edge, i_bin, num_edges
+
+      bin_counts(1:bins) = 0
+      num_edges = i_poly_end(i_poly) - i_poly_start(i_poly) + 1
+      do edge = 1, num_edges
+         call edge_bin_range(i_poly, edge, bins, bin_first, bin_last)
+         do i_bin = bin_first, bin_last
+            bin_counts(i_bin) = bin_counts(i_bin) + 1
+         end do
+      end do
+   end subroutine count_edges_per_bin
+
+   !> Get the inclusive bin range intersected by an edge.
+   pure subroutine edge_bin_range(i_poly, edge, bins, bin_first, bin_last)
+      integer, intent(in) :: i_poly, edge, bins !< Polygon index, local edge index and number of bins.
+      integer, intent(out) :: bin_first, bin_last !< First and last intersected bins.
+
+      integer :: current_point, previous_point
+
+      current_point = i_poly_start(i_poly) + edge - 1
+      previous_point = current_point - 1
+      if (edge == 1) then
+         previous_point = i_poly_end(i_poly)
+      end if
+      bin_first = get_edge_bin(min(ypl_cache(previous_point), ypl_cache(current_point)), y_poly_min(i_poly), &
+                               y_poly_bin_scale(i_poly), bins)
+      bin_last = get_edge_bin(max(ypl_cache(previous_point), ypl_cache(current_point)), y_poly_min(i_poly), &
+                              y_poly_bin_scale(i_poly), bins)
+   end subroutine edge_bin_range
+
+   !> Map a y-coordinate to a clamped one-based latitude bin.
+   pure integer function get_edge_bin(y, y_min, bin_scale, bins) result(i_bin)
+      real(kind=dp), intent(in) :: y, y_min, bin_scale !< Coordinate, polygon minimum and bin scale.
+      integer, intent(in) :: bins !< Number of bins.
+
+      i_bin = int((y - y_min) * bin_scale) + 1
+      i_bin = max(1, min(bins, i_bin))
+   end function get_edge_bin
 
    !> Manually init geometry cache (used for dry points, test_pol_to_cellmask)
    subroutine init_geom_cache(npl_init, xpl_init, ypl_init, zpl_init)
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
@@ -79,7 +79,7 @@ module m_cellmask_from_polygon_set
       type(t_polygon_geometry) :: geometry
       type(t_binned_edge_index) :: edge_index
    contains
-      procedure :: point_mask => polygon_set_point_mask
+      procedure :: is_masked => polygon_set_is_masked
       procedure :: polygon_contains_point => polygon_set_polygon_contains_point
       procedure :: find_netcell => polygon_set_find_netcell
       procedure :: find_cells_crossed_by_polyline => polygon_set_find_cells_crossed_by_polyline
@@ -205,7 +205,7 @@ contains
    end function construct_polygon_geometry
 
    !> Check if a point should be masked, either is_inside a dry-area polygon or outside an enclosure polygon.
-   elemental function polygon_set_point_mask(this, x, y) result(mask)
+   elemental function polygon_set_is_masked(this, x, y) result(mask)
       class(t_polygon_set), intent(in) :: this
       integer :: mask
       real(kind=dp), intent(in) :: x, y !< Point coordinates
@@ -256,7 +256,7 @@ contains
          end if
       end associate
 
-   end function polygon_set_point_mask
+   end function polygon_set_is_masked
 
    !> Test whether a point lies inside one cached polygon.
    elemental function polygon_set_polygon_contains_point(this, x, y, i_poly) result(is_inside)
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
@@ -87,7 +87,7 @@ module m_cellmask_from_polygon_set
 
    interface t_polygon_set
       module procedure construct_polygon_set
-      module procedure construct_netcell_polygon_cache
+      module procedure construct_netcell_polygon_set
    end interface t_polygon_set
 
    integer, parameter :: max_edge_bins = 1024 !< Maximum number of latitude bins per polygon.
@@ -452,7 +452,7 @@ contains
    end function binned_edge_index_get_bin
 
    !> Construct a polygon cache containing all net-cell geometries.
-   function construct_netcell_polygon_cache() result(cache)
+   function construct_netcell_polygon_set() result(cache)
       use network_data
       use m_alloc
 
@@ -492,7 +492,7 @@ contains
 
       cache = t_polygon_set(xpl_init, ypl_init, zpl_init, enable_binning=.false.)
 
-   end function construct_netcell_polygon_cache
+   end function construct_netcell_polygon_set
 
 !> Fast replacement for INCELLS using cached net-cell geometry.
    elemental function polygon_set_find_netcell(this, x, y) result(k)
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
@@ -50,7 +50,7 @@ module m_cellmask_from_polygon_set
    real(kind=dp), allocatable :: x_poly_max(:), y_poly_max(:) !< Polygon bounding box max coordinates, (dim = polygons)
    real(kind=dp), allocatable :: polygon_type(:) !< Polygon type, positive or dmiss = drypoint , negative = enclosure (dim = polygons)
    integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
-   logical, allocatable :: edge_index_enabled(:) !< Whether each polygon uses the binned edge index.
+   logical, allocatable :: polygon_uses_binned_edges(:) !< Whether each polygon uses the binned edge index.
    integer, allocatable :: i_poly_bin_start(:), i_poly_num_bins(:) !< Start and number of bins for each polygon.
    integer, allocatable :: edge_bin_offsets(:), edge_indices(:) !< CSR offsets and ordered local edge indices.
    real(kind=dp), allocatable :: y_poly_bin_scale(:) !< Scale from polygon y-coordinate to bin index.
@@ -136,7 +136,7 @@ contains
       call realloc(polygon_type, polygons, keepExisting=.true.)
 
       if (present(num_query_points) .and. present(x_query) .and. present(y_query)) then
-         call init_polygon_edge_index(num_query_points, x_query, y_query)
+         call init_binned_polygon_edges(num_query_points, x_query, y_query)
       end if
 
       ! check if there are any enclosure polygons
@@ -236,8 +236,8 @@ contains
       if (allocated(i_poly_end)) then
          deallocate (i_poly_end)
       end if
-      if (allocated(edge_index_enabled)) then
-         deallocate (edge_index_enabled)
+      if (allocated(polygon_uses_binned_edges)) then
+         deallocate (polygon_uses_binned_edges)
       end if
       if (allocated(i_poly_bin_start)) then
          deallocate (i_poly_bin_start)
@@ -270,20 +270,21 @@ contains
       logical :: is_inside !< Result
 
       integer :: i_bin, i_start, i_end, n_points
-      logical :: use_edge_index
+      logical :: use_binned_edges
 
       ! Get bounds for this polygon from module arrays
       i_start = i_poly_start(i_poly)
       i_end = i_poly_end(i_poly)
       n_points = i_end - i_start + 1
 
-      use_edge_index = .false.
-      if (allocated(edge_index_enabled)) then
-         use_edge_index = edge_index_enabled(i_poly)
+      use_binned_edges = .false.
+      if (allocated(polygon_uses_binned_edges)) then
+         use_binned_edges = polygon_uses_binned_edges(i_poly)
       end if
 
-      if (use_edge_index) then
-         i_bin = i_poly_bin_start(i_poly) + get_edge_bin(y, y_poly_min(i_poly), y_poly_bin_scale(i_poly), i_poly_num_bins(i_poly)) - 1
+      if (use_binned_edges) then
+         i_bin = get_edge_bin(y, y_poly_min(i_poly), y_poly_bin_scale(i_poly), i_poly_num_bins(i_poly))
+         i_bin = i_poly_bin_start(i_poly) + i_bin - 1
          is_inside = pinpok_raycast(x, y, xpl_cache(i_start:i_end), ypl_cache(i_start:i_end), n_points, &
                                    edge_indices(edge_bin_offsets(i_bin):edge_bin_offsets(i_bin + 1) - 1))
       else
@@ -292,19 +293,20 @@ contains
 
    end function pinpok_elemental
 
-   !> Build latitude-bin edge indices only where the observed query distribution predicts a worthwhile reduction.
-   subroutine init_polygon_edge_index(num_query_points, x_query, y_query)
+   !> Select polygons that benefit from latitude bins and build their shared edge index.
+   !!
+   !! A horizontal ray can intersect only edges whose vertical extent contains the query y-coordinate.
+   !! The index stores those candidate edges per latitude bin. It changes only which edges reach the
+   !! existing ray-casting calculation; it does not approximate or simplify polygon geometry.
+   subroutine init_binned_polygon_edges(num_query_points, x_query, y_query)
       integer, intent(in) :: num_query_points !< Number of points that will be classified.
       real(kind=dp), intent(in) :: x_query(:), y_query(:) !< Points that will be classified.
 
-      integer :: bin_edge_counts(max_edge_bins), bin_query_counts(max_edge_bins), bin_write_positions(max_edge_bins)
-      integer :: bins, edge, edge_end, edge_start, i_bin, i_poly, point
-      integer :: total_bins, write_position
+      integer :: bins, i_poly, num_edges, total_bins
       integer(kind=int64) :: memberships, total_memberships
-      real(kind=dp) :: candidate_work, direct_work, indexed_work
 
-      allocate (edge_index_enabled(polygons), i_poly_bin_start(polygons), i_poly_num_bins(polygons), y_poly_bin_scale(polygons))
-      edge_index_enabled = .false.
+      allocate (polygon_uses_binned_edges(polygons), i_poly_bin_start(polygons), i_poly_num_bins(polygons), y_poly_bin_scale(polygons))
+      polygon_uses_binned_edges = .false.
       i_poly_bin_start = 0
       i_poly_num_bins = 0
       y_poly_bin_scale = 0.0_dp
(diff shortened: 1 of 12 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -235,18 +263,177 @@ elemental function pinpok_elemental(x, y, i_poly) result(is_inside)
...
+
+      current_point = i_poly_start(i_poly) + edge - 1
+      previous_point = current_point - 1
+      if (edge == 1) then
+         previous_point = i_poly_end(i_poly)
+      end if
+      bin_first = get_edge_bin(min(ypl_cache(previous_point), ypl_cache(current_point)), y_poly_min(i_poly), &
+                               y_poly_bin_scale(i_poly), bins)
+      bin_last = get_edge_bin(max(ypl_cache(previous_point), ypl_cache(current_point)), y_poly_min(i_poly), &
+                              y_poly_bin_scale(i_poly), bins)
```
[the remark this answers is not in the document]
FlorisBuwaldaDeltares: added tolerance

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -235,18 +263,177 @@ elemental function pinpok_elemental(x, y, i_poly) result(is_inside)
...
+   subroutine init_binned_polygon_edges()
+      integer :: bins, i_poly, num_edges, total_bins, total_edges
+      integer(kind=int64) :: memberships, total_memberships
+
+      total_edges = 0
+      do i_poly = 1, polygons
+         total_edges = total_edges + i_poly_end(i_poly) - i_poly_start(i_poly) + 1
+      end do
+      if (total_edges < min_edges_for_binning) then
+         return
```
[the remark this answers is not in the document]
FlorisBuwaldaDeltares: simplified everything: we use binning only for dry-area enclosure files as bounding box rejection is inefficient there. No more concern about inefficient binning decision making

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
       integer, intent(in) :: polygon_points !< Number of polygon points
       real(kind=dp), intent(in) :: x_poly(polygon_points), y_poly(polygon_points), z_poly(polygon_points) !< Polygon coordinate arrays
+      logical, intent(in) :: enable_binning !< Whether to build a latitude-binned edge index for large polygon sets.
 
       integer :: i_point, i_start, i_end, i_poly
 
-      if (cellmask_initialized) then
-         call cellmask_from_polygon_set_cleanup()
-      end if
+      call polygon_cache%clear()
```
rene-deltares: Please leave initializing, clearing and deinitializing members of `polygon_cache%geometry` to the `geometry` instance itself. As a general rule, a derived type is the _owner_ of its own members, and if those members are allocatable and need to be initialized, that is done in the contructor of that derived type, or by a factory function of the derived type. `cellmask_from_polygon_set_init` has *no business* allocating, increasing the capacity, deallocating, or even checking if the `polygon_cache` members are initialized. The beauty of the derived types, if properly implemented, is that users of the derived type _don't have to worry_ if the internals of the derived type are in a consistent state. The derived type should be in a consistent state after calling the contructor (or a factory) and after method calls, the derived type should _stay_ in a consistent state. Ideally its internals are not even accessible (or if they are, at least immutable by anything other than its internal methods).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
+      procedure, private :: count_edges_per_bin => binned_edge_index_count_edges_per_bin
+      procedure, private :: edge_bin_range => binned_edge_index_edge_bin_range
+      procedure, nopass, private :: get_bin => binned_edge_index_get_bin
+   end type t_binned_edge_index
+
+   !> Complete cached polygon set, including optional acceleration data and lifecycle state.
+   type, private :: t_polygon_set_cache
+      type(t_polygon_geometry) :: geometry
+      type(t_binned_edge_index) :: edge_index
+      logical :: initialized = .false.
```
rene-deltares: In general, I frown upon these kind of `initialized` members. To me, `initialized` should be true after calling the constructor or a factory for this derived type, and every method call should leave `initialized` true as well. `initialized == .false.` just tells me the instance is in an inconsistent state, which is your job as the implementer to make sure can't happen.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
-   integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
-   logical :: cellmask_initialized = .false. !< Flag indicating if cellmask data structures have been initialized for safety
-   logical :: enclosures_present = .false. !< Flag indicating if any enclosures are present in the polygon dataset
+   !> Coordinates and per-polygon metadata shared by masking and net-cell lookup.
+   type, private :: t_polygon_geometry
+      real(kind=dp), allocatable :: x(:), y(:), z(:)
+      real(kind=dp), allocatable :: x_min(:), y_min(:), x_max(:), y_max(:)
+      real(kind=dp), allocatable :: polygon_type(:)
+      integer, allocatable :: polygon_start(:), polygon_end(:)
+      integer :: polygon_count = 0
```
rene-deltares: This type needs either a contructor or factory to initialize it's member arrays. Now users of this derived type are responsible for initializing it's members and so need to be aware of all of its' internal state. This does not make it very easy to use.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
-   real(kind=dp), allocatable :: x_poly_max(:), y_poly_max(:) !< Polygon bounding box max coordinates, (dim = polygons)
-   real(kind=dp), allocatable :: polygon_type(:) !< Polygon type, positive or dmiss = drypoint , negative = enclosure (dim = polygons)
-   integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
-   logical :: cellmask_initialized = .false. !< Flag indicating if cellmask data structures have been initialized for safety
-   logical :: enclosures_present = .false. !< Flag indicating if any enclosures are present in the polygon dataset
+   !> Coordinates and per-polygon metadata shared by masking and net-cell lookup.
+   type, private :: t_polygon_geometry
+      real(kind=dp), allocatable :: x(:), y(:), z(:)
+      real(kind=dp), allocatable :: x_min(:), y_min(:), x_max(:), y_max(:)
+      real(kind=dp), allocatable :: polygon_type(:)
```
rene-deltares: I see the z coordinates are abused to communicate polygon types. But how many types are there, two right? Only enclosure and non-enclosure types? Wouldn't it make more sense to make it logical instead and call it `is_enclosure` or something?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
-   integer :: polygons = 0 !< Number of polygons stored in module arrays xpl, ypl, zpl
-   real(kind=dp), allocatable :: x_poly_min(:), y_poly_min(:) !< Polygon bounding box min coordinates, (dim = polygons)
-   real(kind=dp), allocatable :: x_poly_max(:), y_poly_max(:) !< Polygon bounding box max coordinates, (dim = polygons)
-   real(kind=dp), allocatable :: polygon_type(:) !< Polygon type, positive or dmiss = drypoint , negative = enclosure (dim = polygons)
-   integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
-   logical :: cellmask_initialized = .false. !< Flag indicating if cellmask data structures have been initialized for safety
-   logical :: enclosures_present = .false. !< Flag indicating if any enclosures are present in the polygon dataset
+   !> Coordinates and per-polygon metadata shared by masking and net-cell lookup.
+   type, private :: t_polygon_geometry
+      real(kind=dp), allocatable :: x(:), y(:), z(:)
```
rene-deltares: Do you need `z` for anything? You've got polygon_type, right?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
+      real(kind=dp), allocatable :: x_min(:), y_min(:), x_max(:), y_max(:)
+      real(kind=dp), allocatable :: polygon_type(:)
+      integer, allocatable :: polygon_start(:), polygon_end(:)
+      integer :: polygon_count = 0
+      logical :: enclosures_present = .false.
+   end type t_polygon_geometry
+
+   !> Latitude-binned lookup table containing ordered candidate polygon edges for ray casting.
+   type, private :: t_binned_edge_index
```
rene-deltares, line 61: What do you gain by having this `t_binned_edge_index` in a separate derived type. As far as I can tell from the time I've spent looking at it: Isn't the algorithm without binning essentially equivalent to the case of having a single bin? You could just put everything in a single derived type instead.

Perhaps you could branch on 'bin_count == 1' or whatever to maybe skip some unnecessary steps in the case there's only one bin (like maybe even the allocation of the bin administration is unnecessary in that case). If I'm wrong and the algorithm is completely different, or if even the `bin_count == 1` checks are hurting the performance too much, then I would suggest writing two concrete implementations of the same algorithm (with the same interface).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
+      is_inside = pinpok_raycast(x, y, geometry%x(first_point:last_point), geometry%y(first_point:last_point), num_points, &
+                                 this%edge_indices(candidate_first:candidate_last))
+   end function binned_edge_index_point_is_inside
+
+   !> Populate the module-level polygon cache and optional acceleration data.
+   subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly, enable_binning)
       use m_alloc
       use geometry_module, only: get_startend
 
       integer, intent(in) :: polygon_points !< Number of polygon points
```
rene-deltares: I think almost this entire subroutine should be replaced by a call to a factory: `t_polygon_set_cache%from_packed_poly_coordinates` or something like that. It should just return the fully initialized instance of the `t_polygon_set_cache`.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
rene-deltares: Do you even need the `polygon_cache` global? It seems on every single call of this subroutine, you start by throwing away all of it's member fields, settings `initialized` to false and building it up from scratch all over again. Why not make your life much easier, get rid of the global `polygon_cache` and just put the thing here in the local scope so it cleans itself up after this subroutine ends? Or am I missing something?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -39,195 +40,266 @@ module m_cellmask_from_polygon_set
...
+      call polygon_cache%clear()
    end subroutine cellmask_from_polygon_set_cleanup
 
-!> Elemental wrapper for cellmask operations using module-level polygon arrays
+   !> Release all coordinates, metadata and acceleration data owned by a polygon cache.
+   subroutine polygon_set_cache_clear(this)
+      class(t_polygon_set_cache), intent(inout) :: this
+
+      associate (geometry => this%geometry)
+         if (allocated(geometry%x)) deallocate (geometry%x)
```
rene-deltares: I'm not sure if you'd like to keep the `geometry` member of the cache, but if you do please delegate this block to `geometry`'s own cleanup routine. Or better yet, leave it to a destructor (if possible the default destructor))

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: oh, you are right, polygon_type used to be put into the z-coordinate back when m_polygon was used. With a separate polygon_type array I don't think we need it any more

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: yes, i think this can be done

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: i considered both, but copilot advised me that bin_count == 1 would introduce unnecessary overhead and making both algorithms a concrete implementation of an abstract type was also not worth it. But given your comments i'll think about it some more

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: currently that would let this type bleed into geometry_module which introduces a whole set of headaches, so I was trying to keep this type refactor contained to this fortran module. But I do agree with you, maybe there's an elegant way to keep it local, i'll have a look

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: z is gone!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: indeed, we now have is_enclosure

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: made a constructor

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: yes agreed, the constructor always makes a properly initialized instance

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: I don't think making a single type is worth it, there will be significant overhead for bin_count == 1. currently if we're binning we're just evaluating against a reduced set of links. Removing that and duplicating the entirety of the raycasting setup etc is also not worth it imo. I don't see an elegant alternative for the current setup so i'd prefer to keep it as is

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: agreed!

FlorisBuwaldaDeltares: i've managed to remove the global, and even managed to keep pinpok_raycast_elemental elemental :D

hrajagers:
Hi @FlorisBuwaldaDeltares, you're using bins per latitude. I guess that you could do the same using longitude bins. I just realized that several of our river models are most east-west oriented (Nederrijn, Waal, Maas downstream of Grave). I'm wondering whether for such models it would make a big difference which binning direction is used. Maybe the overall speed gain isn't big enough to worry about this, but it would be good to test this using one of the RWS acceptance models for rivers. I don't want this to block this improvement, so maybe it can be done as a follow-up action.

FlorisBuwaldaDeltares:
hi @hrajagers, thanks for the suggestion, I hadn't thought of that. This PR is on hold until somebody has time to review it or René is back from holiday, so I have time to work on it still. 

I'll think about whether we can use runtime diagnostics to determine whether to bin horizontally or vertically. i'm hesitant to introduce an mdu keyword just to shave some time off model initialization..

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
 
-   public :: cellmask_from_polygon_set_init, cellmask_from_polygon_set_cleanup, cellmask_from_polygon_set, pinpok_elemental
-   public :: init_cell_geom_as_polylines, point_find_netcell, cleanup_cell_geom_polylines
-   public :: find_cells_crossed_by_polyline
+   public :: t_netcell_set, t_polygon_set
+
+   !> Coordinates and per-polygon metadata shared by masking and net-cell lookup.
+   type, private :: t_polygon_geometry
+      real(kind=dp), allocatable :: x(:), y(:)
+      real(kind=dp), allocatable :: x_min(:), y_min(:), x_max(:), y_max(:)
```
FlorisBuwaldaDeltares: missing apidocs, what is the size of this array?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
thomas-pijls: Nice work

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
 
    private
 
-   public :: cellmask_from_polygon_set_init, cellmask_from_polygon_set_cleanup, cellmask_from_polygon_set, pinpok_elemental
-   public :: init_cell_geom_as_polylines, point_find_netcell, cleanup_cell_geom_polylines
-   public :: find_cells_crossed_by_polyline
+   public :: t_netcell_set, t_polygon_set
+
+   !> Coordinates and per-polygon metadata shared by masking and net-cell lookup.
+   type, private :: t_polygon_geometry
```
thomas-pijls, line 43: In the other derived types you put private in the body of the derived type. Can you do that here as well to make it consistent?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
+      real(kind=dp), allocatable :: x(:), y(:)
+      real(kind=dp), allocatable :: x_min(:), y_min(:), x_max(:), y_max(:)
+      logical, allocatable :: is_enclosure(:)
+      integer, allocatable :: polygon_start(:), polygon_end(:)
```
thomas-pijls: According to the coding guidelines, we should use `dimension(:) ::` instead of `:: var_name(:)`. Could you change that?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
thomas-pijls: Additionally, it would be very nice if you could add api docs to the type attributes

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
+   end interface t_polygon_geometry
+
+   !> Latitude-binned lookup table containing ordered candidate polygon edges for ray casting.
+   type, private :: t_binned_edge_index
+      private
+      integer, allocatable :: polygon_bin_start(:) !< First global bin for each polygon.
+      integer, allocatable :: polygon_num_bins(:) !< Number of latitude bins for each polygon.
+      integer, allocatable :: bin_offsets(:) !< CSR offsets into edge_indices for each global bin.
+      integer, allocatable :: edge_indices(:) !< Ordered local edge indices for all bins.
+      real(kind=dp), allocatable :: polygon_bin_scale(:) !< Scale from polygon y-coordinate to local bin.
```
thomas-pijls: Same here, use `dimension(:)` rather than `var_name(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
 
-   integer :: polygons = 0 !< Number of polygons stored in module arrays xpl, ypl, zpl
-   real(kind=dp), allocatable :: x_poly_min(:), y_poly_min(:) !< Polygon bounding box min coordinates, (dim = polygons)
-   real(kind=dp), allocatable :: x_poly_max(:), y_poly_max(:) !< Polygon bounding box max coordinates, (dim = polygons)
-   real(kind=dp), allocatable :: polygon_type(:) !< Polygon type, positive or dmiss = drypoint , negative = enclosure (dim = polygons)
-   integer, allocatable :: i_poly_start(:), i_poly_end(:) !< Polygon start and end indices in coordinate arrays (dim = polygons)
-   logical :: cellmask_initialized = .false. !< Flag indicating if cellmask data structures have been initialized for safety
-   logical :: enclosures_present = .false. !< Flag indicating if any enclosures are present in the polygon dataset
+   !> Classify a point using only the ordered polygon edges in its latitude bin.
+   pure function binned_edge_index_point_is_inside(this, x, y, i_poly, geometry) result(is_inside)
```
thomas-pijls: Can you replace `this` by `self` in this subroutine?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
+      use geometry_module, only: pinpok_raycast
 
-contains
+      class(t_binned_edge_index), intent(in) :: this
```
thomas-pijls: You can also add an api doc to `this/self`. Maybe that helps when reading the subroutine body and you forget what `self` was again ;)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
+      candidate_first = this%bin_offsets(global_bin)
+      candidate_last = this%bin_offsets(global_bin + 1) - 1
+
+      is_inside = pinpok_raycast(x, y, geometry%x(first_point:last_point), geometry%y(first_point:last_point), num_points, &
+                                 this%edge_indices(candidate_first:candidate_last))
+   end function binned_edge_index_point_is_inside
+
+   !> Construct a consistent polygon cache from packed coordinates separated by missing values.
+   function construct_polygon_set(x_poly, y_poly, z_poly, enable_binning) result(cache)
+      real(kind=dp), intent(in) :: x_poly(:), y_poly(:), z_poly(:) !< Packed polygon coordinate arrays.
```
thomas-pijls: Use `dimension(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -28,62 +28,150 @@
...
 
-      call init_geom_cache(polygon_points, x_poly, y_poly, z_poly)
+   end function construct_polygon_set
+
+   !> Construct polygon geometry and metadata from packed coordinates separated by missing values.
+   function construct_polygon_geometry(x_poly, y_poly, z_poly) result(geometry)
+      use m_alloc
+      use geometry_module, only: get_startend
+
+      real(kind=dp), intent(in) :: x_poly(:), y_poly(:), z_poly(:) !< Packed polygon coordinate arrays.
```
thomas-pijls: Use `dimension(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+      geometry%enclosures_present = any(geometry%is_enclosure)
 
-   end subroutine cellmask_from_polygon_set_init
+   end function construct_polygon_geometry
 
    !> Check if a point should be masked, either is_inside a dry-area polygon or outside an enclosure polygon.
-   elemental function cellmask_from_polygon_set(x, y) result(mask)
-
-      integer :: mask
+   elemental function polygon_set_is_masked(this, x, y) result(is_masked)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
-   end subroutine cellmask_from_polygon_set_cleanup
+   end function polygon_set_is_masked
 
-!> Elemental wrapper for cellmask operations using module-level polygon arrays
-   elemental function pinpok_elemental(x, y, i_poly) result(is_inside)
+   !> Test whether a point lies inside one cached polygon.
+   elemental function polygon_set_polygon_contains_point(this, x, y, i_poly) result(is_inside)
       use geometry_module, only: pinpok_raycast
 
+      class(t_polygon_set), intent(in) :: this
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+      end if
 
-      integer, intent(in) :: npl_init
-      real(kind=dp), intent(in) :: xpl_init(npl_init), ypl_init(npl_init), zpl_init(npl_init)
+      call index%populate_bin_storage(geometry, total_bins, total_memberships)
+
+   end function construct_binned_edge_index
+
+   !> Allocate and populate the shared compressed-row storage for the complete polygon set.
+   subroutine binned_edge_index_populate_bin_storage(this, geometry, total_bins, total_memberships)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+      end do
 
-      xpl_cache = xpl_init
-      ypl_cache = ypl_init
-      zpl_cache = zpl_init
-      npl_cache = npl_init
+   end subroutine binned_edge_index_populate_bin_storage
 
+   !> Count memberships in O(edges), without visiting every bin that an edge spans.
+   function binned_edge_index_count_edge_memberships(this, geometry, i_poly, bins) result(memberships)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+      memberships = 0_int64
+      num_edges = geometry%polygon_end(i_poly) - geometry%polygon_start(i_poly) + 1
+      do edge = 1, num_edges
+         call this%edge_bin_range(geometry, i_poly, edge, bins, bin_first, bin_last)
+         memberships = memberships + int(bin_last - bin_first + 1, int64)
+      end do
+   end function binned_edge_index_count_edge_memberships
+
+   !> Count candidate edges in each latitude bin for one polygon.
+   subroutine binned_edge_index_count_edges_per_bin(this, geometry, i_poly, bins, bin_counts)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+      class(t_binned_edge_index), intent(in) :: this
+      type(t_polygon_geometry), intent(in) :: geometry !< Polygon geometry being indexed.
+      integer, intent(in) :: i_poly, bins !< Polygon index and number of bins.
+      integer, intent(out) :: bin_counts(:) !< Edge count per bin.
```
thomas-pijls: Use `dimensino(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -100,183 +188,281 @@ subroutine cellmask_from_polygon_set_init(polygon_points, x_poly, y_poly, z_poly
...
+            bin_counts(i_bin) = bin_counts(i_bin) + 1
+         end do
+      end do
+   end subroutine binned_edge_index_count_edges_per_bin
+
+   !> Get the inclusive bin range intersected by an edge.
+   !! Both edge endpoints and queries use get_edge_bin, so equal y-coordinates always map to the same
+   !! bin. Filling every bin from the lower endpoint through the upper endpoint preserves all edges
+   !! that the full ray cast could inspect at a query's y-coordinate.
+   pure subroutine binned_edge_index_edge_bin_range(this, geometry, i_poly, edge, bins, bin_first, bin_last)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -308,59 +494,59 @@ subroutine init_cell_geom_as_polylines()
...
-
-   !> call general polygon cleanup and restore previous polygon data
-   subroutine cleanup_cell_geom_polylines()
-      call cellmask_from_polygon_set_cleanup()
-   end subroutine cleanup_cell_geom_polylines
-
-!> Fast replacement for INCELLS using cached geometry in global polygon arrays
-   elemental function point_find_netcell(x, y) result(k)
+!> Fast replacement for INCELLS using cached net-cell geometry.
+   elemental function netcell_set_find_netcell(this, x, y) result(k)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -308,59 +494,59 @@ subroutine init_cell_geom_as_polylines()
...
+            end if
+         end do
+      end associate
 
-   end function point_find_netcell
+   end function netcell_set_find_netcell
 
 !> Find all cells crossed by polyline using brute force on cached geometry. The routine is inclusive of edge cases (touching edges or vertices).
-   subroutine find_cells_crossed_by_polyline(xpoly, ypoly, crossed_cells, error)
+   subroutine netcell_set_find_cells_crossed_by_polyline(this, xpoly, ypoly, crossed_cells, error)
```
thomas-pijls: Replace `this` by `self`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -376,29 +562,30 @@ subroutine find_cells_crossed_by_polyline(xpoly, ypoly, crossed_cells, error)
...
       end if
 
-   end subroutine find_cells_crossed_by_polyline
+   end subroutine netcell_set_find_cells_crossed_by_polyline
 
 !> Find all cells that a segment crosses and mark them in cellmask
-   subroutine find_cells_for_segment(xa, ya, xb, yb, cellmask)
+   subroutine find_cells_for_segment(cache, xa, ya, xb, yb, cellmask)
 
       implicit none
```
thomas-pijls: No `implicit none` in function/subroutine bodies!!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -308,59 +494,59 @@ subroutine init_cell_geom_as_polylines()
...
+   end function netcell_set_find_netcell
 
 !> Find all cells crossed by polyline using brute force on cached geometry. The routine is inclusive of edge cases (touching edges or vertices).
-   subroutine find_cells_crossed_by_polyline(xpoly, ypoly, crossed_cells, error)
+   subroutine netcell_set_find_cells_crossed_by_polyline(this, xpoly, ypoly, crossed_cells, error)
       use m_alloc, only: realloc
       use network_data, only: nump
       use m_missing, only: dmiss
 
       implicit none
```
thomas-pijls: No `implicit none` in function/subroutine bodies!!

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
```diff
@@ -376,29 +562,30 @@ subroutine find_cells_crossed_by_polyline(xpoly, ypoly, crossed_cells, error)
...
 
 !> Find all cells that a segment crosses and mark them in cellmask
-   subroutine find_cells_for_segment(xa, ya, xb, yb, cellmask)
+   subroutine find_cells_for_segment(cache, xa, ya, xb, yb, cellmask)
 
       implicit none
 
+      class(t_polygon_set), intent(in) :: cache
       real(kind=dp), intent(in) :: xa, ya, xb, yb !< Segment endpoints
       integer, intent(inout) :: cellmask(:) !< Cell mask array: 1=crossed, 0=not crossed
```
thomas-pijls: Use `dimension(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/pol_to_cellmask.F90
```diff
@@ -41,34 +41,34 @@ module m_pol_to_cellmask
 contains
 
    !> Create a cellmask from a set of x,y coordinates and a set of polygon points. Any  point inside the polygon is masked as 1.
-   function pol_to_cellmask(polygon_points, x_poly, y_poly, z_poly, num_netcells, x_points, y_points) result(mask)
+   function pol_to_cellmask(polygon_points, x_poly, y_poly, z_poly, num_netcells, x_points, y_points, enable_binning) result(mask)
       use m_alloc, only: realloc
 
       integer, intent(in) :: polygon_points !< Number of polygon points
       integer, intent(in) :: num_netcells !< Number of points to mask
       real(kind=dp), intent(in) :: x_poly(polygon_points), y_poly(polygon_points), z_poly(polygon_points) !< Polygon coordinate arrays
```
thomas-pijls: Use `dimension(polygon_points)`

While debugging I also find that Visual Studio struggles with finding the elements with this kind of array-length specification. VS says it can't find the addresses. Therefore, consider making them `allocatable` and allocate them with size `polygon_points`?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
```diff
@@ -2752,14 +2752,15 @@ subroutine finalize()
...
       use m_array_or_scalar, only: realloc
-      use m_cellmask_from_polygon_set, only: init_cell_geom_as_polylines, point_find_netcell, cleanup_cell_geom_polylines
+      use m_cellmask_from_polygon_set, only: t_netcell_set
       use unstruc_inifields, only: finalize_1dfield_global_values
       use network_data, only: LINK_1D
 
       integer :: j, k, ierr, l, n, itp, kk, k1, k2, nstor, i, ja
       logical :: hyst_dummy(2)
       real(kind=dp) :: area, width, hdx
       type(t_storage), pointer :: stors(:)
```
thomas-pijls: Use `dimension(:)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_pol_to_cellmask.f90
```diff
@@ -445,16 +451,130 @@ subroutine test_pinpok_raycast_complex() bind(C)
...
+      integer, parameter :: grid_size = 41
+      integer, parameter :: query_points = grid_size * grid_size + 8
+      real(kind=dp), parameter :: pi = acos(-1.0_dp)
+
+      integer :: column, original_jins, point, row
+      integer :: actual_mask, expected_mask, polygons_containing_point
+      logical :: inside_large_polygon, inside_small_polygon
+      real(kind=dp) :: angle
+      real(kind=dp) :: x_poly(polygon_points), y_poly(polygon_points), z_poly(polygon_points)
+      real(kind=dp) :: x_query(query_points), y_query(query_points)
```
thomas-pijls: Use `dimension(*_points)`

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_pol_to_cellmask.f90
```diff
@@ -445,16 +451,130 @@ subroutine test_pinpok_raycast_complex() bind(C)
...
+      integer, parameter :: polygon_segments = 6
+      integer, parameter :: points_per_segment = 48
+      integer, parameter :: polygon_points = polygon_segments * points_per_segment + 1
+      integer, parameter :: query_points = 14
+
+      integer :: actual_mask, original_jins, point, segment, step
+      real(kind=dp) :: fraction
+      real(kind=dp) :: vertex_x(polygon_segments + 1), vertex_y(polygon_segments + 1)
+      real(kind=dp) :: x_poly(polygon_points), y_poly(polygon_points), z_poly(polygon_points)
+      real(kind=dp) :: x_query(query_points), y_query(query_points)
```
thomas-pijls: Use `dimension(*_points)`

thomas-pijls (changes requested):
Great work, especially putting al the polygon stuff into derived types and making the procudures type-bound. That was long overdue!

Requested a lot of style/coding guideline changes

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: will add apidocs. There is a difference between a private type and declaring private in its body. The body declaration means that all variables in the type body are private, but not the type itself. If you scroll up there's also a private typee with private members. It's the same as putting private in your module declaration, it means that from this point onward, everything declared is private.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: will do, I also updated the fortran.instructions.md to reflect this

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: will do

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
thomas-pijls: Thanks for the explanation, I didn't know that subtlety. Learned something today :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
thomas-pijls: I agree with Floris' 2 type approach

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/pol_to_cellmask.F90
FlorisBuwaldaDeltares: these should not be allocatable, they are passed by reference

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings.f90
FlorisBuwaldaDeltares: this line is not touched by my PR so i don't know why you want me to change it, but since i'm doing another commit anyway i can change it

thomas-pijls (approved):
Great work and thanks for processing my style comments :)

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: done, t_polygon_set_cache (renamed t_polygon_geometry) now has a default constructor that returns a fully initialized instance

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/prepost/m_cellmask_from_polygon_set.f90
FlorisBuwaldaDeltares: yes, we now rely everywhere on the default destructor deallocating local objects that go out of scope

PR #1226, opened 2026-08-25
UNST-10068; map atmosphericpressure ext-alias to airPressure for bc support

arthurvd:
...because even with recent cleanup in ec_addtimespacerelation, the call to `ecCreateInitializeBCFileReader()` uses the given (ext) quantity name as only name that will be searched in the .bc header. atmosphericpressure was the only remaining alias quantity, all others are unique, so quickest solution was to remap atmosphericpressure to airpressure in the existing function `quantity_name_config_file_to_internal_name()`.

# What was done 

- e.g. … 
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
@@ -4341,7 +4341,7 @@ contains
       case ('normalvelocitybnd')
          itemPtr1 => item_normalvelocitybnd
          dataPtr1 => zbndn
-      case ('airpressure', 'atmosphericpressure')
+      case ('airpressure')
          itemPtr1 => item_atmosphericpressure
          dataPtr1 => air_pressure
       case ('pseudoairpressure')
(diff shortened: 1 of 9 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/meteo1.f90
```diff
@@ -4194,6 +4194,8 @@ pure function quantity_name_config_file_to_internal_name(quantity_input_name) re
          quantity_internal_name = 'sea_ice_thickness'
       case ('bedrocksurfaceelevation')
          quantity_internal_name = 'bedrock_surface_elevation'
+      case ('atmosphericpressure')
```
FlorisBuwaldaDeltares, line 4197: I think we can now remove 'atmosphericpressure' from addtimespacerelation:

Image

FlorisBuwaldaDeltares: and also here:

Image

FlorisBuwaldaDeltares (commented):
one small cleanup suggestion, otherwise a nice and easy fix!

Question, we do not need to do anything for the combo quantities (like airpressure_windx_windy) because they are already airpressure-only right?

FlorisBuwaldaDeltares (approved):
nice bit of extra cleanup :)

JulienGroenenboom (approved):

PR #1231, opened 2026-08-26
Fm/task/unst 8868 implement v3 structures for 3d

jjnoort:
# What was done 

Structures for 3d in Flow1d

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated \
All testcases in F208
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-8868


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/distribute_linearized_3d_structure_coefficients.f90
rene-deltares, line 94: In the debugger I get NaN values in `au2` here. The line evaluates to `au2 = 0 * NaN`, and the result is NaN in linux. I just have to continue a few times before it appears. Apparently in Windows something else happens. But I suspect this is why the waterlevel variables contain NaNs in the his file.

The ff3 value in this case is uninitialized I think? It already contained `NaN` before the loop at line 77, and the `au(2, 7)` is zero, so it is not set in the loop either. Maybe the `au1-3` assigments need `if` guards too. Or in the loop in line 77, if `au(i, L0) /= 0`, you set ff3 to zero as well? Not sure

rene-deltares, line 78: Possible NaN fix? (see other comment further on):
`ff3(:, LL - Lb + 1) = 0.0_dp`

test/deltares_testbench/data/cases/e02_dflowfm/f208_3D_hydraulic_structures/c011_gate_sub_3Dz/reference_win64.dvc
```diff
@@ -1,6 +1,6 @@
 outs:
-- md5: 629990068ad530cc673a6df5d01cf077.dir
-  size: 16184421
+- md5: ffc33f0b0f516a4a4d77967fd2184f3f.dir
```
rene-deltares, line 2: this case (c011) has platform specific references (along with c015, c016, c017, c023) so I would expect reference_lnx64.dvc to have changes as well. If the code is OK I can upload them if you want.

robin-deltares (approved):

PR #1235, opened 2026-08-27
UNST-10275 determine long culvert orientation based on flownodes, not netnodes.

FlorisBuwaldaDeltares:
To support partitioned long culverts, loop through all long culvert coordinate pairs instead of only the first one.

# What was done 

- change set_longculvert_flow_direction to be flownode/link based instead of netnode/link based as these orientations may differ
- now loop through all coordinates to support partitioned long culverts
- add unit test verifying behavior
- exclude long culvert links from other structures
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link


src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_longculverts.f90
```diff
@@ -702,10 +703,175 @@ subroutine test_allowed_flowdir_follows_coordinate_order() bind(C)
...
+      call f90_expect_eq(longculverts(1)%orientation, 1, &
+             cstr("orientation reconstruction must find the second segment as the first local flow link"))
+
       call default_longculverts()
    end subroutine test_allowed_flowdir_follows_coordinate_order
    !$f90tw)
 
+   !> Creates two culverts with opposite input-coordinate orders and an exact duplicate of the first.
+   subroutine create_allowed_flowdir_structure_file(filename, allowed_flowdir)
+      character(len=*), intent(in) :: filename
```
rene-deltares: Ugh, this needs to be replaced by the new file helper. It is bad enough that we have to create files for unit tests, but at least reuse the helpers.

FlorisBuwaldaDeltares: oops, copilot is very fond of reinventing the wheel, didn't catch this one, sorry

FlorisBuwaldaDeltares: we now have a single shared long culvert structurefile helper, that reuses the create_file helper

.github/instructions/fortran.instructions.md
```diff
@@ -41,6 +41,10 @@ pain.
   local variables, derived-type components, allocatables, pointers, and fixed-size arrays.
   For example, use `real(kind=dp), dimension(:), intent(in) :: values` instead of
   `real(kind=dp), intent(in) :: values(:)`.
+- Prefer lines up to 160 characters. Do not add continuation line breaks below this limit
```
rene-deltares: Not sure if I agree with this. Is this in our coding guidelines? I think 160 characters is rather high and when calling subroutines with lots of named arguments, and I sometimes prefer them on separate lines. If everyone else disagrees with me then fine, but this PR is not the place to add this without discussing with the rest of the team.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
```diff
@@ -865,47 +910,64 @@ end subroutine longculvertsToProfs
...
-      if (netnode_1 <= 0 .or. netnode_2 <= 0) then
+      flownode_1 = ln(1, L_flow)
+      flownode_2 = ln(2, L_flow)
+      if (flownode_1 <= 0 .or. flownode_2 <= 0) then
          return
       end if
 
-      netnode_1_is_LC_node_2 = equal(xk(netnode_1), longculverts(ilongc)%xcoords(2)) .and. equal(yk(netnode_1), longculverts(ilongc)%ycoords(2))
-      netnode_2_is_LC_node_2 = equal(xk(netnode_2), longculverts(ilongc)%xcoords(2)) .and. equal(yk(netnode_2), longculverts(ilongc)%ycoords(2))
+      do coordinate_index = 1, min(size(longculverts(ilongc)%xcoords), size(longculverts(ilongc)%ycoords)) - 1
```
rene-deltares: Do I gather from this that the x coords and y coords are not always the same size? Or is this defensive programming?
In case of the latter: I think it's kind of a bummer that in all of these subroutines we can't make any assumptions about the state of the module variables, and so the learned behavior is to do lots of checks if the data is even in a consistent state to begin with. And if those checks fail, we just give up, sometimes without warning even. Even if we maintain global state we should do a better job making sure that only subroutines in this module modify the longculvert data, and each subroutine leaves the administration in a consistent state, so that in other subroutines we can just make assumptions about the validity of the state.

FlorisBuwaldaDeltares: defensive programming, but in this case you are right: this is overly defensive as we are sure that x-coords and y-coords are the same size (we check during initialization.

After our OOP polygon cache rework i'm kind of liking constructors that always return a valid object, so I agree with you completely here

.github/instructions/fortran.instructions.md
FlorisBuwaldaDeltares: ok I just got annoyed that copilot keeps line-breaking at 82 characters but you are right maybe this is not the time to add this

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/compute/m_longculverts.f90
FlorisBuwaldaDeltares: but let's fix that later

rene-deltares (approved):
Thanks, looks good. From issue description the 'remove_longculvert_links' seemed like a bonus. Does it fix another bug?

FlorisBuwaldaDeltares:
> Thanks, looks good. From issue description the 'remove_longculvert_links' seemed like a bonus. Does it fix another bug?

Yes, the 'remove_longculvert_links' was actually the breakthrough! I was investigating why the long culverts behaved so oddly in the RMM model. 4 culverts were not flowing, and one was. They were supposed to be unidirectional and for that reason I kept thinking the links must be reversed.. but what actually happened was the Haringvlietsluis marked the long culvert links as "theirs", and thus, if the haringvlietsluis was closed.. it closed the long culverts. 

I will update the issue

PR #1237, opened 2026-08-28
UNST-10288 Move GTest to Conan and update it to 1.18. Fix bug in F90T…

harmenwierenga:
…W which misinterpreted Fortran C_BOOL as C++ bool, use proper byte instead

I kept googletest as a shared library, because it is built using icpx and causes linker failures if linked statically into GCC-built libraries such as test_pre_c_sumo.

I restored the old post-build test discovery for pre-c-sumo unit tests.

# Issue link UNST-10288


AabhasSrivastava (approved):
LGTM. Is nothing needed for signing on Windows dlls with respect to gtest?

harmenwierenga:
> LGTM. Is nothing needed for signing on Windows dlls with respect to gtest?

We should not install gtest, so the signing list is not affected

KoenBussemakerDeltares (approved):
Nice, another one in conan! Let's hope the TC gods are with you so you can merge this :)

PR #1242, opened 2026-09-02
Implement communicating constituents in fm adapter

erwin-mulder:
# What was done 

Implemented communications of fixed number (10) of constituents C01, C02, .. C10.
In the FM adapter. Also updated the precice-config.xml in example 12.

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [x] Tests updated : example 12
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
UNST-10260


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -143,6 +153,11 @@ function precice_adapter_constructor(config_file, name, is_communicator_set, com
       adapter_instance%cell_center_mesh_coordinates_2d = cell_center_mesh_coordinates_2d
       adapter_instance%cell_center_mesh_coordinates_3d = cell_center_mesh_coordinates_3d
 
+      ! Set up constituents (fixed to NUM_CONSTITUENTS and named C01..C10 for now)
+      do constituent_index = 1,NUM_CONSTITUENTS
+         write(constituent_name, '(A, I2.2)') 'C', constituent_index
+         adapter_instance%quantities%constituents(constituent_index) = quantity_t(standard_name=constituent_name, is_active=.true.)
```
AabhasSrivastava, line 159: What happens if the model does not have 10 constituents? How is that error/stop handled?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
```diff
@@ -528,6 +571,9 @@ subroutine precice_adapter_add_to_fm_administration(self)
             ! TODO: Check whether the area needs to be set at all. It might only be needed if momentum needs to be passed through from the source location to the sink location.
             source_sinks%area(source_sinks%num_total) = ABS(self%sources_sinks_discharge(i)) / self%sources_momentum_magnitude_weighted(i)
          end if
+         do constituent_index = 1, NUMCONST
```
AabhasSrivastava: Same question here, what if the number of constituents in FM differ from what is sent via preC-SUMO?

AabhasSrivastava (commented):
Are we hard fixing that 10 constituents need to be defined? If there are less than 10 then should we do some value (like -999 or 0). We did agree that if the model has more than 10 constituents, we should stop with an error message.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: This just generates the different preCICE related definitions, it has no link to the constituents being actually present in FM.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_manager/precice_adapter/adapter.F90
erwin-mulder: NUMCONST is the number of constituents that is actually present in FM. Any additional entries coming from preC-SUMO are discarded and should be placeholder values anyway. As mentioned below, NUMCONST should be less or equal to 10. (Even though that is not currently enforced.) If we would want to reinforce this loop we could make it loop to min(NUM_CONSTITUENTS, NUMCONST) instead.

hrajagers: If it's not clear to the current developer that numconst refers to FM constituents (that's a variable, isn't it, so it should be lower case) and NUM_CONSTITUENTS (=10) to the constituents in the preCICE coupler, wouldn't it be a good idea to use a better identifiable name like NUM_CONSTITUENTS_IN_COUPLER. I personally think that that name is slightly on the long side, but still acceptable if used sparingly.

erwin-mulder: I renamed it to NUM_COUPLED_CONSTITUENTS. It is slightly shorter and only used inside adapter.f90, so I deem it sparingly and limited enough. I also used numconst (in lower case) in the adapter, even when the definition is in caps. I do not want to go fix all the 161 instances elsewhere, which includes the actual declaration of the thing,

AabhasSrivastava (approved):
LGTM

PR #1244, opened 2026-09-02
UNST-9674 Synced kernel keyword deprecation state with MDU spec

KoenBussemakerDeltares:
# What was done 

See also mdu spec counterpart https://github.com/Deltares/Delft3DFM.File.Specifications/pull/98

These keywords were marked as obsolete by the mdu spec but the kernel did not reflect this. Ideally we would mark them as obsolete but this created unnecessary risk right before the upcoming release. After discussion with Julien we decided to make them deprecated instead so the user gets a warning instead of an error.

# Issue link

https://issuetracker.deltares.nl/browse/UNST-9674

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_deprecated_keywords.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_deprecated_keywords.f90
@@ -30,6 +30,7 @@ contains
       call add_deprecated_keyword(deprecated_mdu_keywords, 'Geometry', 'circumcenterTolerance', DEPRECATED, 'Once the keyword is removed/becomes obsolete, a fixed tolerance will be used.')
       call add_deprecated_keyword(deprecated_mdu_keywords, 'Geometry', 'IniFieldFile', DEPRECATED, &
                                   'Use [external forcing] extForceFileNew instead. Note: you should update some keywords and values in the file contents to the new format. See the User Manual for details.')
+      call add_deprecated_keyword(deprecated_mdu_keywords, 'Geometry', 'Grdang', DEPRECATED, 'It is no longer supported, the provided value is ignored.')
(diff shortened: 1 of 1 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_deprecated_keywords.f90
```diff
@@ -38,9 +39,13 @@ subroutine default_fm_deprecated_keywords()
       call add_deprecated_keyword(deprecated_mdu_keywords, 'Numerics', 'Vertadvtypsal', DEPRECATED, 'Use verticalAdvectionType instead.')
       call add_deprecated_keyword(deprecated_mdu_keywords, 'Numerics', 'Vertadvtyptem', DEPRECATED, 'Use verticalAdvectionType instead.')
       call add_deprecated_keyword(deprecated_mdu_keywords, 'Numerics', 'maxItPresDens', DEPRECATED)
+      call add_deprecated_keyword(deprecated_mdu_keywords, 'Numerics', 'filter', DEPRECATED, 'It is no longer supported, the provided value is ignored.')
+      call add_deprecated_keyword(deprecated_mdu_keywords, 'Numerics', 'relax', DEPRECATED, 'It is no longer supported, the provided value is ignored. Use [numerics] lateral_fixedweir_relax instead.')
```
thomas-pijls: Relax can stay obsolete, it was part of the big batch of obsolete keywords and is removed from the testbench

thomas-pijls (changes requested):
Relax can stay obsolete

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/fm_deprecated_keywords.f90
KoenBussemakerDeltares: True, but there might be mdu's outside of the testbench that suddenly cause problems (the verschilanalyse comes to mind). Better to be safe than sorry right before the release.

thomas-pijls (approved):
Great work

PR #1250, opened 2026-09-03
UNST-10296: Keep com file on NcFormat 3 (classic)

mihai-pricope:
# What was done 

NetCDF mode 4 (HDF5) does not support having a file open for read and write in the same memory context in the same time. (Due to some internal NetCDF C library caching or so). Dwaves com file stays open in FM by EC Module, but it's also written for Dwaves communication. 

Dwaves seems to be able to write it (it's in a different netcdf memory context) but FM can't. 

So this will default the NetCDF mode to classic for this file (mode 3) regardless of overall FM settings. This way we can write it an read it in the same time. 

# Evidence of the work done 

- [ ]	Video/figures \
- [X]	Not applicable 

# Tests 
- [ ] Tests updated \
- [X]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [X]	Not applicable 

# Issue link
https://issuetracker.deltares.nl/browse/UNST-10296


adrimourits (approved):
Sometimes, it is what it is.

PR #1253, opened 2026-09-04
Datavalue is now time dependent if any other block of the same quanti…

FlorisBuwaldaDeltares:
…ty is time-dependent, otherwise time-independent.

# What was done 

- scan extfiles and register whether quantities are time dependent
- datavalue checks registry and takes registry value for time dependence
- added static field datavalue path, that adds timespacerelation and updates it once
 

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
@@ -1505,6 +1505,10 @@ contains
          end if
       end if
 
+      if (present(tgt_item1) .and. associated(targetItemPtr1)) then
+         tgt_item1 = targetItemPtr1
+      end if
+
       ec_addtimespacerelation = .true.
       return
 
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
@@ -51,6 +51,7 @@ contains
       use m_meteo, only: item_waqfun, item_waqsfun
       use m_ec_parameters, only: ec_undef_int
       use m_source_sink, only: source_sinks
+      use m_spatial_field, only: deallocate_time_dependent_spatial_quantities
       use m_unstruc_model_data, only: extfile_new_list
       use messageHandling, only: warn_flush, err_flush, msgbuf, LEVEL_FATAL
       use processes_input, only: num_time_functions, num_spatial_time_fuctions, nosfunext
@@ -124,6 +125,8 @@ contains
          call split_filename(file_names(i_ext), base_dirs(i_ext), fnam)
 
       end do
+
+      call scan_time_dependent_spatial_inputs(bnd_ptrs)
 
       ! Register all global WAQ functions before creating any EC target fields.
       ! Growing funinp after a field points into it would invalidate that pointer.
@@ -227,7 +230,7 @@ contains
             istat, &
             prefix='While reading '''//trim(file_names(i_ext))//'''', &
             print_context_keywords=['quantity', 'dataFile'] &
-         )
+            )
 
          if (allocated(itpenzr)) then
             deallocate (itpenzr)
@@ -248,6 +251,8 @@ contains
          call tree_destroy(bnd_ptrs(i_ext)%node_ptr)
       end do
 
+      call deallocate_time_dependent_spatial_quantities()
+
       if (res) then
          iresult = DFM_NOERR
       else
@@ -255,6 +260,69 @@ contains
       end if
 
    end subroutine init_new
+
+   !> Collect quantities with an input that is updated during the time loop.
+   subroutine scan_time_dependent_spatial_inputs(bnd_ptrs)
+      use m_meteo, only: quantity_name_config_file_to_internal_name
+      use m_spatial_field, only: t_spatial_field_input, read_spatial_field_block, is_static_file_type, &
+                                 allocate_time_dependent_spatial_quantities, register_time_dependent_spatial_quantity
+      use precision_basics, only: comparereal
+      use string_module, only: str_tolower
+      use timespace, only: convert_method_string_to_integer, get_default_method_for_file_type, &
+                           update_method_with_weightfactor_fallback, update_method_in_case_extrapolation
+      use tree_data_types, only: tree_data, tree_data_ptr
+      use tree_structures, only: tree_get_name, tree_num_nodes
+
+      type(tree_data_ptr), dimension(:), intent(in) :: bnd_ptrs
+
+      type(tree_data), pointer :: bnd_ptr
+      type(tree_data), pointer :: block_ptr
+      type(t_spatial_field_input) :: input
+      integer :: i, i_ext, max_num_quantities, num_items
+      character(len=:), allocatable :: group_name
+
+      max_num_quantities = 0
+      do i_ext = 1, size(bnd_ptrs)
+         max_num_quantities = max_num_quantities + tree_num_nodes(bnd_ptrs(i_ext)%node_ptr)
+      end do
+      call allocate_time_dependent_spatial_quantities(max_num_quantities)
+
+      do i_ext = 1, size(bnd_ptrs)
+         bnd_ptr => bnd_ptrs(i_ext)%node_ptr
+         num_items = tree_num_nodes(bnd_ptr)
+         do i = 1, num_items
+            block_ptr => bnd_ptr%child_nodes(i)%node_ptr
+            group_name = str_tolower(trim(tree_get_name(block_ptr)))
+            select case (group_name)
+            case ('spatial', 'meteo', 'parameter', 'initial')
+               input = read_spatial_field_block(block_ptr)
+            case default
+               cycle
+            end select
+
+            if (comparereal(input%data_value, dmiss) /= 0) then
+               cycle
+            end if
+            input%quantity = quantity_name_config_file_to_internal_name(input%quantity)
+            if (len_trim(input%quantity) == 0 .or. len_trim(input%forcing_file_type) == 0) then
+               cycle
+            end if
+            if (len_trim(input%interpolation_method) > 0) then
+               input%method = convert_method_string_to_integer(input%interpolation_method)
+               call update_method_with_weightfactor_fallback(input%forcing_file_type, input%method)
+            else
+               input%method = get_default_method_for_file_type(input%forcing_file_type)
+            end if
+            if (input%method == -1) cycle
+            call update_method_in_case_extrapolation(input%method, input%is_extrapolation_allowed)
+
+            if (.not. is_static_file_type(input%forcing_file_type, input%method)) then
+               call register_time_dependent_spatial_quantity(input%quantity)
+            end if
+         end do
+      end do
+
+   end subroutine scan_time_dependent_spatial_inputs
 
    !> Checks the version number of the external forcing file and opens it, returning a pointer to the tree of external forcings file boundary blocks.
    subroutine check_version_number_and_open_external_forcing_file(external_force_file_name, bnd_ptr, major, iresult)
@@ -967,9 +1035,9 @@ contains
       use tree_data_types, only: tree_data
       use fm_location_types, only: parse_spatial_location_type, UNC_LOC_S, UNC_LOC_U, UNC_LOC_3DV, UNC_LOC_S3D, SPATIAL_LOCATION_1D, SPATIAL_LOCATION_2D, SPATIAL_LOCATION_ALL
       use m_meteo, only: ec_addtimespacerelation, ec_gettimespacevalue_by_itemID, ecInstancePtr
-      use m_flowtimes, only: tzone, tunit
+      use m_flowtimes, only: irefdate, tzone, tunit, tstart_user
       use m_ec_parameters, only: ec_undef_int
-      use timespace_parameters, only: WEIGHTFACTORS, FIELD1D
+      use timespace_parameters, only: WEIGHTFACTORS, FIELD1D, DATAVALUE
       use properties, only: prop_get
       use m_alloc, only: realloc, reallocP
       use m_spatial_field, only: t_spatial_field_input, read_spatial_field_block, validate_spatial_field_input, &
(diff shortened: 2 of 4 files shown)


KoenBussemakerDeltares:
For the record, I strongly disagree with this solution. It feels like we are fixing symptoms and not causes. We somehow lose information during the ext conversion, which we then try to fix with some associative trickery after the fact. This is error prone, difficult to explain (especially to users) and likely contains many edge cases.

Imagine having two entries in the ext file that have the same quantity type, but different masks. These apply to different sections of the grid and are essentially independent. This code would add spooky-action-at-a-distance: changing one entry could magically influence the other. This could easily turn into a source of really annoying bugs, both for our users and for us to fix.

This is also a piece of code that will have to exist "forever", as removing it would potentially break existing models for our users. You could easily end up in a situation where fixing a problem with it would fix things for one user, but break it for the others. If we instead implement it as part of the extforce conversion tool then it can be something that runs just one time. That's no pain-free guarantee either, but it's more manageable.

I suggest we take the time to come up with a more robust solution. Lately we have been working on making certain quantities optionally time-variant, which is something we can (and should) take into account as well. I am much more in favor of something explicit (e.g. a dedicated setting for each quanity) rather than doing questionable after-the-fact fixes.

thomas-pijls:
I agree with Koen and I also disagree with this solution along the same reasons.

I appreciate the work you put in and your drive to quickly come with a solution to this problem, however I also think we should have a discussion about a more logical and easier to maintain long-term solution.

For now we (regretably) have to use the deprecated `dataFileType = polygon` fix as a short-term solution.

Let's shelf this for now and discuss this when Arthur and I are back from vacation in 2 weeks!

FlorisBuwaldaDeltares:
> For the record, I strongly disagree with this solution. It feels like we are fixing symptoms and not causes. We somehow lose information during the ext conversion, which we then try to fix with some associative trickery after the fact. This is error prone, difficult to explain (especially to users) and likely contains many edge cases.

We don't lose information, datavalue was always time-independent in the old ext. This problem comes from wishing to phase-out datafiletype = polygon and merging both time-independent and time-dependent into datavalue. 

> 
> Imagine having two entries in the ext file that have the same quantity type, but different masks. These apply to different sections of the grid and are essentially independent. This code would add spooky-action-at-a-distance: changing one entry could magically influence the other. This could easily turn into a source of really annoying bugs, both for our users and for us to fix.

Yes, spooky-action-at-a-distance but only for datavalue. Datavalue would be the only quantity influenced by the others, as a spooky feature not a spooky bug.

> 
> This is also a piece of code that will have to exist "forever", as removing it would potentially break existing models for our users. You could easily end up in a situation where fixing a problem with it would fix things for one user, but break it for the others. If we instead implement it as part of the extforce conversion tool then it can be something that runs just one time. That's no pain-free guarantee either, but it's more manageable.

The only two things that need to become datavalue are the old "windspeedfactor" and "solarradiationfactor", which are thankfully always time-dependent. Everything else can indeed stay datafiletype=polygon so I don't think this is very painful for the conversion tool.
> 
> I suggest we take the time to come up with a more robust solution. Lately we have been working on making certain quantities optionally time-variant, which is something we can (and should) take into account as well. I am much more in favor of something explicit (e.g. a dedicated setting for each quanity) rather than doing questionable after-the-fact fixes.

I've been thinking on this problem all day and the options we have are

1. separate time dependent & independent inputs (polygon & datavalue)
2. user-supplied field in datavalue
3. this solution (datavalue copies other blocks time-dependence)

that's about it. I understand your concerns, I just implemented this so we have it on the shelf if we need it.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -256,6 +261,62 @@ module subroutine init_new(iresult)
...
+      use m_meteo, only: quantity_name_config_file_to_internal_name
+      use m_spatial_field, only: t_spatial_field_input, read_spatial_field_block, select_spatial_field_method, &
+                                 is_static_file_type, allocate_time_dependent_spatial_quantities, &
+                                 register_time_dependent_spatial_quantity
+      use precision_basics, only: comparereal
+      use string_module, only: str_tolower
+      use tree_data_types, only: tree_data, tree_data_ptr
+      use tree_structures, only: tree_get_name, tree_num_nodes
+
+      type(tree_data_ptr), dimension(:), intent(in) :: bnd_ptrs
```
arthurvd:
```suggestion
      type(tree_data_ptr), dimension(:), intent(in) :: bnd_ptrs !< List of already loaded external forcings files.
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -256,6 +261,62 @@ module subroutine init_new(iresult)
...
+
+            if (comparereal(input%data_value, dmiss) /= 0) then
+               cycle
+            end if
+            input%quantity = quantity_name_config_file_to_internal_name(input%quantity)
+            if (len_trim(input%quantity) == 0 .or. len_trim(input%forcing_file_type) == 0) then
+               cycle
+            end if
+            input%method = select_spatial_field_method(input%forcing_file_type, input%interpolation_method, input%is_extrapolation_allowed)
+            if (input%method == -1) cycle
```
arthurvd:
```suggestion
            if (input%method == METHOD_UNKNOWN) then
               cycle
            end if
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
```diff
@@ -187,13 +199,32 @@ pure function is_static_file_type(forcing_file_type, method) result(is_static)
...
+      logical, intent(in) :: is_extrapolation_allowed !< Whether to select the extrapolating variant of the method.
+      integer :: method
+
+      if (len_trim(interpolation_method) > 0) then
+         method = convert_method_string_to_integer(interpolation_method)
+         call update_method_with_weightfactor_fallback(forcing_file_type, method)
+      else
+         method = get_default_method_for_file_type(forcing_file_type)
+      end if
+      if (method /= -1) then
```
arthurvd:
```suggestion
      if (method /= METHOD_UNKNOWN) then
```

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -880,20 +942,29 @@ function read_3d_sigma_field(quantity, target_x, target_y, mask, kx, forcing_fil
...
                                     filetype, method, oper, z=zcs, pkbot=pkbot, pktop=pktop, &
                                     varname=variable_name, tgt_item1=ec_item)
-      res = res .and. ec_gettimespacevalue_by_itemID(ecInstancePtr, ec_item, irefdate, tzone, &
-                                                     tunit, tstart_user, target_data)
+      if (is_static_field) then
+         res = res .and. ec_gettimespacevalue_by_itemID(ecInstancePtr, ec_item, irefdate, tzone, &
+                                tunit, tstart_user, target_data)
+      else
+         ec_item = ec_undef_int
+      end if
```
arthurvd: Why was this needed?

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/ec_addtimespacerelation.f90
```diff
@@ -1505,6 +1505,10 @@ module logical function ec_addtimespacerelation(name, x, y, mask, vectormax, fil
          end if
       end if
 
+      if (present(tgt_item1) .and. associated(targetItemPtr1)) then
+         tgt_item1 = targetItemPtr1
+      end if
```
arthurvd: How come that this did not cause any trouble before? (I think I know, but just want to double check with you).

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
```diff
@@ -409,4 +433,44 @@ function file_extension_conflicts_with_type(forcing_file, file_type, valid_exten
...
+      time_dependent_spatial_quantities = ''
+   end subroutine allocate_time_dependent_spatial_quantities
+
+   !> Helper routine to avoid cyclic dependencies
+   subroutine deallocate_time_dependent_spatial_quantities()
+      if (allocated(time_dependent_spatial_quantities)) then
+         deallocate (time_dependent_spatial_quantities)
+      end if
+      num_time_dependent_spatial_quantities = 0
+   end subroutine deallocate_time_dependent_spatial_quantities
```
arthurvd: It's a bit confusing that the allocate relies on the call site to set the module variable `num_time_dependent_spatial_quantities`, whereas the deallocate *does* reset `num_time_dependent_spatial_quantities` to zero.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -1337,17 +1335,83 @@ subroutine test_frictioncoefficient_static_field_populated_at_init() bind(C)
...
+      threshold_abort = LEVEL_FATAL
+      call initialize_ec_module()
+      ierr = m_polygon_destructor()
+      extfile_new_list = [EXT_FILE]
+
+      iresult = DFM_NOERR
+      call init_new(iresult)
+
+      call f90_expect_eq(iresult, DFM_NOERR, 'combined static friction fields should initialize')
+      call f90_expect_eq(ja_friction_coefficient_time_dependent, 0, &
```
arthurvd: Why does this properly test that datavalue with static input is applied only once?
I feel that this test is now constructed in such a way (with only `init_new()`, and no repeated `gettimespacevalue()`) that we don't even challenge EC module to do repeated updates.

arthurvd (commented):
Here my comments. Overall it fixes the bug I believe, but minor improvements may be nice. Let's call and discuss it this afternoon.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -880,20 +942,29 @@ function read_3d_sigma_field(quantity, target_x, target_y, mask, kx, forcing_fil
...
       integer, intent(inout) :: ec_item
       real(dp), pointer, intent(out) :: target_data(:)
+      logical, intent(in) :: is_static_field
       logical :: res
 
       integer, pointer :: pkbot(:), pktop(:)
 
-      call reallocP(target_data, ndkx, fill=dmiss, keepExisting=.false.)
+      if (is_static_field) then
+         call reallocP(target_data, ndkx, fill=dmiss, keepExisting=.false.)
```
FlorisBuwaldaDeltares: for a static field, we need a temporary buffer to hold the result until we expand it to the actual 3D array later

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -880,20 +942,29 @@ function read_3d_sigma_field(quantity, target_x, target_y, mask, kx, forcing_fil
...
+      else
+         target_data => null()
```
FlorisBuwaldaDeltares: The target_data must be null() for a time-dependent 3D sigma field! if it is not nulled, addtimespacerelation registers it as a custom array target instead of the normal one (nudgesalinitytemperature). Subsequent updates will update this temporary target_data array instead of the actual module array

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
```diff
@@ -880,20 +942,29 @@ function read_3d_sigma_field(quantity, target_x, target_y, mask, kx, forcing_fil
...
       pktop => ktop
 
       res = ec_addtimespacerelation(quantity, target_x, target_y, mask, kx, forcing_file, &
                                     filetype, method, oper, z=zcs, pkbot=pkbot, pktop=pktop, &
                                     varname=variable_name, tgt_item1=ec_item)
-      res = res .and. ec_gettimespacevalue_by_itemID(ecInstancePtr, ec_item, irefdate, tzone, &
-                                                     tunit, tstart_user, target_data)
+      if (is_static_field) then
+         res = res .and. ec_gettimespacevalue_by_itemID(ecInstancePtr, ec_item, irefdate, tzone, &
+                                tunit, tstart_user, target_data)
```
FlorisBuwaldaDeltares: In the static field case we wanted an update here. The update worked because tgt_item1 would be an actual ec item, and for the dynamic case it didn't matter: at the end of ec_addtimespacerelation ec_item would still be -999 as nudgesalinitytemperature is handled internally.

however after writing back we must skip this update, as otherwise nudgesalinitytemperature writes back item id 1, temperature, and gets a separate update, which is wrong if the operand is anything other than "override"

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/fm_external_forcings_init.f90
FlorisBuwaldaDeltares: ah, this else ec_item = ec_undef_int was the necessary on the branch this commit was cherry-picked from: fm/task/UNST-9963_extforce_time_varying_generically

if the ec-item wasn't reset, the ec item would be added to the generic list of time-dependent items that needed an update.

Since it is not relevant on this branch, i will revert the change

src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_kernel/timespace/m_spatial_field.f90
FlorisBuwaldaDeltares: I've thought about it, i think this is correct. The allocate only cares about the requested allocation size, it doesn't care about the amount of quantities registered.

The registration counter is only incremented when a quantity is registered, and it is reset when the array is deallocated.

I considered not resetting it upon deallocate, but then you'll have to do it explicitly somewhere else, so this still seems like the cleanest option to me.

src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
FlorisBuwaldaDeltares: gettimespacevalue would update, since the ec relation is not destroyed.

I've added an fm_external_forcings_update to check whether the relation is not reapplied.
Also added a test to check whether both quantites are updated, and not just one of them

arthurvd (approved):

PR #1254, opened 2026-09-07
UNST-10281 Lower MinTimestepBreak to prevent error in e02_f901_c202

harmenwierenga:
# What was done 
Set MinTimestepBreak = 0.0005 s, this prevents the error that the time step becomes too small on both windows and linux with OneAPI 2026.

Testcase: e02_f901_c202_tidal_flume_kepsilon_zmodel_equi_3d

# Issue link
UNST-10281


JulienGroenenboom (changes requested):
As discussed, please check whether the "platform_specific" are still needed

harmenwierenga:
> As discussed, please check whether the "platform_specific" are still needed

I uploaded new references with the new windows compiler, and it works for both Linux and Windows on both OneAPI 2024 and 2026

JulienGroenenboom (approved):

PR #1255, opened 2026-09-07
support initialverticalsedfrac profiles, both Z and sigma

FlorisBuwaldaDeltares:
# What was done 

- case ('initialverticalsedfracprofile', 'initialverticalsigmasedfracprofile') added to initial 3D target resolver
- setinitialverticalprofile renamed to setinitialverticalprofilez
- setinitialverticalprofile now scans the quantity name to determine z or sigma layers (note, this is very awful, but making z or sigma a separate field for initial vertical profile quantities requires work from the hydrolib core team (@JulienGroenenboom thoughts?)
- two small vibe coded unit tests

# Evidence of the work done 

- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [ ]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [ ]	Not applicable 

# Issue link

### DIFF
--- a/src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
+++ b/src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
@@ -1900,19 +1900,86 @@ contains
    !$f90tw)
 
    !$f90tw TESTCODE(TEST, test_init_spatial_fields_integration, test_initialverticalsedfracprofile, test_initialverticalsedfracprofile,
+   !> Verifies that an initialverticalsedfracprofile [Spatial] block populates
+   !! the selected sediment constituent using absolute z coordinates.
    subroutine test_initialverticalsedfracprofile() bind(C)
-      call test_initial_vertical_sediment_profile(.false.)
+      use m_flow, only: kmx, kbot, ktop, zws, ndkx
+      use m_flowgeom, only: ndx2D, ndxi
+      use m_sediment, only: stm_included
+      use m_transportdata, only: constituents, const_names, NUMCONST
+      use m_flowtimes, only: irefdate, tzone, tstart_user
+      use m_polygon, only: m_polygon_destructor
+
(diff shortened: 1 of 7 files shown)


src/engines_gpl/dflowfm/packages/dflowfm_kernel/test/src/test_init_spatial_field_integration.f90
```diff
@@ -1899,6 +1899,114 @@ subroutine test_initialverticalsalinityprofile() bind(C)
...
+   !$f90tw)
+
+   !$f90tw TESTCODE(TEST, test_init_spatial_fields_integration, test_initialverticalsigmasedfracprofile, test_initialverticalsigmasedfracprofile,
+   subroutine test_initialverticalsigmasedfracprofile() bind(C)
+      call test_initial_vertical_sediment_profile(.true.)
+   end subroutine test_initialverticalsigmasedfracprofile
+   !$f90tw)
+
+   !> Verify that a vertical sediment profile populates its named constituent using z or sigma coordinates.
+   subroutine test_initial_vertical_sediment_profile(use_sigma)
```
rene-deltares: I see you are using "test parameterization" here. That's fine, but rather than having to branch on a boolean, I would prefer you would pass things like "quantityname", "expected_constituent_value" and "profile_coordinates" or something to this subroutine.

FlorisBuwaldaDeltares: I agree, will fix

FlorisBuwaldaDeltares: upon second thought, i think the parameterization threshold for me is 3. I have split it into two separate tests

rene-deltares:
I have converted e02_f039_c09 and e02_f039_c10 with the extforce converter. Those used to crash, but I verified that after this one is merged they not only run, but even pass the testbench!

rene-deltares (approved):
Looks good to me, and also more and nicer error messages

PR #1271, opened 2026-09-11
UNST-10125: Add retry workaround for cpython issue #131795

robin-deltares:
When this error is thrown, retry.

https://github.com/python/cpython/issues/131795

### DIFF
--- a/ci/teamcity/Delft3D/windows/test.kt
+++ b/ci/teamcity/Delft3D/windows/test.kt
@@ -144,10 +144,6 @@ object WindowsTest : BuildType({
                 uv pip sync pip/win-requirements.txt
                 if %%ERRORLEVEL%% NEQ 0 exit /b 1
 
-                rem Wait for five seconds. Kludge to get rid of the "-1073741819" exit codes we've
-                rem been dealing with during the module import phase of "Python TestBench.py".
-                ping -n 5 -w 1000 localhost > nul
-
                 rem Retry only on STATUS_ACCESS_VIOLATION (0xC0000005 / -1073741819).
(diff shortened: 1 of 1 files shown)


ci/teamcity/Delft3D/windows/test.kt
```diff
@@ -144,11 +144,28 @@ object WindowsTest : BuildType({
                 uv pip sync pip/win-requirements.txt
                 if %%ERRORLEVEL%% NEQ 0 exit /b 1
 
-                rem Wait for five seconds. Kludge to get rid of the "-1073741819" exit codes we've 
-                rem been dealing with during the module import phase of "Python TestBench.py"
+                rem Wait for five seconds. Kludge to get rid of the "-1073741819" exit codes we've
+                rem been dealing with during the module import phase of "Python TestBench.py".
                 ping -n 5 -w 1000 localhost > nul
```
rene-deltares: You might as well remove this sleep

rene-deltares (changes requested):
It's ugly, but the sleep was also ugly already, so no regression there...
I would still get rid of the old sleep. Since it's not needed anymore with the retries. And also, without the initial 5 second sleep, you should see more "retrying..." lines in the logs. Can you restart the entire pipeline with a testcase filter of 'asdfasdfasdfasdf' to verify that we indeed get a few of these retries? Thanks!

ci/teamcity/Delft3D/windows/test.kt
robin-deltares: https://github.com/Deltares/Delft3D/pull/1271/changes/6ea7ca693d5fcc750eb589eeca55632524f22d5e

PR #1276, opened 2026-09-15
UNST-8777 Convert ext for testcases in dimr_dflowfm_waq_coupling_aggregation_win64.xml

KoenBussemakerDeltares:
# What was done 

Migrated all cases including old file removal. All c20x cases went without a hitch. For the c30x cases the entries from the old ext file were added to the (already existing) new ext files, but this happened 4 times because of the different partition mdu files. To avoid this I migrated the only original mdu files and applied the same changes to the partition mdu's files by hand.

I created an issue for the duplicate entry problem: https://github.com/Deltares/HYDROLIB-core/issues/1208

# Evidence of the work done 

[dvc_diff.txt](https://github.com/user-attachments/files/32241399/dvc_diff.txt)

# Issue link

https://issuetracker.deltares.nl/browse/UNST-8777


FlorisBuwaldaDeltares (approved):
LGTM, no old ext files, no .baks, all .tims gone and replaced by .bc :)

PR #1279, opened 2026-09-16
Several small fixes

JulienGroenenboom:
to be merge-committed (instead of squashed) 

# What was done
- Remove non-existent izbndpos == 2 option
- Change warning to error when `use_salinity_freezing_point .and. temperature_min >= 0.0_dp`

# Evidence of the work done 
- [ ]	Video/figures \
- [ ]	Not applicable 

# Tests 
- [ ] Tests updated \
- [x]	Not applicable 

# Documentation  
- [ ]	Documentation updated \
- [x]	Not applicable 

# Issue link
-


src/engines_gpl/dflowfm/packages/dflowfm_kernel/src/dflowfm_data/unstruc_model.f90
```diff
@@ -1286,7 +1286,7 @@ subroutine readMDUFile(filename, istat)
       if (use_salinity_freezing_point .and. temperature_min >= 0.0_dp) then
          write (msgbuf, '(a,g0,a)') 'salinityDependentFreezingPoint is set to true, but Tempmin = ', temperature_min, &
             ' is not below 0 degrees Celsius. This may lead to incorrect results.'
-         call mess(LEVEL_WARN, msgbuf)
+         call mess(LEVEL_ERROR, msgbuf)
```
hrajagers, line 1289: Hi Julien, I support the transition from warning to error, but we need to change the message ... the user may no longer get incorrect results since we stop. It's not easy to come up with a proper yet concise description of the problem. Consider something like: "Set tempMin to a value below the freezing point of the highest expected salinity." instead of "This may lead to incorrect results.'"

PR #1287, opened 2026-09-18
UNST-10342 Converted ext files for DSLE cases

KoenBussemakerDeltares:
# What was done 

Converted ext file for 8 DSLE and 2 parallel DSLE cases. The regular cases converted without a problem, the parallel cases needed the partitioned mdu's to be updated by hand (only removing the old ext file reference).

# Evidence of the work done 

[dvc_diff.txt](https://github.com/user-attachments/files/32370332/dvc_diff.txt)

# Issue link

https://issuetracker.deltares.nl/browse/UNST-10342


rene-deltares (approved):
Great, looking good!

PR #1288, opened 2026-09-18
UNST-10106 Removed a bunch of duplicate mdu keywords from several tes…

KoenBussemakerDeltares:
…tbench tests

# What was done 

The dflowfm_io test branch flagged these tests as having duplicate keywords in the mdu file. When dflowfm parses the mdu file only the first encounter is parsed and the rest is ignored. All additional entries were removed.

No references have been updated.

# Evidence of the work done 

I used copilot to create a diff file of all mdu changes. It can't be viewed directly since these mdu's are tracked by dvc. Had to change the extension from .diff to .txt for github to accept it. Rename to .diff an use VS Code to view, it uses clear color coding.
[mdu_duplicate_keyword_fixes.txt](https://github.com/user-attachments/files/32381998/mdu_duplicate_keyword_fixes.txt)

# Issue link


arthurvd (approved):
I hand-compared the MDU of the first case, and that one is correct. I assume the other 8 files will contain similar good changes.

